PR #43, opened 2011-04-08
SVM module is now compliant with SQL API to the extent possible.

keesiong:


agorajek:
Thanks Kee Siong! I will review it in the next few days.

methods/kernel-machines/src/pg_gp/sql/online_sv_test.sql_in
```diff
@@ -1,31 +1,41 @@
 -- Example usage for regression:
 -- \i online_sv.sql_in
 
-select MADLIB_SCHEMA.svm_generate_reg_data(1000, 5);
-select MADLIB_SCHEMA.svm_regression('MADLIB_SCHEMA.svm_train_data', 'myexp', false);
-select MADLIB_SCHEMA.svm_predict('myexp', '{1,2,4,20,10}') > 0;
-select MADLIB_SCHEMA.svm_predict('myexp', '{1,2,4,20,-10}') < 0;
+select MADLIB_SCHEMA.svm_generate_reg_data('MADLIB_SCHEMA.svm_train_data', 10000, 5);
+select * from MADLIB_SCHEMA.svm_regression('MADLIB_SCHEMA.svm_train_data', 'regs', false, 'MADLIB_SCHEMA.svm_dot');
+select * from MADLIB_SCHEMA.svm_predict('regs', '{1,2,4,20,10}', 'MADLIB_SCHEMA.svm_dot') > 0;
```
agorajek: All lines were you have "SELECT \* FROM func() > 0;"  fails now. The comparison works OK if you skip "\* FROM" part but I will let you decide what you planned to test here. I think your current OUT file is from the previous version, cause it looks good.

keesiong: Sorry that's negligent of me to make last minute changes without checking. They are fixed now.
The output of the script will be slightly different each time. I will revisit this when we settle on some standard test procedure.
In the mean time, does the preprocessor work on the files in the expected folder. I have madlib instead of MADLIB_SCHEMA in expected/online_sv_test.out.

PR #45, opened 2011-04-20
Moving PLDA code from contrib to main repository after significant revisions.

keesiong:


eugenefratkin:
I merged the branch. We need this code for EMC demo

PR #46, opened 2011-06-06
Added svec to/from text conversion functions; Eugene to merge.

keesiong:


agorajek:
Please add a test call for each of the new functions.

PR #47, opened 2011-06-08
Added a function that allows creation of SVEC from an array of positions and array of values

eugenefratkin:
created a function svec_cast_positions_float8arr that creates svec from array of positions and array of values and target array size

svec_cast_positions_float8arr(INT[], FLOAT[], INT)


agorajek:
This is fantastic but a test case is missing...

eugenefratkin:
Yep, I forgot. I will added it
On 6/8/11 1:20 PM, "agorajek"
reply@reply.github.com
 wrote:

PR #48, opened 2011-06-08
Added a function that allows creation of SVEC from an array of positions and array of values and test cases

eugenefratkin:


agorajek:
Eugene, you added new function svec_cast_positions_float8arr(int8[],float8[],int8), but your new tests are for:
- float8arr_equals_internal,
- svec_to_string
- svec_from_string
  Am I missing something?

Also, I am not sure you tried it yourself before committing but I get this on install-check:

```
-- This vfunction test svec creation from position array
select madlib.float8arr_equals_internal('{1,2,4,6,2,5}'::INT8[], '{.2,.3,.4,.5,.3,.1}'::FLOAT8[], 10000);
psql:/tmp/madlib/log/20110608-160840/svec/test/svec.sql_in.tmp:179: ERROR:  function madlib.float8arr_equals_internal(bigint[], double precision[], integer) does not exist
LINE 1: select madlib.float8arr_equals_internal('{1,2,4,6,2,5}'::INT...
               ^
HINT:  No function matches the given name and argument types. You may need to add explicit type casts.
```

keesiong:
SparseData.c, line 188, @param array ~> @param array_pos.
And need @param end too.

keesiong:
SparseData.c, line 243, type_of_data Size of "weapons"?

keesiong:
SparseData.c, line 248, move "int compar(...) ..." before comment block for posit_to_sdata().

eugenefratkin:
There is bug - line 303 in SparseData.c should be outside of the IF statement and line 305 inside. Will fix tomorrow

keesiong:
Doing the swap for line 303 results in something incorrect, I think.

Also, is this intended? 
testdb=# select MADLIB.svec_cast_positions_float8arr('{3}'::INT8[], '{1}'::FLOAT8[], 2);

##  svec_cast_positions_float8arr 

 {2,1}:{0,1}

keesiong:
The for loop on line 273 gives the (false) impression that we're stepping through index[i] one-by-one, when in reality we are processing blocks at a time. For clarity, the for loop should probably be made a while loop.

keesiong:
Also, I thought Noelle wanted the option to specify a default value too?

eugenefratkin:
Damn right! - weapons. What are you a hippie or something? :) Do not know
where it came from, I'll fix it.

On 6/8/11 10:15 PM, "keesiong"
reply@reply.github.com
 wrote:

eugenefratkin:
Not sure why its incorrect? As far as this behavior... I am not sure what
is the correct schema. The goal of the last number is to be able to
declare array longer than the last position, but I guess it is misleading
if it is ignored when last position is greater than the limit and this is
OK. I guess I will through an error.

On 6/9/11 4:04 AM, "keesiong"
reply@reply.github.com
 wrote:

eugenefratkin:
Actually this is me who wanted it... I made the canges.
Eugene

On 6/9/11 4:14 AM, "keesiong"
reply@reply.github.com
 wrote:

keesiong:
Is this ready to be merged now, Eugene? Do you want to do it, Aleks, or shall I?

eugenefratkin:
Yes, I think it is. I cannot do it by protocol.:)

Best regards,
Eugene Fratkin

agorajek:
KeeSiong, have you tested these changes with the install-check script?

svec_cast_positions_float8arr is defined as (int8[],float8[],int8,float8), but called in the test script as (bigint[], double precision[], integer):

select madlib.svec_cast_positions_float8arr('{1,2,4,6,2,5}'::INT8[], '{.2,.3,.4,.5,.3,.1}'::FLOAT8[], 10000);
psql:/tmp/madlib/log/20110613-174603/svec/test/svec.sql_in.tmp:179: ERROR:  function madlib.svec_cast_positions_float8arr(bigint[], double precision[], integer) does not exist

keesiong:
Oops. Eugene, can you fix that up and I'll let Aleks do the merging this time.

agorajek:
Yep, let Eugene do the fix. But we also found a worse bug: http://jira.madlib.net/browse/MADLIB-130

PR #50, opened 2011-06-13
Reworked plda code that relies on sum(array[]) function only available on GP; Jira issue: MADLIB-118

keesiong:


agorajek:
Reviewing and testing...

agorajek:
Kee Siong, please let me have a lock on PLDA till tomorrow. I am struggling with fixing Python/SQL array expressions on Postgres...

PR #56, opened 2011-06-27
Fix for plda that allows it to run for more than maxfiles number of iterations

sokelvin:
Changed DROP/CREATE tables in each iteration to TRUNCATE.


eugenefratkin:
I will be out of office until June 28th. I this is urgent please contact Annika.Jimenez@emc.com for assistance.

PR #62, opened 2011-07-04
fix mfv merge logic

jhellerstein:
fix the merge logic in mfvsketch_merge_c as used by the mfvsketch_quick_histogram UDA.


jhellerstein:
Should address JIRA MADLIB-157.

PR #69, opened 2011-07-18
Modified decision tree, quantile and svec_sfv.

sokelvin:
now can run multiple times in a single session without complaining.
new quantile function plus doc update.
modified svec_sfv so now it checks whether dictionary is sorted or not. if unsorted, will error out.

### DIFF
--- a/src/config/Version.yml
+++ b/src/config/Version.yml
@@ -1,1 +1,1 @@
-version: 0.2.0beta
+version: 0.3.0master


methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -902,13 +923,21 @@ begin
...
-			UPDATE tree2 SET chisq = answer.chisq WHERE id = selection; 
+			--UPDATE tree2 SET feature = answer.feature WHERE id = selection;
+			--UPDATE tree2 SET probability = answer.probability WHERE id = selection;
+			--UPDATE tree2 SET maxclass = answer.maxclass WHERE id = selection;
+			--UPDATE tree2 SET infogain = answer.infogain WHERE id = selection;
+			--UPDATE tree2 SET cat_size = category_size[2] WHERE id = selection; 
+			--UPDATE tree2 SET live = 0 WHERE id = selection; 
+			--UPDATE tree2 SET chisq = answer.chisq WHERE id = selection; 
+			
+			EXECUTE 'UPDATE tree2 SET feature = '||answer.feature||' WHERE id =' || selection || ';';
```
agorajek: All these UPDATEs should be run as one statement. If tree2 gets large the time wasted on re-scanning the table on every UPDATE could be significant.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -920,19 +949,32 @@ begin
...
+				--UPDATE tree2 SET probability = 1.0 WHERE id = selection;
+				--UPDATE tree2 SET chisq = 1.0 WHERE id = selection;
+				--UPDATE tree2 SET maxclass = category_class WHERE id = selection;
+				--UPDATE tree2 SET infogain = 0 WHERE id = selection; 
+				--UPDATE tree2 SET cat_size = category_size[2] WHERE id = selection;
+				
+				EXECUTE 'INSERT INTO tree2 (tree_location, hash, feature, probability, maxclass, infogain, live, parent_id) VALUES(ARRAY['|| array_to_string(location,',') ||'],
+			 		 MADLIB_SCHEMA.hash_array(ARRAY['|| array_to_string(temp_location,',') ||']), 0, 1, 1, 1, 1,'||selection||')';
+				
+				EXECUTE 'UPDATE tree2 SET feature = 1 WHERE id = '|| selection||';';
```
agorajek: Same idea: combine all UPDATEs into one.

src/ports/postgres/modules/quantile/quantile.sql_in
```diff
@@ -55,35 +60,195 @@ Module grp_countmin for an approximate quantile implementation.
  *
  * This function computes the specified quantile value. It reads the name of the
  * table, the specific column, and computes the quantile value based on the
- * fraction specified as the third argument.
+ * fraction specified as the third argument. The functionality is the same as <tt>quantile</tt> except this implementation is designed to work more efficiently with large tables.
  */
-CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.quantile( table_name TEXT, col_name TEXT, quantile FLOAT) RETURNS FLOAT AS $$
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.quantile_large(table_name TEXT, col_name TEXT, quantile FLOAT) RETURNS FLOAT AS $$
```
agorajek: Not that I want to be picky, but we may use "quantile_big" instead of "quantile_large". Not only we get a shorter function name but we also get closer to the spirit of "big data".

Another thing to consider would be to rename "quantile" to "quantile_small" and create a new "quantile" that choses to run _small or _big version depending on table size info found in the system catalog (see pg_class: relpages and reltuples).

PR #71, opened 2011-08-01
Changes to madpack

agorajek:
Florian, please review and pull these when you get a chance:

1) Removed madpack dependency on PygreSQL.
2) Added user confirmation requirement for renaming previous MADlib schema during an update.


agorajek:
Refers to: http://jira.madlib.net/browse/MADLIB-217

agorajek:
OK Florian... this is ready for a review.

PR #72, opened 2011-08-11
Install-check updates

agorajek:
@fschopp: ready for review and merge.


src/madpack/madpack.py
```diff
@@ -887,30 +890,54 @@ def main(argv):
...
             sql_files = maddir_mod + '/' + module + '/test/*.sql_in'
             for sqlfile in glob.glob(sql_files):
             
+                result = 'PASS'
+
+                # Check if the SQL file is safe (= has no DROP statements)
+                f = open(sqlfile, 'r')
+                lines = f.readlines()
+                for line in lines:
+                    if re.search(r"\s*DROP\s+", line, re.IGNORECASE) \
```
fschopp: This only checks whether there is "DROP\s+" somewhere in the SQL file. This test yields a false positive if an identifier ending in "drop" is used.
So this protects against one particular error, but does not give real security. Would it be an option to create a role just for testing? This would imply that installcheck can only be run by users that are allowed to create another role, though. But then, there is no need for ordinary users to run installcheck anyway, is there?

src/madpack/madpack.py
```diff
@@ -178,6 +177,11 @@ def __run_sql_file(schema, maddir_mod, module, sqlfile, tmpfile, logfile):
     try:
         f = open(tmpfile, 'w')
         
+        # Add the before SQL
+        if pre_sql:
```
fschopp, line 181: `pre_sql` is never used, except for the following `f.writelines()`. What did you want to do with it?

src/madpack/madpack.py
```diff
@@ -887,30 +890,54 @@ def main(argv):
...
+
+                # Check the exit status
+                if retval != 0:
+                    __error("Failed executing %s" % tmpfile, False)  
+                    __error("Check the log at %s" % logfile, False) 
                     result = 'FAIL'
+                    keeplogs = True
                 # If error log file exists
-                elif os.path.getsize(logfile) > 0:
+                elif os.path.isfile(logfile) and os.path.getsize(logfile) > 0:
```
fschopp, line 950: Could you add a comment why it is an error if the log file is empty?

methods/decision_tree/src/pg_gp/sql/dec_trees.sql_in
```diff
@@ -1,15 +1,14 @@
```
fschopp, line 0: Shouldn't this (and some of the other files) be in a 'test' directory?

methods/array_ops/src/pg_gp/test/array_ops.sql_in
```diff
@@ -28,5 +38,7 @@ begin
...
 end $$ language plpgsql;
 
-SELECT MADLIB_SCHEMA.install_test();
-DROP FUNCTION MADLIB_SCHEMA.install_test();
\ No newline at end of file
+---------------------------------------------------------------------------
+-- Test
+---------------------------------------------------------------------------
+SELECT install_test();
\ No newline at end of file
```
fschopp, line 44: Does it matter whether there is a newline at the end of file?

methods/kernel_machines/src/pg_gp/sql/kernel_mach.sql_in
```diff
@@ -50,14 +47,13 @@ select * from MADLIB_SCHEMA.svm_novelty_detection('svm_train_data', 'nds', false
...
-select MADLIB_SCHEMA.svm_drop_model('clsp');
-select MADLIB_SCHEMA.svm_drop_model('clss');
-select MADLIB_SCHEMA.svm_drop_model('nds');
-select MADLIB_SCHEMA.svm_drop_model('regp');
-select MADLIB_SCHEMA.svm_drop_model('regs');
+--select MADLIB_SCHEMA.svm_drop_model('clsp');
+--select MADLIB_SCHEMA.svm_drop_model('clss');
+--select MADLIB_SCHEMA.svm_drop_model('nds');
+--select MADLIB_SCHEMA.svm_drop_model('regp');
+--select MADLIB_SCHEMA.svm_drop_model('regs');
```
fschopp: I don't think there is a need to keep those commented-out statements.

src/ports/postgres/modules/bayes/test/bayes.sql_in
```diff
@@ -63,43 +62,43 @@ declare
...
 begin
 	-- prepare training data: equal priors
-	DROP TABLE IF EXISTS data CASCADE;	
-	CREATE TABLE data( class INT, attrib FLOAT[] );
-	INSERT INTO data SELECT 1, ARRAY[fill_feature(id,0.3,num1),fill_feature(id,0.8,num1)] FROM generate_series(1,num1) as id;
-	INSERT INTO data SELECT 2, ARRAY[fill_feature(id,0.5,num2),fill_feature(id,0.5,num2)] FROM generate_series(1,num2) as id;
+	--DROP TABLE IF EXISTS data CASCADE;	
+	CREATE TABLE data_1( class INT, attrib FLOAT[] );
+	INSERT INTO data_1 SELECT 1, ARRAY[fill_feature(id,0.3,num1),fill_feature(id,0.8,num1)] FROM generate_series(1,num1) as id;
+	INSERT INTO data_1 SELECT 2, ARRAY[fill_feature(id,0.5,num2),fill_feature(id,0.5,num2)] FROM generate_series(1,num2) as id;
```
fschopp, line 68: What's the reason for these renaming changes?

src/madpack/madpack.py
agorajek: Yes, we could create a role for install-check, but this would mean that we would need to modify pg_hba.conf during this operation. I am almost OK with this idea, but I would like to give this another thought and do this after 0.2.1 patch.

src/madpack/madpack.py
agorajek, line 181: Here `pre_sql` is written to the temp file which is later executed using `psql`. It is used for example to adjust the SEARCH_PATH for each test file. See lines 898-899.

src/madpack/madpack.py
fschopp: > Yes, we could create a role for install-check, but this would mean that we would need to modify pg_hba.conf during this operation.

Is this really so? When you change the user in psql, does it establish a new connection?

src/madpack/madpack.py
agorajek, line 950: Would that work? :

```
            # If error log file exists and is not empty.
            # Empty log file means that either test file was empty or
            # something went wrong.
```

methods/decision_tree/src/pg_gp/sql/dec_trees.sql_in
agorajek, line 0: Remember, this is the legacy directory, which has several old conventions. I did not want to modify the dir names. Naturally once we move this to the new `src` we will drop the `src/pg_src` prefix and change `sql` dir to `test`.

methods/array_ops/src/pg_gp/test/array_ops.sql_in
agorajek, line 44: Not for `psql` or `madpack`. Why?

methods/kernel_machines/src/pg_gp/sql/kernel_mach.sql_in
agorajek: Agreed. Done: https://github.com/agorajek/madlib/commit/a83c06ac963f547da34e4a063a500a831df79e16

src/ports/postgres/modules/bayes/test/bayes.sql_in
agorajek, line 68: Since we can not use DROP, we need to make sure the tables we create have different names.

src/madpack/madpack.py
agorajek: Yes. I mentioned this issue during LA offsite. And few people said that this approach would be acceptable. I am still not 100% sure though.

fschopp: Have you seen this link? http://www.postgresql.org/docs/8.2/interactive/sql-set-role.html
It looks it would solve our issues.

src/madpack/madpack.py
fschopp, line 181: I see. It's a bit confusing because pre_sql can be arbitrary, but the comment added to the file is alway "Set SEARCH_PATH ...".

src/madpack/madpack.py
fschopp, line 950: Or "Since every single statement in the test file gets logged, an empty log file indicates an empty or a failed test."?

methods/array_ops/src/pg_gp/test/array_ops.sql_in
fschopp, line 44: In the back of my mind, I remember an issue around that. And I saw that some test files an with a newline and some don't.
Unfortunately, I don't remember what the issue was. In fact, I might not remember correctly at all.

src/madpack/madpack.py
agorajek: Wow, I think you are right Florian. Somehow I assumed it will behave the same as psql `\c` command. Thanks!

src/madpack/madpack.py
agorajek, line 950: OMG. You do know how to write concise comments :)

src/madpack/madpack.py
agorajek, line 181: I see your point. I will move that comment inside the `pre_sql` itself.

src/madpack/madpack.py
fschopp, line 950: Yesterday I was wondering why an empty log file should indicate an error. In particular if only warnings were logged, then it should be the ideal situation. So adding _why_ a log should not be empty helps a code reviewer save some seconds. :)

agorajek, line 950: I fully agree.

fschopp:
Improved and merged in.

PR #75, opened 2011-08-25
UDF that mimicks CREATE TABLE AS. Workaround for MADLIB-241

fschopp:
_The_ workaround. Please pull me!

### DIFF
--- a/doc/mainpage.dox
+++ b/doc/mainpage.dox
@@ -75,5 +75,11 @@ Usefull links:
 
     @defgroup grp_svec Sparse Vectors
     @ingroup grp_support
+    
+    @defgroup grp_compatibiity Compatibility
+    @ingroup grp_support
+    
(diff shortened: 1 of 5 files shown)


agorajek:
Florian, did you forget a test script or did not make one on purpose?

FYI: I can add one...

fschopp:
That's an oversight. Yes, please add a test. Thanks!

fschopp:
We should clearly document that this is a workaround (Action item, just added to the MADLIB-241), which is not meant to be used without need. Whenever possible, using standard SQL is preferred. Beyond that, I don't think there is anything else we can do. Running a count(*) would just take as long as creating the table itself, wouldn't it?

agorajek:
Yeah, I think that we should only advise on using this workaround UDF, rather that add some check that would slow it down.

PR #76, opened 2011-09-08
Fixed warnings in C code

fschopp:
I turned on almost all warnings (static analysis) that gcc supports. The goal should be to write warning-free code. Please carefully review my changes to your code that I did to fix the warnings.

Commit message:
Fixed all warnings in legacy C code. Unfortunately, you need gcc >= 4.6 so that all warnings disappear.
Changes included:
- Replaced calls to random() with POSIX function calls. For that I set the _XOPEN_SOURCE macro for all C files.
- Turned off strict aliasing for legacy C code.
- Avoid warning about uninitalized use of variables in mfvsketch.c. I actually moved some code here.


fschopp:
See also http://jira.madlib.net/browse/MADLIB-228.

fschopp:
All merged in.

PR #77, opened 2011-09-28
General improvements

agorajek:
- fix for MADLIB-256
- "reinstall" option added to madpack

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -1026,3 +1026,5 @@ if __name__ == "__main__":
     # Optional log files cleanup    
     if keeplogs is False:
         shutil.rmtree(tmpdir)
+    else:
+        print "INFO: Log files saved in " + tmpdir
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -659,24 +659,28 @@ def main(argv):
   
 """)
             
-    parser.add_argument('command', metavar='COMMAND', nargs=1, 
-                         choices=['install','update','uninstall','version','install-check'], 
-                         help = "One of the following options:\n"
-                              + "  install/update : run sql scripts to load into DB\n"
-                              + "  uninstall      : run sql scripts to uninstall from DB\n"
-                              + "  version        : compare and print MADlib version (binaries vs database objects)\n"
-                              + "  install-check  : test all installed modules\n")
+    parser.add_argument(
+        'command', metavar='COMMAND', nargs=1, 
+        choices=['install','update','uninstall','reinstall','version','install-check'], 
+        help = "One of the following options:\n"
+            + "  install/update : run sql scripts to load into DB\n"
+            + "  uninstall      : run sql scripts to uninstall from DB\n"
+            + "  reinstall      : performs uninstall and install\n"
+            + "  version        : compare and print MADlib version (binaries vs database objects)\n"
+            + "  install-check  : test all installed modules\n"
+    )
   
-    parser.add_argument('-c', '--conn', metavar='CONNSTR', nargs=1, dest='connstr', default=None,
-                         help= "Connection string of the following syntax:\n"
-                             + "  [user[/password]@][host][:port][/database]\n" 
-                             + "If not provided default values will be derived for PostgerSQL and Greenplum:\n"
-                             + "- user: PGUSER or USER env variable or OS username\n"
-                             + "- pass: PGPASSWORD env variable or runtime prompt\n"
-                             + "- host: PGHOST env variable or 'localhost'\n"
-                             + "- port: PGPORT env variable or '5432'\n"
-                             + "- db: PGDATABASE env variable or OS username\n"
-                             )
+    parser.add_argument(
+        '-c', '--conn', metavar='CONNSTR', nargs=1, dest='connstr', default=None,
+        help= "Connection string of the following syntax:\n"
+            + "  [user[/password]@][host][:port][/database]\n" 
+            + "If not provided default values will be derived for PostgerSQL and Greenplum:\n"
+            + "- user: PGUSER or USER env variable or OS username\n"
+            + "- pass: PGPASSWORD env variable or runtime prompt\n"
+            + "- host: PGHOST env variable or 'localhost'\n"
+            + "- port: PGPORT env variable or '5432'\n"
+            + "- db: PGDATABASE env variable or OS username\n"
+            )
 
     parser.add_argument('-s', '--schema', nargs=1, dest='schema', 
                          metavar='SCHEMA', default='madlib',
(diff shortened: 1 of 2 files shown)


methods/svec/src/pg_gp/SparseData.c
```diff
@@ -59,8 +63,16 @@ SparseData makeEmptySparseData(void) {
...
-	sdata->index->maxlen=0;
+
+	// -- KS fix
+	sdata->vals->data[0] = '\0';
+	sdata->index->data[0] = '\0';
+	sdata->vals->maxlen = 1;
+	sdata->index->maxlen = 1;
+
+	// sdata->vals->maxlen=0;
+	// sdata->index->maxlen=0;
```
cwelton: Remove dead code.

methods/svec/src/pg_gp/SparseData.c
```diff
@@ -59,8 +63,16 @@ SparseData makeEmptySparseData(void) {
 	pfree(sdata->index->data);
 	sdata->vals->data  = palloc(1);
 	sdata->index->data = palloc(1);
```
cwelton: Preferable might be to use `initStringInfoOfSize()` instead of palloc + munging the string info by hand.  

For that matter `makeSparseDataOfSize()` might be a more generally useful function than makeEmptySparseData

methods/svec/src/pg_gp/SparseData.c
```diff
@@ -94,7 +106,18 @@ SparseData makeInplaceSparseData(char *vals, char *index,
 	sdata->index->len = indexsize;
 	sdata->index->maxlen = sdata->index->len;
 	sdata->type_of_data = datatype;
+
+	// These will fail for sure
+	Assert(sdata->vals->maxlen > sdata->vals->len);
+	Assert(sdata->index->maxlen > sdata-index->len);
```
cwelton: I can see that these will fail... but why are we adding code that we _know_ is wrong???  I'm confused.

methods/svec/src/pg_gp/SparseData.c
```diff
@@ -317,8 +356,15 @@ SparseData posit_to_sdata(char *array, int64* array_pos, size_t width, Oid type_
 	
 	/* Add the final tallies */
 	sdata->unique_value_count = sdata->vals->len/width;
-	sdata->total_value_count = count;
+	// sdata->total_value_count = count;
```
cwelton: Remove dead code.

methods/svec/src/pg_gp/SparseData.c
```diff
@@ -317,8 +356,15 @@ SparseData posit_to_sdata(char *array, int64* array_pos, size_t width, Oid type_
...
+	sdata->total_value_count = end;
```
cwelton: I'd like an explanation of this change.

From what I can tell `total_value_count` should be a count of the number of total values in the sparse vector, which I given the function sounds like `count` to me:

>  @param count Number of element in the array

Whereas `end` should be, what, the total size allocated?

> @param end Target size of the element

So I would think that end is the desired allocation size.  But then I could be reading this wrong.

methods/svec/src/pg_gp/SparseData.c
```diff
@@ -553,13 +611,16 @@ SparseData concat(SparseData left, SparseData right) {
...
+	vals = (char *)palloc(sizeof(char)*(val_len+1)); // KS fix
+	index = (char *)palloc(sizeof(char)*(ind_len+1)); // KS fix
 	
 	memcpy(vals          ,left->vals->data,l_val_len);
 	memcpy(vals+l_val_len,right->vals->data,r_val_len);
 	memcpy(index,          left->index->data,l_ind_len);
 	memcpy(index+l_ind_len,right->index->data,r_ind_len);
+
+	vals[val_len] = '\0'; // KS fix
+	index[ind_len] = '\0'; // KS fix
```
cwelton: Why is managing all of that by hand better than using `appendBinaryStringInfo()`?  Using the right API we could allocate and copy less memory.

methods/svec/src/pg_gp/operators.c
```diff
@@ -672,6 +672,115 @@ Datum svec_l2norm(PG_FUNCTION_ARGS)
 	PG_RETURN_FLOAT8(accum);
 }
 
+PG_FUNCTION_INFO_V1( l2norm_two_svecs );
+/**
+ *  l2norm_two - Computes the l2norm distance between two SVECs.
```
cwelton: Inherit all my comments from the other CR here.

src/madpack/madpack.py
```diff
@@ -802,41 +806,24 @@ def main(argv):
...
 
         if __get_rev_num(dbrev) == ['0']:
             __info("Nothing to uninstall. No version found in schema %s." % schema.upper(), True)
             return
 
+        __info("*** Uninstalling MADlib ***", True)
+
         __info("***************************************************************************", True)
         __info("* Schema %s and all database objects depending on it will be dropped!" % schema.upper(), True)
         __info("* This is potentially very dangerous operation!                      ", True)
```
cwelton: We do need to spend more time thinking about upgrade/reinstall.  The fact that reinstall will drop the old madlib which in turn implies dropping svec which in turn implies dropping any user data that uses svec is a serious issue.  It's so serious that "This is a potentially very dangerous operation" is putting it lightly.   Might be good to expand this check and report exactly what it will end up dropping.

methods/svec/src/pg_gp/operators.c
agorajek: Removed from this pull-request. This belongs to k-means group.

src/madpack/madpack.py
agorajek: Addressed by adding a list of tables/columns to be dropped during the uninstallation. See: https://github.com/agorajek/madlib/commit/5b3dd7abb30583a037de364cdbaf70c17366bec8

methods/svec/src/pg_gp/SparseData.c
agorajek: All comments on SparseData.c referred to Kee Siong.

agorajek: Per Kee Siong's suggestion SparseData.c updates have been dropped from this pull request.

src/madpack/madpack.py
```diff
@@ -802,45 +803,55 @@ def main(argv):
...
+        __info("*** Uninstalling MADlib ***", True)        
+        __info("***********************************************************************************", True)
         __info("* Schema %s and all database objects depending on it will be dropped!" % schema.upper(), True)
-        __info("* This is potentially very dangerous operation!                      ", True)
-        __info("***************************************************************************", True)
+        if affected_objects:
+            __info("* If you continue the following data will be lost (schema : table.column : type):", True)
+            for ao in affected_objects:
+                __info ( '* - ' + ao['schema'] + ' : ' + ao['relation'] + '.' + ao['column'] + ' : ' + ao['type'], True);
+        __info("***********************************************************************************", True)
```
cwelton, line 854: I like this as an interim solution.

methods/svec/src/pg_gp/SparseData.c
cwelton: Thanks, it makes sense as a separate Pull.

PR #78, opened 2011-10-13
madlib + gppkg

abhigp:
I have created a flag which can be set ON or OFF during the configure step. In order to build a gppkg, 
the user will have to run the following commands,

./configure -DGPPKG_BUILD=ON
cd build
make gppkg

This will turn off the automatic dependency checking feature for the madlib rpm as this is currently a requirement for gppkg. 

In order to build a normal rpm, the packager has to run the configure step again without the GPPKG_BUILD flag.


deploy/RPM/CMakeLists.txt
```diff
@@ -16,15 +16,26 @@ set(CPACK_RPM_PACKAGE_LICENSE "New BSD License" PARENT_SCOPE)
...
 
-set(_PACKAGE_REQUIRES "lapack >= 3.0, m4 >= 1.4")
-if(RH_VERSION AND RH_VERSION VERSION_LESS "6.0")
-    # on RH/CentOS 5, there is no Python 2.6 or higher in the defaul
-    # repositories.
-    set(CPACK_RPM_PACKAGE_REQUIRES "python, ${_PACKAGE_REQUIRES}" PARENT_SCOPE)
-else()
-    set(CPACK_RPM_PACKAGE_REQUIRES "python >= 2.6, ${_PACKAGE_REQUIRES}" PARENT_SCOPE)
+if(GPPKG_BUILD)
+    set(CPACK_RPM_PACKAGE_PROVIDES "/bin/sh" PARENT_SCOPE)
```
fschopp, line 28: What is that good for? Why /bin/sh?

deploy/CMakeLists.txt
```diff
@@ -37,11 +41,52 @@ set(CPACK_PACKAGE_VERSION_MAJOR ${MADLIB_VERSION_MAJOR})
...
 
+
 # CPACK_PACKAGING_INSTALL_PREFIX has to be set in the generator-specific
 # section!
 
 # -- Set settings for specific package generators ------------------------------
 
+if(GPPKG_BUILD)
+    add_custom_target(gppkg 
+                      COMMAND make package
```
fschopp: Usually, this would be a handled with `add_dependencies(...)`

deploy/CMakeLists.txt
```diff
@@ -37,11 +41,52 @@ set(CPACK_PACKAGE_VERSION_MAJOR ${MADLIB_VERSION_MAJOR})
...
+if(GPPKG_BUILD)
+    add_custom_target(gppkg 
+                      COMMAND make package 
+                      COMMAND cp -rf ${CMAKE_SOURCE_DIR}/deploy/package .
+                      COMMAND rm -rf package/gppkg/deps
+                      COMMAND mkdir package/gppkg/deps
+                      COMMAND mv lapack-*-*.rpm package/gppkg/deps/
+                      COMMAND mv blas-*-*.rpm package/gppkg/deps/
+                      COMMAND cp ${CMAKE_SOURCE_DIR}/build/madlib-*.rpm package/gppkg/
+                      COMMAND sed -i".bk" 's/os/${CMAKE_SYSTEM_NAME}/g' package/gppkg/gppkg_spec.yml
```
fschopp: Looks dangerous. "os" is not very distinctive.

deploy/CMakeLists.txt
```diff
@@ -37,11 +41,52 @@ set(CPACK_PACKAGE_VERSION_MAJOR ${MADLIB_VERSION_MAJOR})
...
+                      COMMAND rm -f package/gppkg/gppkg_spec.yml.bk
+                      COMMAND gppkg --build package/gppkg 
+                      COMMAND mv *.gppkg ${CMAKE_SOURCE_DIR}/build
+                     )   
+
+    add_custom_target(lapack_rpm
+                      COMMAND rm -rf lib
+                      COMMAND mkdir lib
+                      COMMAND cp /usr/lib/liblapack.so* lib/
```
fschopp: Is it not lib64 for 64 bit?

deploy/package/SPECS/lapack.spec
```diff
@@ -0,0 +1,34 @@
...
+Release:        %{release}
+Group:          Development/Tools
+Prefix:         /temp
+AutoReq:        no
+AutoProv:       no
+BuildArch:      %{buildarch} 
+Provides:       lapack-%{version}-%{release}.%{buildarch}.rpm 
+
+%description
+The Pgcrypto package provides cryptographic package for the Greenplum Database.
```
fschopp: ?

deploy/CMakeLists.txt
```diff
@@ -37,11 +41,52 @@ set(CPACK_PACKAGE_VERSION_MAJOR ${MADLIB_VERSION_MAJOR})
...
+                      COMMAND rm -rf package/gppkg/deps
+                      COMMAND mkdir package/gppkg/deps
+                      COMMAND mv lapack-*-*.rpm package/gppkg/deps/
+                      COMMAND mv blas-*-*.rpm package/gppkg/deps/
+                      COMMAND cp ${CMAKE_SOURCE_DIR}/build/madlib-*.rpm package/gppkg/
+                      COMMAND sed -i".bk" 's/os/${CMAKE_SYSTEM_NAME}/g' package/gppkg/gppkg_spec.yml
+                      COMMAND rm -f package/gppkg/gppkg_spec.yml.bk
+                      COMMAND gppkg --build package/gppkg 
+                      COMMAND mv *.gppkg ${CMAKE_SOURCE_DIR}/build
+                     )
```
fschopp: CMake has commands for all that (file system operations, substitutions, etc.) that guarantee portability. It's fine here, but I wouldn't want to include it in the MADlib main repository like that.

fschopp:
See my line annotations. I believe this works fine for generating gppkg packages. But as it stands now, I don't think we should include your modifications in the MADlib main repository for several reasons:
- The gppkg sections make strong assumptions about the system setup and how CMake is used (-> portability issues)
- For being pulled into the main repository, there should be a clearer separation between gppkg and general sections (e.g., provide a general option RPM_DONT_DECLARE_DEPENDENCIES and add a Contrib subdirectory with a separate CMakeLists.txt)

deploy/RPM/CMakeLists.txt
abhigp, line 28: gppkg uses a separate database than the rpm database present on the system and hence has no knowledge of the programs installed on the system. So when you try to install this package, you will get an error saying that sh has not been installed. This line tells rpm that the required dependency is satisfied by the package itself and will prevent rpm from erorring out. The sh dependency is automatically added as a dependency by the rpm when there are post/pre hooks in the spec file.

deploy/package/SPECS/lapack.spec
abhigp: My bad. Forgot to change the description.

deploy/CMakeLists.txt
abhigp: From the build process on the vm 

-- Found a LAPACK library: /usr/lib/liblapack.so
-- Found a BLAS library: /usr/lib/libblas.so

So I went ahead and included these libraries.

deploy/RPM/CMakeLists.txt
fschopp, line 28: I see. I suggest adding exactly this comment, so that people reading the source know.

deploy/CMakeLists.txt
fschopp: But:

```
[schopf@localhost build]$ uname -a
Linux localhost.localdomain 2.6.18-238.el5 #1 SMP Thu Jan 13 15:51:15 EST 2011 x86_64 x86_64 x86_64 GNU/Linux
[schopf@localhost build]$ ldd src/lib/libarmadillo.so
        linux-vdso.so.1 =>  (0x00007fffd6595000)
        libblas.so.3 => /usr/lib64/libblas.so.3 (0x00002aadf6dce000)
        liblapack.so.3 => /usr/lib64/liblapack.so.3 (0x00002aadf7024000)
        libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00002aadf772d000)
        libm.so.6 => /lib64/libm.so.6 (0x00002aadf7a2d000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00002aadf7cb1000)
        libc.so.6 => /lib64/libc.so.6 (0x00002aadf7ebf000)
        libgfortran.so.1 => /usr/lib64/libgfortran.so.1 (0x00002aadf8217000)
        /lib64/ld-linux-x86-64.so.2 (0x0000003dc1c00000)
```

fschopp: Have you tested your package on a system that does not have LAPACK/BLAS installed in /usr/lib64?

abhigp: Yes. You are right. Running ldd on libarmadillo points to the libraries in lib64.

abhigp: I tested package on a system without /usr/lib4 and it did not work
correctly. I used the libraries from lib64 and installed it on a machine
without the libraries and it worked properly. I will make this change.

On 10/27/11 5:30 PM, "Florian Schoppmann"
reply@reply.github.com
wrote:

abhigp:
Hi Florian,

```
Thanks for the input. I will make the changes as you have requested
```

and get back to you with the new changes.

Thanks
Abhijit

On 10/27/11 5:02 PM, "Florian Schoppmann"
reply@reply.github.com
 wrote:

fschopp:
As discussed: Not integrating at this point.

PR #80, opened 2011-12-01
Rewrite Decision tree

aojwang:
The decision tree implementation will based on c4.5, including:
1. Supporting information gain, gain ratio and gini as split criterion 
2. Supporting both continuous attribute and discrete attribute
3. Supporting missing attribute
4. Supporting both tabular format input and sparse vector format input
5. Supporting reduce error pruning and error based pruning.
6. Make the training and classification scalable.

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.c
+++ b/methods/decision_tree/src/pg_gp/decision_tree.c
@@ -1,254 +1,1580 @@
+/* ----------------------------------------------------------------------- *//**
+ *
+ * @file decision_tree.c
+ *
+ * @brief decision tree related utility functions written with C
+ *
+ *//* -------------------------------------------------------------------- */
 #include "postgres.h"
 #include "fmgr.h"
 #include "utils/array.h"
 #include "math.h"
 #include "catalog/pg_type.h"
 
+/*
+ * Copy from float_specials.h of svec to deal with the NVP value
+ * If we can directly refer to the header file, then we should use it instead.
+ */
+#define MKINT64(x) (x##ULL)
+#define NEG_QUIET_NAN_MIN64    MKINT64(0xFFF8000000000001)
+#define NVP_i		NEG_QUIET_NAN_MIN64
+static const uint64 COMPVEC_I[] = {NVP_i};
+static const double *COMPVEC = (double *)COMPVEC_I;
+#define NVP		COMPVEC[0]
+#define IS_NVP(x)  (memcmp(x,&(NVP),sizeof(double)) == 0)
+
 #ifndef NO_PG_MODULE_MAGIC
 PG_MODULE_MAGIC;
 #endif
 
+/*
+ * This function judge whether certain element in Array is null.
+ * That array should be already converted to bitmap.
+ *
+ * Parameters:
+ *      bitmap: The bitmap converted from the original array.
+ *      offset: The index of the element to be tested. 0 is the first 
+ *              element in the array.
+ * Return:
+ *      true:  The specified element is null.
+ *      false: The specified element is not null.
+ */
+static bool array_is_null
+    (
+        const bits8*    bitmap, 
+        int             offset
+    )
+{
+	return bitmap == 0 ? false : !(bitmap[offset / 8] & (1 << (offset % 8)));
+}
+
+/*
+ * For float value, we cannot directly compare it with zero.
+ * This function is used to judge whether a float value is 0.
+ *
+ * Parameters:
+ *      value:      a float value
+ *
+ * Return:
+ *      True:       The value passed in is zero.
+ *      False:      The value passed in is not zero.
+ *
+ */
+static bool is_float_zero(float8 value)
+{
+    float8 margin_value = 0.00000000000001;
+    if( value < margin_value && value > -margin_value )
+        return true;
+    else
+        return false;
+}
+
 Datum hash_array( PG_FUNCTION_ARGS);
 
+/*
+ * This function computes one hash value from one
+ * integer array. 
+ *
+ * Parameters:
+ *      first para: The integer array.
+ * Return:
+ *      The integer hash value.
+ */
 PG_FUNCTION_INFO_V1(hash_array);
 Datum hash_array( PG_FUNCTION_ARGS)
 {
-	ArrayType *state  = PG_GETARG_ARRAYTYPE_P(0);
-	int dimstate = ARR_NDIM(state);
+    ArrayType *state  = PG_GETARG_ARRAYTYPE_P(0);
+    int dimstate = ARR_NDIM(state);
     int *dimsstate = ARR_DIMS(state);
-	int numstate = ArrayGetNItems(dimstate,dimsstate);
+    int numstate = ArrayGetNItems(dimstate,dimsstate);
     int32 *vals_state=(int32 *)ARR_DATA_PTR(state);
-	
+
     unsigned long hash = 65599;
     unsigned short c;
     int i = 0;
-    
+
     for (;i<numstate;i++)
-	{
-		c = vals_state[i];
-		hash = c + (hash << 7) + (hash << 16) - hash;
-	}
+    {
+        c = vals_state[i];
+        hash = c + (hash << 7) + (hash << 16) - hash;
+    }
 
-	PG_RETURN_INT32(hash);
+    PG_RETURN_INT32(hash);
 }
 
-static float8 ChiSquareStatistic(float8* values, int from, int size, float8 fract, float8 total){
-	int i = from, j = 1;
-	float8 mult = 0;
-	float8 estimate = 0;
-	float8 result = 0;
-	
-	if((fract == 0.0)||(total == 0)){
-		return 0;
-	}
-	mult = fract/total;
-	for(; i < (size+from); ++i, ++j){
-		estimate = ((float8)values[j])*mult;
-		if(estimate > 0){
-			result += ((float8)values[i] - estimate)*((float8)values[i] - estimate)/estimate;
-		}
-	}
-	return result;
+/*
+ * Below two arrays are taken from Documenta Geigy Scientific
+ * Tables (Sixth Edition), p185 (with modifications).)
+ */
+static float CONFIDENCE_LEVEL[] = {  0,  0.001, 0.005, 0.01, 0.05, 0.10, 0.20, 0.40, 1.00},
+      CONFIDENCE_DEV[] = {4.0,  3.09,  2.58,  2.33, 1.65, 1.28, 0.84, 0.25, 0.00};
+
+/*
+ * Compute the addition errors for each node. This algorithm is based on
+ * the definiton of 'error-based pruning'.
+ * The detailed description of that pruning strategy can be found in the paper
+ * of 'Error-Based Pruning of Decision Trees Grown on Very Large Data Sets Can Work!'.
+ *
+ * Parameters:
+ *      totalCases:  It specifies how many cases in the training set fall into
+ *                   the current branch.
+ *      numOfErrors: This argument should be equal to 
+ *                   'totalCases - cases belonging to maxClass'.
+ *      confLevel:   Please refer to the paper for detailed definition
+ * Return:
+ *      The computed additional error that should be added to the original error
+ *      number.
+ *
+ */
+static float compute_added_errors
+    (
+        float totalCases, 
+        float numOfErrors, 
+        float confLevel
+    )
+{
+    static float coeff=0;
+
+    if ( ! coeff )
+    {
+        int i = 0;
+        while ( confLevel > CONFIDENCE_LEVEL[i] ) i++;
+
+        coeff = CONFIDENCE_DEV[i-1] +
+                  (CONFIDENCE_DEV[i] - CONFIDENCE_DEV[i-1]) *
+                  (confLevel - CONFIDENCE_LEVEL[i-1]) /
+                  (CONFIDENCE_LEVEL[i] - CONFIDENCE_LEVEL[i-1]);
+
+        coeff = coeff * coeff;
+    }
+
+    if (numOfErrors < 1E-6)
+    {
+        return totalCases * (1 - exp(log(confLevel) / totalCases));
+    }
+    else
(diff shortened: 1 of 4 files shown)


methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *
+ * Return:
+ *      True:       The value passed in is zero.
+ *      False:      The value passed in is not zero.
+ *
+ */
+static bool is_float_zero(float8 value)
+{
+    //double only has up to 15 valid digits
+    float8 margin_value = 1E-13;
```
cwelton, line 66: Why not use DBL_EPSILON?  It's a standard definition that should be defined for the platform and give you a more portable result.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *
+ */
+static float compute_added_errors
+    (
+        float totalCases, 
+        float numOfErrors, 
+        float confLevel
+    )
+{
+    static float coeff=0;
```
cwelton, line 169: Why is this variable static?  This seems dangerous.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+
+    if ( ! coeff )
+    {
+        int i = 0;
+        while ( confLevel > CONFIDENCE_LEVEL[i] ) i++;
```
cwelton, line 174: No checking has been done on the range of confLevel, this has the potential to compare to uninitialized memory.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+
+        coeff = CONFIDENCE_DEV[i-1] +
```
cwelton, line 176: If confLevel == 0 this refers to a negative offset in the array assigning values based on stack memory.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+/*
+ * Below two arrays are taken from Documenta Geigy Scientific
+ * Tables (Sixth Edition), p185 (with modifications).)
+ */
+static float CONFIDENCE_LEVEL[] = {  0,  0.001, 0.005, 0.01, 0.05, 0.10, 0.20, 0.40, 1.00},
+      CONFIDENCE_DEV[] = {4.0,  3.09,  2.58,  2.33, 1.65, 1.28, 0.84, 0.25, 0.00};
+
+/*
+ * Compute the addition errors for each node. This algorithm is based on
+ * the definiton of 'error-based pruning'.
```
cwelton, line 147: Is there a reference for 'error-based-pruning'?  Hard to evaluate if the formulas given here are correct.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+    {
+        return 0.67 * (totalCases - numOfErrors);
+    }
+    else
+    {
+        float tmp = (numOfErrors + 0.5 + coeff/2
+                + sqrt(coeff * ((numOfErrors + 0.5) * (1 - (numOfErrors + 0.5)/totalCases) + coeff/4)) )
+             / (totalCases + coeff);
+        return (totalCases * tmp - numOfErrors);
+    }
```
cwelton, line 206: Multiple magic numbers above:
- 1E-6 : coarse standin for DBL_EPSILON?
- 0.9999 : coarse standin for 1- DBL_EPSILON?  Why isn't this just < 1?
- 0.67 : ???

We need some description of what is going on here.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *                  should begin.
+ *      size:       It specifies the total number of values used for
+ *                  chisquare calculation.
+ *      fract:      It specifies the number of cases belonging to this
+ *                  fraction.
+ *      total:      It specifies the total number of cases.
+ * Return:
+ *      The computed chisquare value.
+ *
+ */
```
cwelton, line 255: I'd prefer to see this based on the boost implementation, but given that this is using C instead of C++ I concede that is more difficult.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
-	int32 posclasses = PG_GETARG_INT32(3);
-	int32 trueclass = PG_GETARG_INT32(5);
-   	ArrayType *pgarray;
+    int numstate = ArrayGetNItems(dimstate,dimsstate);
+    int64 *vals_state=(int64 *)ARR_DATA_PTR(state);
+
+    int classifiedClass = PG_GETARG_INT32(1);
+    int originalClass = PG_GETARG_INT32(2);
+    ArrayType *pgarray;
+
```
cwelton, line 362: This is being exposed as a user function.  You cannot simply take the data pointer of the array, add some user supplied constant and start modifying memory there without checking that the memory is within the bounds of the array.  That is an extremely serious and severe security vulnerability not to mention can cause massive data corruption and other issues.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+    if(originalClass != classifiedClass)
+    	++vals_state[0];
+
+    ++vals_state[originalClass];
```
cwelton, line 366: Under normal circumstances you are never allowed to modify pass by reference parameters.  You must always make a copy and modify your copy instead.

The one exception to that is when the function is called as an aggregate transition function, in which case you are allowed to modify the aggregate state (first parameter) in place.  However it is absolutely necessary that you first verify that the function is being called in an aggregate context.  Otherwise it is incorrect behavior and must be avoided.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
 
-	vals_state[0] += trueweight;
-	vals_state[trueclass] += trueweight;
-	
-	vals_state[(int)(truevalue*(posclasses+1))] += trueweight;
-	vals_state[(int)(truevalue*(posclasses+1) + trueclass)] += trueweight;
-  
     pgarray = construct_array((Datum *)vals_state,
```
cwelton, line 368: After having updated the input in place above you then go and construct a new array to return here.  This doesn't make sense to me.  If we are updating in place then we should return our input, if we are not then we should have constructed this array earlier and modified it.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *      branch
+ */
+Datum compute_rep(PG_FUNCTION_ARGS);
+
+PG_FUNCTION_INFO_V1(compute_rep);
+Datum compute_rep(PG_FUNCTION_ARGS) {
+    ArrayType *state  = PG_GETARG_ARRAYTYPE_P(0);
+    int dimstate = ARR_NDIM(state);
+    int *dimsstate = ARR_DIMS(state);
+    int numstate = ArrayGetNItems(dimstate,dimsstate);
```
cwelton, line 397: Please don't have two variables named "dimstate" and "dimsstate", way too easy to confuse these two.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+
+    int64 *result = palloc(sizeof(int64)*2);
+
+    ArrayType *pgarray;
+
+    int64 max = vals_state[1];
+    int64 sum = max;
+    int maxid = 1;
+
+    for(int i = 2; i < numstate; ++i)
```
cwelton, line 408: This cannot possibly be correct for all inputs.  Consider input arrays with different dimensionality.

cwelton, line 408: I'm going to assume for the moment that this code is one of the things that will go away when we convert over to the new split methodology.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *                            denote how many records there are originally. 
+ *      5 numOfClasses:       The total number of distinct classes.   
+ *      6 trueclass:          The true class this record belongs to.
+ *      7 has_cont_feature:   For all those features, is one of them is continuous.
+ * Return:
+ *      The array containing all the information for the calculation of 
+ *      split gain. 
+ */
+PG_FUNCTION_INFO_V1(aggr_infogain);
+Datum aggr_infogain(PG_FUNCTION_ARGS) {
```
cwelton, line 504: I'm assuming this is also going away with the implementation.  Skipping review.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *       1 value:   It defines the initial value. All the
+ *                  elements of that array are initialized
+ *                  with that value.
+ * Return:
+ *      The array requested by user.
+ */
+Datum malloc_and_set(PG_FUNCTION_ARGS);
+
+PG_FUNCTION_INFO_V1(malloc_and_set);
+Datum malloc_and_set(PG_FUNCTION_ARGS) {
```
cwelton, line 1337: We should not expose functions name malloc_\* to SQL.   Consider what the function is doing and give it a better name.  Perhaps build_float8_array()?  I assume this is intended to be internal to the DT module?

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+    float8 *result = (float8*)palloc(sizeof(float8)*size); 
+    ArrayType *pgarray;
 
-PG_FUNCTION_INFO_V1(mallocset);
-Datum mallocset(PG_FUNCTION_ARGS) {
-	int size = PG_GETARG_INT32(0);
-	int value = PG_GETARG_INT32(1);
-	float8 *result = (float8*)palloc(sizeof(float8)*size); 
-	ArrayType *pgarray;
+    memset(result, value, sizeof(float8)*size);
```
cwelton, line 1343: Note memset is "char" based not "int" based.  For that matter if the user is building a float8 array the logical choice here would be to accept a float8 value for input?

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
-Datum WeightedNoReplacement(PG_FUNCTION_ARGS) {
-	int value1 = PG_GETARG_INT32(0);
-	int value2 = PG_GETARG_INT32(1);
-	int64 *result = (int64*)palloc(sizeof(int64)*value1);
-	int32 i = 0;
-	int32 k = 0;
-	ArrayType *pgarray;
-	float to_select = value1;
+PG_FUNCTION_INFO_V1(malloc_and_set_int64);
+Datum malloc_and_set_int64(PG_FUNCTION_ARGS) {
```
cwelton, line 1365: Same comments as malloc_and_set_float8 above.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+Datum weighted_no_replacement(PG_FUNCTION_ARGS) {
+    int value1 = PG_GETARG_INT32(0);
+    int value2 = PG_GETARG_INT32(1);
+    int64 *result = (int64*)palloc(sizeof(int64)*value1);
+    int32 i = 0;
+    int32 k = 0;
+    ArrayType *pgarray;
+    float to_select = value1;
+
+    for(;i < value2; i++){
```
cwelton, line 1406: This algorithm is wrong.  

It can easily produce an output array consisting of uninitialized memory.  Worse it has the potential of corrupting memory.

Also for large values of "value2" this would be an extremely inefficient solution even if it was correct.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *       0 x:    a float value
+ *       1 y:    a float value
+ *                  
+ * Return:
+ *       The value which is smaller than the other.
+ */
+Datum min(PG_FUNCTION_ARGS);
+
+PG_FUNCTION_INFO_V1(min);
+Datum min(PG_FUNCTION_ARGS) {
```
cwelton, line 1437: Unnecessary.  Remove this function completely and use float8smaller()

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+ *       0 x:    a float value
+ *       1 y:    a float value
+ *                  
+ * Return:
+ *       0- If x<y; 1- If x>=y.
+ */
+Datum is_less(PG_FUNCTION_ARGS);
+
+PG_FUNCTION_INFO_V1(is_less);
+Datum is_less(PG_FUNCTION_ARGS) {
```
cwelton, line 1462: Unnecessary, remove this function completely and use float8lt() or simply the '<' operator.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1631 @@
...
+    PG_RETURN_INT32(ret);
+}
+
+/*
+ * The function used for gamma calculation. We should leverage other
+ * implementation in the boost lib. 
+ * We will delete this function after the integration.
+ */ 
+float8 gammaln(float8 x) {
+    float8 xnum = 0;
```
cwelton, line 1479: Better if we could simply use the  boost implementation.  

Distribution functions would be better expressed as their own methods rather than being a component of DT like this.

cwelton:
Closing dead pull request.

PR #83, opened 2011-12-16
Fix assertion condition, in IS_SCALAR case.

haradh1:
This is a small fix for install-check failure. Assertion condition is overlooked in the scalar case.

http://jira.madlib.net/browse/MADLIB-304

### DIFF
--- a/methods/svec/src/pg_gp/operators.c
+++ b/methods/svec/src/pg_gp/operators.c
@@ -241,7 +241,8 @@ Datum svec_change(PG_FUNCTION_ARGS)
 	int midlen = middle->total_value_count;
 	SparseData head = NULL, tail = NULL, ret = NULL;
 
-	Assert(midlen == changed->dimension);
+	Assert((IS_SCALAR(changed) && midlen == 1) ||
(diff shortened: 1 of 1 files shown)


methods/svec/src/pg_gp/operators.c
```diff
@@ -241,7 +241,8 @@ Datum svec_change(PG_FUNCTION_ARGS)
 	int midlen = middle->total_value_count;
 	SparseData head = NULL, tail = NULL, ret = NULL;
 
-	Assert(midlen == changed->dimension);
+	Assert((IS_SCALAR(changed) && midlen == 1) ||
```
fschopp, line 244: I think it is helpful to add a comment to the source code and/or reference the MADlib Jira.

haradh1, line 244: Well, I think it's obvious without links or comments. Do we need every fix linking to Jira??

haradh1, line 244: But feel free to modify it when you pull it. Thanks

cwelton:
Merged.

PR #85, opened 2011-12-20
Greenplum 4.1 compatibility / Underflow-safe logistic function

fschopp:
These additions were the direct response to user requests. They should not be needed for GPDB4.2 any more, but they are needed for earlier versions.

### DIFF
--- a/src/ports/greenplum/modules/compatibility/compatibility.sql_in
+++ b/src/ports/greenplum/modules/compatibility/compatibility.sql_in
@@ -20,15 +20,23 @@ known limitations in support of the SQL standard or bugs.
 
 This module contains workarounds for the following issues:
 
-- Using <tt>CREATE TABLE AS</tt> with the output of certain MADlib functions
-  fails with the error “function cannot execute on segment because it issues a
-  non-SELECT statement”. The workaround is:
+- <tt>CREATE TABLE <em>table_name</em> AS <em>query</em></tt> statements where
+  <em>query</em> contains certain MADlib functions fails with the error
+  “function cannot execute on segment because it issues a non-SELECT statement”.
+  The workaround is:
   <pre>SELECT \ref create_table_as('<em>table_name</em>', $$
     <em>query</em>
 $$, 'BY (<em>column</em>, [...]) | RANDOMLY');</pre>
+- <tt>INSERT INTO <em>table_name</em> <em>query</em></tt> where <em>query</em>
+  contains certain MADlib functions fails with the error “function cannot
+  execute on segment because it issues a non-SELECT statement”. The workaround
+  is:
+  <pre>SELECT \ref insert_into('<em>table_name</em>', $$
+    <em>query</em>
+$$);</pre>
 
 @note
-These function are not installed on other DBMSs (and not needed there).
+These functions are not installed on other DBMSs (and not needed there).
 Workarounds should be used only when necessary. For portability and best
 performance, standard SQL should be prefered whenever possible.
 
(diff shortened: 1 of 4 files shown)


src/ports/greenplum/modules/compatibility/compatibility.sql_in
```diff
@@ -41,7 +49,22 @@ performance, standard SQL should be prefered whenever possible.
 
 */
 
-CREATE FUNCTION MADLIB_SCHEMA.internal_create_table_as(
+CREATE FUNCTION MADLIB_SCHEMA.create_schema_pg_temp()
```
cwelton, line 52: Given that the schema is created automatically why do you need this?

src/ports/greenplum/modules/compatibility/compatibility.sql_in
```diff
@@ -41,7 +49,22 @@ performance, standard SQL should be prefered whenever possible.
...
+    IF pg_my_temp_schema() = 0 THEN
+        -- The pg_temp schema does not exist, yet. Creating a temporary table
+        -- will create it.
+        EXECUTE 'CREATE TEMPORARY TABLE _madlib_temp_table AS SELECT 1;
+            DROP TABLE pg_temp._madlib_temp_table CASCADE;';
+    END IF;
+END;
+$$;
+
+CREATE FUNCTION MADLIB_SCHEMA.internal_create_empty_table_as(
```
cwelton: Why create _empty_ table?  This seems like a weird abstraction and none of the callers need this behavior.  Keeping it just "create_table_as" makes a lot more sense to me.  

Naming conventions are that internal functions are prefixed with "_", yes?

src/ports/greenplum/modules/compatibility/compatibility.sql_in
fschopp, line 52: The pg_temp schema is only created automatically by creating the first temporary table. Before that, it might not exist.

We create a function in the temp schema, pg_temp._madlib_temp_function(), and referencing this function does not automatically create pg_temp.

src/ports/greenplum/modules/compatibility/compatibility.sql_in
fschopp: (My old comment was not quite to the point. Update:)
internal_create_empty_table_as() is the difference between create_table_as() and insert_into(). And we do want to expose insert_into() to the user. In order to avoid redundant code, I think there is good reason to factor out internal_create_empty_table_as() from create_table_as(), and implement the latter in terms of internal_create_empty_table_as() and insert_into().

Regarding the "_" prefix: I do not like it for SQL functions because UDFs written in C either would have to follow a different naming scheme, or the name of the symbol exported by the dynamic library would have to be different than the SQL name. (The actual problem is that the current naming suggestions are still somewhat a draft version. I propose the prefix "internal_" for private SQL or C functions)

src/ports/postgres/modules/regress/logistic.sql_in
```diff
@@ -427,3 +427,26 @@ CREATE FUNCTION MADLIB_SCHEMA.logregr(
...
+ * In order for the outcome of exp(x) to be within the range of the minimum
+ * positive double-precision number (i.e., 2^(-1074)) and the maximum positive
+ * double-precision number (i.e., (1 + (1 - 2^52)) * 2^1023), x has to be within
+ * the natural logarithm of these numbers, so roughly in between -744 and 709.
+ * However, 1 + exp(x) will just evaluate to 1 if exp(x) is less than the
+ * machine epsilon (i.e., 2^(-52)) or, equivalently, if x is less than the
+ * natural logarithm of that; i.e., in any case if x is less than -37.
+ * Note that taking the reciprocal of the largest double-precision number will
+ * not cause an underflow. Hence, no further checks are necessary.
+ */
```
cwelton, line 443: Not related to the other changes.  I encourage keeping logically separate changes in separate merge requests to facilitate issue tracking.

src/ports/greenplum/modules/compatibility/compatibility.sql_in
cwelton: For the naming conventions my main comment is that we should be consistent and right now others on the team are using "_" per the draft document.  If this is something that you think we need to discuss then we should review the document and discuss anything that we do not agree with.

fschopp: Re: Your first post.
Thanks for your comment and your persistence. I could not find a reason any more for having internal_create_empty_table_as() instead of internal_create_table_as(). So internal_create_empty_table_as() is history now.

fschopp: Re: Your second post.
Yes, we agree here. The naming conventions need an overhaul.

cwelton:
Looks good.  Pulled.

PR #87, opened 2011-12-21
Integrate modifications to C++ abstraction layer

fschopp:
Major design change in C++ abstraction layer:
- Moved from a design based on abstract base classes to a policy-based design (the C++ AL is now essentially a template and header-only library)
- Advantages:
  - Remove as much overhead as possible
  - Simpler design (though possibly syntactically a bit harder)
  - Less code to maintain
  - Much more straihtforward control flow
  - No unnecessary object creation. Completely got rid of using shared_ptr
- Disadvantages:
  - The dbconnector part is now also header-only and it needs to include the postgres headers. Hence, there is the potential of name clashes (though it is possible to circumvent these problems).
  - Slightly higher memory usage because of template code. Probably irrelevant in practice though.


CMakeLists.txt
```diff
@@ -75,13 +79,18 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "SunPro")
...
+        OUTPUT_VARIABLE GNUCXX_VERSION)
+
     # We need the 1998 standard plus amendments (ISO/IEC 14882:2003) plus TR1
     # Unfortunately, we only get this with gnu++98
     # Special notes:
     # - long long is not part of the C++ 1998/2003 standard, but it is such a
     # common (and useful) extension that we do not want to hear warnings about
     # it.
-    set(CMAKE_CXX_FLAGS "-std=gnu++98 -Wall -Wextra -pedantic -Wno-long-long" CACHE STRING
+    # FIXME: -msse2 should only be added on i386
```
cwelton, line 92: Please add jiras for any outstanding items.

src/CMakeLists.txt
```diff
@@ -33,26 +36,26 @@ if(NOT BOOST_TAR_SOURCE)
     set(BOOST_TAR_SOURCE ${BOOST_URL})
 endif (NOT BOOST_TAR_SOURCE)
 
-# We always download Armadillo (unless it is already present in
+# We always download Eigen (unless it is already present in
```
cwelton, line 39: I don't see any changes to the licenses/ directory as part of this pull request, but we are changing our list of dependencies.

Please remove any licenses that we are no longer referencing and add any new licenses due to the dependency on Eigen.

src/CMakeLists.txt
```diff
@@ -121,67 +124,42 @@ else(Boost_FOUND)
...
-# BLAS_NAMES and LAPACK_NAMES are variables used by the Armadillo
-# build script
-if(SOLARIS)
-    list(APPEND ARMADILLO_CMAKE_ARGS
-        -DBLAS_NAMES=sunperf
-        -DLAPACK_NAMES=sunperf
+string(REGEX MATCH "^[a-zA-Z]+" _EIGEN_PROTOCOL ${EIGEN_TAR_SOURCE})
+if(_EIGEN_PROTOCOL STREQUAL "https")
+    # FIXME: This is somewhat ugly
+    # Unfortunately, CMake does not come with SSL support for file downloads.
```
cwelton, line 136: If this is something you think we should address then a jira should be filed.

src/CMakeLists.txt
```diff
@@ -121,67 +124,42 @@ else(Boost_FOUND)
...
+else(_EIGEN_PROTOCOL STREQUAL "https")
+    set(_EIGEN_SOURCE
+        URL ${EIGEN_TAR_SOURCE} URL_MD5 ${EIGEN_TAR_MD5}
     )
-endif(SOLARIS)
+endif(_EIGEN_PROTOCOL STREQUAL "https")
 
-ExternalProject_Add(EP_armadillo
-    DEPENDS ${ARMADILLO_DEPENDS}
+# FIXME: Eigen is a third-party source that is patched in-place. Other
```
cwelton, line 149: Do we support a user installed version of eigen, or must we always download it.  If the former is patching in place legitimate?

src/dbal/EigenLinAlgTypes/EigenLinAlgTypes.hpp
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @file EigenLinAlgTypes.hpp
+ *
+ *//* ----------------------------------------------------------------------- */
+
+#ifndef MADLIB_EIGEN_LINALGTYPES_HPP
+// Note: We also use MADLIB_EIGEN_LINALGTYPES_HPP for a workaround for Doxygen:
+// *_{impl|proto}.hpp files should be ingored if not included by
+// EigenLinAlgTypes.hpp
+#define MADLIB_EIGEN_LINALGTYPES_HPP
```
cwelton, line 11: Can you explain what's going on with doxygen here.  All this doxygen workaround is really gross.

src/dbal/EigenLinAlgTypes/SymmetricPositiveDefiniteEigenDecomposition_impl.hpp
```diff
@@ -0,0 +1,154 @@
...
+inline
+void
+EigenTypes<MapOptions>
+    ::SymmetricPositiveDefiniteEigenDecomposition<MatrixType>::computeExtras(
+    const MatrixType &inMatrix, int inExtras) {
+    
+    if (inExtras & ComputePseudoInverse) {
+        mPinv.resize(inMatrix.rows(), inMatrix.cols());
+
+        // FIXME: No hard-coded constant here
```
cwelton, line 123: Fix desirable.

src/ports/postgres/dbconnector/AnyType_impl.hpp
```diff
@@ -0,0 +1,584 @@
...
+
+    PG_TRY(); {
+        if (boost::indeterminate(ioTargetIsComposite))
+            ioTargetIsComposite = isRowTypeInCache(inTargetTypeID,
+                type_is_rowtype(inTargetTypeID));
+        
+        if (ioTargetIsComposite) {
+            // Don't ereport errors. We set typmod < 0, and this should
+            // not cause an error because compound types in another
+            // compound can never be transient. (I think)
```
cwelton, line 305: Good question. I'll have to think about this.

src/ports/postgres/dbconnector/AnyType_impl.hpp
```diff
@@ -0,0 +1,584 @@
...
+        // It seems not to be necessary to release the tupleDesc
+        // E.g., in plython.c ReleaseTupleDesc() is not called
+        funcClass = get_call_result_type(inFCInfo, &targetTypeID, &tupleDesc);
+    } PG_CATCH(); {
+        exceptionOccurred = true;
+    } PG_END_TRY();
+    
+    if (exceptionOccurred)
+        throw std::invalid_argument("An exception occurred while "
+            "gathering inormation about the PostgreSQL return type.");
```
cwelton, line 469: s/inormation/information

src/ports/postgres/dbconnector/AnyType_proto.hpp
```diff
@@ -0,0 +1,98 @@
...
+ * - A scalar value, which is just a PostgreSQL Datum
+ * - A function composite value, which is a virtual composite value consisting
+ *   of all function arguments
+ * - A native composite value, which is a PostgreSQL HeapTupleHeader
+ * - A return composite value, which is a vector of PostgreSQL datums
+ */
+class AbstractionLayer::AnyType {
+public:
+    AnyType();
+//    AnyType(const AnyType &inOriginal);
```
cwelton: Remove dead code

src/ports/postgres/dbconnector/TypeTraits.hpp
```diff
@@ -0,0 +1,106 @@
...
+    AbstractionLayer::MutableArrayHandle<double>,
+    dbal::ArrayType,
+    dbal::Mutable,
+    PointerGetDatum(value.array()),
+    reinterpret_cast<ArrayType*>(
+        needMutableClone
+          ? DatumGetArrayTypePCopy(value)
+          : DatumGetArrayTypeP(value))
+);
+
```
cwelton, line 101: Should add task to extend list of supported types to include more datatypes, including:
- text data types
- bytea
- arrays other than floats
- svecs
- etc

cwelton:
Wow.  Definitely a larger change than I was expecting.  I'm afraid my C++ isn't good enough to evaluate the template code very well.   

I don't see any glaring problems.

src/CMakeLists.txt
fschopp, line 149: We can work with a user-supplied Eigen tarball, but we never use a user-installed version of Eigen.

src/dbal/EigenLinAlgTypes/EigenLinAlgTypes.hpp
fschopp, line 11: The problem with Doxygen is this: We don't want the header files to be processed alone because they may lack the namespace environment -- i.e., if they are processed alone, the declared classes/functions/etc. appear at global scope in the documentation. Therefore, the header should be inlined into the files that include them. Unfortunately, this only works in Doxygen when the header files are also included in the set of source files. And that means that the header files are processed alone, too. In the end, we then end up with two descriptions for each documentation object -- an unwanted one in the global namespace, and the correct one in the proper namespace. A workaround is this #ifndef macro.

src/ports/postgres/dbconnector/AnyType_impl.hpp
```diff
@@ -0,0 +1,584 @@
...
+    Oid targetTypeID;
+    TupleDesc tupleDesc;
+    TypeFuncClass funcClass;
+    bool exceptionOccurred = false;
+
+    PG_TRY(); {
+        // FIXME: get_call_result_type is tagged as expensive in funcapi.c
+        // It seems not to be necessary to release the tupleDesc
+        // E.g., in plython.c ReleaseTupleDesc() is not called
+        funcClass = get_call_result_type(inFCInfo, &targetTypeID, &tupleDesc);
```
haradh1, line 462: So, what was your conclusion about fn_extra cache that I suggested earlier? Wasn't it sufficient for the solution?

src/ports/postgres/dbconnector/AnyType_impl.hpp
```diff
@@ -0,0 +1,584 @@
...
+    if (!targetIsComposite && isComposite())
+        throw std::logic_error("Invalid type conversion requested. "
+            "Composite type supplied but PostgreSQL expects simple type.");
+
+    // tupleDesc can be NULL if the return type is not composite
+    return getAsDatum(targetTypeID, isComposite(), tupleDesc);
+}
+
+/**
+ * @brief Return a PostbreSQL Datum representing the current object
```
haradh1: s/PostbreSQL/PostgreSQL/
LOL

src/dbal/EigenLinAlgTypes/EigenTypes_proto.hpp
```diff
@@ -0,0 +1,96 @@
...
+struct EigenTypes {
+protected:
+    template <
+        class EigenType,
+        bool EigenTypeIsConst = boost::is_const<EigenType>::value
+    > struct DefaultHandle;
+    
+    template <class EigenType>
+    struct DefaultHandle<EigenType, true> {
+        typedef AbstractionLayer::ArrayHandle<typename EigenType::Scalar> type;
```
haradh1, line 23: I know that "class" and "typename" in template are identical, but what's your rule of the usage of them? Maybe better if they're consistent?

src/ports/postgres/dbconnector/AnyType_impl.hpp
fschopp, line 462: After all the other changes, I tested linear regression with 100 independent variables and 1,000,000 rows. It was essentially as fast as a count(*). Therefore I held off implementing caching for get_call_result_type(). I will reconsider it once/if we have a case where get_call_result_type() really causes a slowdown.

src/dbal/EigenLinAlgTypes/EigenTypes_proto.hpp
fschopp, line 23: I used "class" whenever only a class (user-defined type) can be used, and "typename" if also primitive types are acceptable. As far as I remember, this is also Scott Meyer's suggestion in "Effective C++", 3rd ed., item 42.

src/CMakeLists.txt
fschopp, line 39: Done.

src/CMakeLists.txt
fschopp, line 136: No, adding SSL support to cmake is beyond the scope of MADlib. Let me know if you are aware of another non-https download link for Eigen.

src/ports/postgres/dbconnector/TypeTraits.hpp
fschopp, line 101: Absolutely. But as long as the C++ AL is/has been MADlib-internal only and not exposed for general use, I have only implemented on an as-needed basis.

PR #89, opened 2011-12-21
Integrate changes to logistic regression to conform to improved C++ abstraction layer

fschopp:
This also includes Aaron Feng's incremental-gradient optimizer.


src/ports/postgres/modules/regress/logistic.sql_in
```diff
@@ -383,8 +454,10 @@ BEGIN
     -- function in a subquery
     IF optimizer = 'irls' OR optimizer = 'newton' THEN
         fnName := 'internal_logregr_irls_result';
-    ELSE
+    ELSIF optimizer = 'cg' THEN
         fnName := 'internal_logregr_cg_result';
+	ELSE
+		fnName := 'internal_logregr_igd_result';
```
cwelton: 1. Indentation is a little funny here.
2. Shouldn't we have an optimizer name for igd as well?

src/modules/regress/logistic.cpp
```diff
@@ -626,14 +592,247 @@ AnyType LogisticRegressionIRLS::distance(AbstractDBInterface & /* db */,
...
+        X_transp_AX += inOtherState.X_transp_AX;
+        logLikelihood += inOtherState.logLikelihood;
+        return *this;
+    }
+            
+    /**
+     * @brief Reset the inter-iteration fields.
+     */
+    inline void reset() {
+		// FIXME: HAYING: stepsize if hard-coded here now
```
cwelton, line 694: Should figure out a long term strategy for the stepsize, should have jira.

PR #92, opened 2011-12-22
Rewrite Decision tree

aojwang:
The decision tree implementation will based on c4.5, including:
1. Supporting information gain, gain ratio and gini as split criterion 
2. Supporting both continuous attribute and discrete attribute
3. Supporting missing attribute
4. Supporting both tabular format input and sparse vector format input
5. Supporting reduce error pruning and error based pruning.
6. Make the training and classification scalable.

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.c
+++ b/methods/decision_tree/src/pg_gp/decision_tree.c
@@ -79,9 +79,9 @@ Datum ebp_calc_errors(PG_FUNCTION_ARGS)
 
     if (!is_float_zero(1 - conf_level))
     {
-		if (!(conf_level >= MIN_CONFIDENCE_LEVEL &&
-              conf_level <= MAX_CONFIDENCE_LEVEL))
-       		elog(ERROR,"invalid conf_level: %lf",conf_level);
+    	if (!(conf_level >= MIN_CONFIDENCE_LEVEL &&
+                    conf_level <= MAX_CONFIDENCE_LEVEL))
+            elog(ERROR,"invalid conf_level: %lf",conf_level * 100);
 
 		/* calculate the coeff */
 		while (conf_level > CONFIDENCE_LEVEL[i]) i++;
@@ -128,10 +128,6 @@ float ebp_calc_errors_internal
     if (num_errors < 0)
         elog(ERROR, "num_errors is less than zero");
 
-	if (!(conf_level >= MIN_CONFIDENCE_LEVEL && 
-                conf_level <= MAX_CONFIDENCE_LEVEL))
-        elog(ERROR,"invalid conf_level: %lf",conf_level);
-
     if(coeff <= 0)
         elog(ERROR,"invalid coeff: %lf",coeff);
 
@@ -165,31 +161,6 @@ float ebp_calc_errors_internal
 }
 
 /*
- * This function allocate int64 array.
- * Parameters:
- *       size:    the size of array.
- *       value:   the initial value. All the
- *                elements of that array are initialized
- *                with that value.
- * Return:
- *      The int64 array allocated on current memory context.
- */
-
-static int64* alloc_int64_array(int size, int value)
-{
-    if(size <= 0)
-        elog(ERROR,"invalid array size:%d",size);
-
-    int64 *result   = (int64*)palloc(sizeof(int64) * size);
-
-    if (!result)
-        elog(ERROR, "memory allocation failure");
-
-    memset(result, value, sizeof(int64) * size);
-    return result;
-}
-
-/*
  * The step function for aggregating the class counts while doing Reduce Error Pruning (REP).
  *
  * Parameters:
@@ -215,6 +186,9 @@ Datum rep_aggr_class_count_sfunc(PG_FUNCTION_ARGS)
     bool need_reconstruct_array     = false;
 
 
+    if (max_num_of_classes < 2)
+    	elog(ERROR, "the number of classes: (%d) should be greater than 2", max_num_of_classes);
+
     if (!(original_class > 0 && 
           original_class <= max_num_of_classes))
         elog(ERROR,"original_class:%d,max_num_of_classes:%d",
@@ -232,7 +206,7 @@ Datum rep_aggr_class_count_sfunc(PG_FUNCTION_ARGS)
     	 * We assume the maximum number of classes is limited (up to millions),
     	 * so that the allocated array won't break our memory limitation.
     	 */
-        class_count_data		= alloc_int64_array(max_num_of_classes + 1, 0);
+        class_count_data		= palloc0(sizeof(int64) * max_num_of_classes + 1);
         array_length 			= max_num_of_classes + 1;
         need_reconstruct_array 	= true;
 
@@ -278,14 +252,12 @@ Datum rep_aggr_class_count_sfunc(PG_FUNCTION_ARGS)
         	construct_array(
         		(Datum *)class_count_data,
                 array_length,
-                FLOAT8OID,
+                INT8OID,
                 sizeof(int64),
                 true,
                 'd'
                 );
 
-        if(!class_count_array)
-            elog(ERROR, "array construction failure");
     }
     PG_RETURN_ARRAYTYPE_P(class_count_array);
 }
@@ -422,21 +394,18 @@ Datum rep_aggr_class_count_ffunc(PG_FUNCTION_ARGS)
     	construct_array(
          (Datum *)result,
          2,
-         FLOAT8OID,
+         INT8OID,
          sizeof(int64),
          true,
          'd'
         );
-
-    if(!result_array)
-        elog(ERROR, "array construction failure");
 
     PG_RETURN_ARRAYTYPE_P(result_array);
 }
 PG_FUNCTION_INFO_V1(rep_aggr_class_count_ffunc);
 
 /*
- * We use a 15-element array to keep the state of the
+ * We use a 14-element array to keep the state of the
  * aggregate for calculating splitting criteria values (SCVs).
  * The enum types defines which element of that array is used
  * for which purpose.
@@ -487,11 +456,6 @@ enum SCV_STATE_ARRAY_INDEX
      */
     SCV_STATE_TRUE_TOTAL_COUNT,
     /*
-     * 1 We are computing initial entropy before split.
-     * 0 We are computing the gain for certain split.
-     */ 
-    SCV_STATE_IS_CALC_PRE_SPLIT,
-    /*
      * The ID of the class with the most elements.
      */ 
     SCV_STATE_MAX_CLASS_ID,
@@ -500,7 +464,7 @@ enum SCV_STATE_ARRAY_INDEX
 };
 
 /*
- * We use a 12-element array to keep the final result of the
+ * We use a 11-element array to keep the final result of the
  * aggregate for calculating splitting criteria values (SCVs).
  * The enum types defines which element of that array is used
  * for which purpose.
@@ -524,11 +488,6 @@ enum SCV_FINAL_ARRAY_INDEX
     /* whether the selected feature is continuous or discrete*/
     SCV_FINAL_IS_CONT_FEATURE,
     /*
-     * 1 We are computing initial entropy before split.
-     * 0 We are computing the gain for certain split.
-     */     
-    SCV_FINAL_CALC_PRE_SPLIT,
-    /*
      * The ID of the class with the most elements.
      */     
     SCV_FINAL_CLASS_ID,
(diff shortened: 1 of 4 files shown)


methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,902 @@
...
+                true,
+                'd'
+                );
+
+        if(!class_count_array)
+            elog(ERROR, "Array construction failure.");
+    }
+    else
+    {
+        class_count_array = PG_GETARG_ARRAYTYPE_P(0);
```
cwelton:
``` C
if (fcinfo->context && IsA(fcinfo->Context, AggState))
  class_count_array = PG_GETARG_ARRAYTYPE_P(0);
else
  class_count_array = PG_GETARG_ARRAYTYPE_P_COPY(0);
```

You must always make sure to guard pass by reference parameters and never update them in place unless you have guaranteed that you are in an aggregate context and it is the first parameter.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,902 @@
...
+ *      class_count_array   The array used to store the accumulated information.
+ *                          [0]: the total number of mis-classified cases
+ *                          [i]: the number of cases belonging to the ith class
+ *      classified_class    The predicted class based on our trained DT model.
+ *      original_class      The real class value provided in the validation set.
+ *      max_num_of_classes  The total number of distinct class values.
+ * Return:
+ *      An updated state array.
+ */
+Datum rep_aggr_class_count_sfunc(PG_FUNCTION_ARGS)
```
cwelton, line 219: I thought we weren't going to build arrays of number of distinct values anymore?  This is not a scalable solution.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,902 @@
...
+        array_dim2              = ARR_NDIM(class_count_array2);
+        p_array_dim2            = ARR_DIMS(class_count_array2);
+        array_length2           = ArrayGetNItems(array_dim2,p_array_dim2);
+        class_count_data2       = (int64 *)ARR_DATA_PTR(class_count_array2);
+
+        if (array_length != array_length2)
+            elog(ERROR, "The size of the two arrays must be the same.");
+
+        for (int index =0; index < array_length; index++)
+            class_count_data[index] += class_count_data2[index];
```
cwelton, line 408: Same as above, you can't update pass by reference in place unless you have validated that the function is called in an aggregate context.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,902 @@
...
+	float total_cases,
+	float num_errors,
+	float conf_level,
+	float coeff
+    )
+{
+	Assert(total_cases > 0);
+	Assert(num_errors >= 0);
+	Assert(conf_level >= MIN_CONFIDENCE_LEVEL && conf_level <= MAX_CONFIDENCE_LEVEL);
+	Assert(coeff >= 0);
```
cwelton: Asserts are compiled out of optimized code.  You have not protected this code from bad user input.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+ */
+
+static int64* alloc_int64_array(int size, int value)
+{
+    if(size <= 0)
+        elog(ERROR,"invalid array size:%d",size);
+
+    int64 *result   = (int64*)palloc(sizeof(int64) * size);
+
+    if (!result)
```
haradh1: palloc never returns NULL. If you want zero-set array, use palloc0. Also most of negative size will be denied due to the limit of 1GB allocation inside palloc, as the negative size will be interpreted as a huge positive value by casting.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -0,0 +1,2030 @@
...
+$$ LANGUAGE PLPGSQL;
+
+/*
+ * @brief Format a string with the value in allvalues array
+ * @param format_str the format string
+ * @param allvalues the specified elements in format_str
+ *
+ * @note All the char ('%') in format_str will be replaced with allvalues
+ *       The number of '%' should be equal with the length of array allvalues
+ */
```
haradh1, line 74: I now know why format is necessary for DT. We wouldn't need format function if we went with pl/python instead of pl/pgsql all the way, which follows kmeans style. And I believe we agreed to use pl/python as far as possible.

methods/decision_tree/src/pg_gp/decision_tree.c
haradh1, line 219: I "guess" it assumes the number of distinct _class_ values is not so many. Building distinct feature values need to be scalable, though.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+        	construct_array(
+        		(Datum *)class_count_data,
+                array_length,
+                FLOAT8OID,
+                sizeof(int64),
+                true,
+                'd'
+                );
+
+        if(!class_count_array)
```
haradh1: construct_count_array() doesn't return NULL.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+    int64 *class_count_data         = (int64 *)ARR_DATA_PTR(class_count_array);
+    int64 *result                   = palloc(sizeof(int64)*2);
 
-	memset(result, value, sizeof(float8)*size);
-	
-	pgarray = construct_array((Datum *)result,
-		size, FLOAT8OID,
-		sizeof(float8),true,'d');
-    PG_RETURN_ARRAYTYPE_P(pgarray);
+    if (!result)
```
haradh1: Same as above.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+     * the root node of the subtree being processed
+     * class_count_data[0] the total number of mis-classified cases
+     */
+    result[1] = class_count_data[0] - (sum - max);
+
+    ArrayType* result_array =
+    	construct_array(
+         (Datum *)result,
+         2,
+         FLOAT8OID,
```
haradh1: Use INT8OID instead of FLOAT8OID

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+    ++class_count_data[original_class];
+
+    if( need_reconstruct_array )
+    {
+        /* construct a new array to keep the aggr states. */
+        class_count_array =
+        	construct_array(
+        		(Datum *)class_count_data,
+                array_length,
+                FLOAT8OID,
```
haradh1: Use INT8OID instead of FLOAT8OID

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+                sizeof(int64),
+                true,
```
haradh1, line 318: Whether INT8/FLOAT8 is byval or not is dependent on platform (in PostgreSQL; GPDB uses only true). See FLOAT8PASSBYVAL.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+    	elog(ERROR, "invalid aggregation data array");
+
+    float8 init_scv = scv_state_data[SCV_STATE_INIT_SCV];
+
+    int result_size = 12;
+    float8 *result = palloc(sizeof(float8) * result_size);
+    if (!result)
+    	elog(ERROR, "memory allocation failure");
+
+    memset(result, 0, sizeof(float8) * result_size);
```
haradh1, line 1209: Use palloc0, as above.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+    SCV_STATE_MAX_CLASS_ELEM_COUNT
+};
+
+/*
+ * We use a 12-element array to keep the final result of the
+ * aggregate for calculating splitting criteria values (SCVs).
+ * The enum types defines which element of that array is used
+ * for which purpose.
+ */
+enum SCV_FINAL_ARRAY_INDEX
```
haradh1, line 555: enum is not bad approach, but there could be another way to do this, namely to map structure to array data memory.

<pre>
SCVFin fin = (SCVFin) ARR_DATA_PTR(...)
</pre>

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,895 @@
...
+    }
+    else
+    if (num_errors < 0.9999)
+    {
+        float tmp = total_cases * (1 - exp(log(conf_level) / total_cases));
+        return tmp + num_errors * 
+            (ebp_calc_errors_internal(total_cases, 1.0, conf_level, coeff) - tmp);
+    }
+    else
+    if (num_errors + 0.5 >= total_cases)
```
haradh1, line 189: Sorry if I miss something (probably), but I couldn't find the background of this (and the line below) heuristic in the design document. Could you elaborate it?

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang: remove the line of code.

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang: Thanks. Done.

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang, line 1209: Thanks. Done.

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang, line 555: Thanks. we will keep it now, since it's up to the compiler how to layout the member variable in the structure.

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang: Changed according to the last code review. Thanks.

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang, line 189: It's from c45 to calculate the additional errors for a node. Details please refer to this page: http://books.google.com/books?id=AABrrnZWgPEC&pg=PA278&lpg=PA278&dq=Taken+from+Documenta+Geigy+Scientific+Tables+(Sixth+Edition)&source=bl&ots=z7WRjXXNYJ&sig=Ao-X5ey57Oofw26T1miBSZg1ea4&hl=en&sa=X&ei=Zk8GT8b0CoTy0gHgl_GRAg&ved=0CEQQ6AEwBA#v=onepage&q=Taken%20from%20Documenta%20Geigy%20Scientific%20Tables%20(Sixth%20Edition)&f=false

haradh1, line 189: Ah, great, thanks.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1046 @@
...
+ *  The coefficient is the square of the number of standard deviations
+ *  corresponding to the selected confidence level.
+ *  (Taken from Documenta Geigy Scientific Tables (Sixth Edition),
+ *  p185 (with modifications).)
+ */
+static float CONFIDENCE_LEVEL[] = {0, 0.001, 0.005, 0.01, 0.05, 0.10, 0.20, 0.40, 1.00};
+static float CONFIDENCE_DEV[]   = {4.0, 3.09, 2.58, 2.33, 1.65, 1.28, 0.84, 0.25, 0.00};
+
+#define MIN_CONFIDENCE_LEVEL 0.001
+#define MAX_CONFIDENCE_LEVEL 100.0
```
cwelton, line 48: By increasing MAX_CONFIDENCE_LEVEL to be a value larger than the largest value in CONFIDENCE_LEVEL you have _once again_ created conditions that will reference uninitialized memory.

See:
`while (conf_level > CONFIDENCE_LEVEL[i]) i++;`

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1046 @@
...
+
+#define check_conf_level(cf) \
+			cf = (cf) * 100; \
+			if ((cf) < MIN_CONFIDENCE_LEVEL || (cf) > MAX_CONFIDENCE_LEVEL) \
+				elog(ERROR,"invalid conf_level: %lf",cf);
```
cwelton: A) Use elog only for exceptional situations that shouldn't occur.  Use ereport for errors that can be caused by bad user input.
B) When giving an error about user input not in a valid range you should indicate what the valid range is as part of the errdetail.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1046 @@
...
+
+float ebp_calc_errors_internal
+(
+	float total_cases,
+	float num_errors,
+	float conf_level,
+	float coeff
+);
```
cwelton, line 72: C Data types:
float     - 4 bytes
double - 8 bytes on most platforms

Postgres Datatypes
float4   - 4 bytes
float8   - 8 bytes.

In ebp_calc_errors you are dealing with float8s which you then pass into this function, downcasting them to "float".  This is not appropriate, you should remain in 8 byte floating point.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1,254 +1,1046 @@
...
+ *  p185 (with modifications).)
+ */
+static float CONFIDENCE_LEVEL[] = {0, 0.001, 0.005, 0.01, 0.05, 0.10, 0.20, 0.40, 1.00};
+static float CONFIDENCE_DEV[]   = {4.0, 3.09, 2.58, 2.33, 1.65, 1.28, 0.84, 0.25, 0.00};
+
+#define MIN_CONFIDENCE_LEVEL 0.001
+#define MAX_CONFIDENCE_LEVEL 100.0
+
+#define check_conf_level(cf) \
+			cf = (cf) * 100; \
```
haradh1: It is surprising to me that this macro overwrites input value. Is that intentional?? I also saw

<pre>
ERROR:  invalid conf_level: 2500.000000 (decision_tree.c:137)
</pre>

for the shortest form of training API.

cwelton:
Filed several jiras on outstanding issues, including:
- MADLIB-327 - Decision trees break on GPDB 4.0 and Postgres 8.4
- MADLIB-326 - Decision Trees: bad inplace update of pass by reference values
- MADLIB-325 - Decision Trees: Downcasting of floating point datatypes
- MADLIB-324 - Decision Trees: Overrun of stack memory

All of these need to be addressed prior to release, but from a manageability perspective I'd like to get the current work in.

PR #93, opened 2011-12-27
python array for GPDB 4.2 and Postgres 9.1

yuj18:
make sure arrays are passed into python as strings.


cwelton:
That should work too.

PR #94, opened 2012-01-04
Add support to specify a temporary directory location for madpack.

cwelton:
Purpose: support artifact collection during pulse builds.

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -692,6 +692,9 @@ def main(argv):
     parser.add_argument('-l', '--keeplogs', dest='keeplogs', default=False, 
                          action="store_true", help="Do not remove installation log files.")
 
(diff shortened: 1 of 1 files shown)


src/madpack/madpack.py
```diff
@@ -692,6 +692,9 @@ def main(argv):
     parser.add_argument('-l', '--keeplogs', dest='keeplogs', default=False, 
                          action="store_true", help="Do not remove installation log files.")
 
+    parser.add_argument('-d', '--directory', dest='tmpdir', default = '/tmp/',
```
haradh1, line 695: I"d prefer names like --tempdir --tmpdir for long option.

agorajek, line 695: I tend to agree with Hitoshi and can change it to --tmpdir during the merge.

cwelton, line 695: Sounds good to me.

PR #95, opened 2012-01-05
Fix to array_agg non-support for GPDB<4.1 and Postgres<9.0.

agorajek:
This is to fix the recent install-check failures.

Caleb, Florian or Hitoshi - please review and merge (if looks ok).

### DIFF
--- a/methods/array_ops/src/pg_gp/array_ops.sql_in
+++ b/methods/array_ops/src/pg_gp/array_ops.sql_in
@@ -236,3 +236,14 @@ LANGUAGE C IMMUTABLE;
  */
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.normalize(x float8[]) RETURNS float8[]
 AS 'MODULE_PATHNAME', 'array_normalize' LANGUAGE C IMMUTABLE STRICT;
+
+/**
+ * @brief ARRAY_AGG aggregate for compatibility with GPDB < 4.1 and Postgres < 9.0
+ *        This is a slower solution than the built in array_agg that appears
+ *        in later GPDB and Postgres versions
+ */
+CREATE AGGREGATE MADLIB_SCHEMA.array_agg( anyelement) ( 
+   SFUNC     = array_append, 
+   STYPE     = anyarray 
+   m4_ifdef( `GREENPLUM',`, PREFUNC   = array_cat')
+);
--- a/src/ports/greenplum/CMakeLists.txt
+++ b/src/ports/greenplum/CMakeLists.txt
@@ -175,14 +175,26 @@ if(GREENPLUM_FOUND)
 
 # -- 6. Preprocess/copy all Python/SQL files -----------------------------------
 
-    set(M4_ARGUMENTS_GREENPLUM
-        "-DGREENPLUM"
-    )
+    # Make sure we support older Greenplum versions while still using the
+    # latest features of the newer ones
+    if(${_GP_VERSION} VERSION_LESS "4.1")
+        set(M4_ARGUMENTS_GREENPLUM_PRE_41
+            "-DGREENPLUM_PRE_41"
+        )
+        set(M4_ARGUMENTS_GREENPLUM
+            "-DGREENPLUM"
+        )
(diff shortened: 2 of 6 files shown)


methods/array_ops/src/pg_gp/array_ops.c
```diff
@@ -1251,3 +1251,26 @@ Datum array_normalize(PG_FUNCTION_ARGS)
...
+	
+	/* Read the argument */
+	if (fcinfo->context && IsA(fcinfo->context, AggState))
+		transarray = PG_GETARG_ARRAYTYPE_P(0);
+	else
+		transarray = PG_GETARG_ARRAYTYPE_P_COPY(0);
+		
+	PG_RETURN_ARRAYTYPE_P(transarray);
+	
+}
```
cwelton: This function doesn't really do anything.  You can omit it entirely.

methods/array_ops/src/pg_gp/array_ops.sql_in
```diff
@@ -236,3 +236,17 @@ LANGUAGE C IMMUTABLE;
...
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.normalize(x float8[]) RETURNS float8[]
 AS 'MODULE_PATHNAME', 'array_normalize' LANGUAGE C IMMUTABLE STRICT;
+
+/**
+ * @brief ARRAY_AGG aggregate final function.
+ */
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.array_agg_final( anyarray ) 
+RETURNS anyarray
+AS 'MODULE_PATHNAME' LANGUAGE C IMMUTABLE;
+
```
cwelton: A comment that this is only here to support historical versions would be appropriate.

Possibly even m4_ifdef so that in only gets defined on the historical versions that do not have this aggregate already defined.

This implementation will be substantially slower than the builtin version for building larger arrays.

methods/array_ops/src/pg_gp/array_ops.c
agorajek: Thanks, I totally forgot about this.

methods/array_ops/src/pg_gp/array_ops.sql_in
agorajek: Better comment added. I thought about defining the array_agg only for older versions but since we are not using version based dynamic code adjustments on the SQL side yet it would require a bit more madpack rewrite and I need to leave it for now.

I also realized that the M4 flags (e.g. GREENPLUM40 representing GPDB < 9.0) may be a bit confusing so I changed them to:
- GREENPLUM_PRE_41 for GDBP < 4.1
- POSTGRES_PRE_90 for Postgres < 9.0

agorajek:
Ready for another check and (hopefully) merge.

methods/array_ops/src/pg_gp/array_ops.sql_in
fschopp: Instead of introducing lots of new macros I suggest having a single version macro that can be checked directly.

src/ports/greenplum/CMakeLists.txt
```diff
@@ -175,14 +175,26 @@ if(GREENPLUM_FOUND)
 
 # -- 6. Preprocess/copy all Python/SQL files -----------------------------------
 
-    set(M4_ARGUMENTS_GREENPLUM
-        "-DGREENPLUM"
-    )
+    # Make sure we support older Greenplum versions while still using the
+    # latest features of the newer ones
+    if(${_GP_VERSION} VERSION_LESS "4.1")
+        set(M4_ARGUMENTS_GREENPLUM_PRE_41
```
fschopp, line 181: No need to add a new variable here.
The following is sufficient (and much preferred):
`set(M4_ARGUMENTS_GREENPLUM "-DGREENPLUM_PRE_41 -DGREENPLUM")`

fschopp, line 181: Even better, of course, add a version macro instead of macros for each version.

PR #97, opened 2012-01-10
Kmeans fix related to Linux builds

agorajek:
The real problem existed on all OSes, but on OSX it showed up as NaN while on Linux as error.
One more commit for doc cleanup and one for extra sanity checks.
Ready for review and merge.


cwelton:
Merged.

PR #98, opened 2012-01-12
add support for gpdb4.0, postgres8.4, and bug fix

aojwang:


cwelton:
When fixing bugs please list bugs fixed.  Ideally separate fixes should have separate pull requests.

From what I can tell this includes fixes for:
- MADLIB-324 - Decision Trees: Confusing naming convensions and lack of comments around CONFIDENCE levels.
- MADLIB-326 - Decision Trees: bad inplace update of pass by reference values
- MADLIB-317 - DT will create global table and it has privelege issue 
- MADLIB-328 - Display error for donation data set

It is useful to understand which changes were motivated by what issues.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -28,6 +28,11 @@
 #endif
 
 /*
+ * Postgres8.4 doesn't have such macro, so we add here
+ */
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(*(x)))
```
cwelton, line 34: Should only define if it doesn't already exist, otherwise differences in whitespace etc can cause issues.

```
#ifndef ARRAY_SIZE
#define ARRAY_SIZE(x) ...
#endif
```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -584,7 +579,45 @@ BEGIN
                             ROWS BETWEEN 1 FOLLOWING AND  UNBOUNDED FOLLOWING) 
                         ELSE 
                             NULL 
-                        END AS gt,
+                        END AS gt,';
+    ELSE
+    -- old db version does not support starting from 1 following/curr row.
```
cwelton, line 580: What version doesn't support this?  In my tests it works fine on GPDB 4.0.  Is this a Postgres issue?

``` sql
select x, 
  sum(x) over (order by x rows between unbounded preceding and unbounded following) - 
  sum(x) over (order by x rows between unbounded preceding and current row),
  sum(x) over (order by x desc rows between unbounded preceding and current row) - x, 
  sum(x) over (order by x desc rows between unbounded preceding and 1 preceding), 
  sum(x) over (order by x rows between 1 following and unbounded following) 
from generate_series(1,10) x order by x;
 x  | ?column? | ?column? | sum | sum 
----+----------+----------+-----+-----
  1 |       54 |       54 |  54 |  54
  2 |       52 |       52 |  52 |  52
  3 |       49 |       49 |  49 |  49
  4 |       45 |       45 |  45 |  45
  5 |       40 |       40 |  40 |  40
  6 |       34 |       34 |  34 |  34
  7 |       27 |       27 |  27 |  27
  8 |       19 |       19 |  19 |  19
  9 |       10 |       10 |  10 |  10
 10 |        0 |        0 |     |    
(10 rows)

select version();
                                                                                               version                                                                                               
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 8.2.14 (Greenplum Database 4.0.6.0 build 4) on i386-apple-darwin9.8.0, compiled by GCC i686-apple-darwin9-gcc-4.0.1 (GCC) 4.0.1 (Apple Inc. build 5493) compiled on Jul 29 2011 18:04:32
(1 row)
```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -1239,7 +1275,7 @@ BEGIN
 	    SELECT MADLIB_SCHEMA.__format
             (
             'SELECT s1.assigned_nid, 
-                MAX(ARRAY[s1.info_impurity, s1.fid, s1.split_value, s1.class_prob, s1.class_id, s1.total_size]::FLOAT8[]) as info
+                MAX(ARRAY[s1.info_impurity, s1.fid, coalesce(s1.split_value, ''NaN''::FLOAT8), s1.class_prob, s1.class_id, s1.total_size]::FLOAT8[]) as info
```
cwelton, line 1218: What issue is this fixing?

methods/decision_tree/src/pg_gp/sql/dec_trees.sql_in
```diff
@@ -562,8 +562,8 @@ begin
         validation_set,cont_features,null,'id',class_column, 100,missing_operation,30000,tree_dep,0,0,0);
    
 	score_accuracy = MADLIB_SCHEMA.c45_score(tree_name,score_set,'f');
-    -- For those two datasets, we should not get an accuracy lower than 0.95
-    IF (score_accuracy<0.95) THEN 
+    -- For those two datasets, we should not get an accuracy lower than 0.9
+    IF (score_accuracy<0.9) THEN
```
cwelton, line 566: Why did you have to lower the accuracy here?  The new DT should be getting better accuracy than the old one.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -19,51 +19,58 @@ END
 $$ LANGUAGE PLPGSQL;
 
 
+-- We need create cast from boolean to char if db do not support it.
+-- postgresql 8.4 and beyond already support it.
+m4_ifdef(`GREENPLUM',
+         `DROP CAST IF EXISTS (BOOLEAN AS TEXT) CASCADE;');
```
cwelton: This is bad.  If this cast exists then chances are decent that madlib doesn't own it - we should not assume that we can drop it.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -19,51 +19,58 @@ END
...
     (
-    val BOOLEAN
+    val FLOAT8
     );
     
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__to_char
     (
-    val BOOLEAN
+    val FLOAT8
     )
```
cwelton: This is strange.  Why do we need to cast floats to text?

methods/decision_tree/src/pg_gp/decision_tree.c
aojwang, line 34: Thanks. Done.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang, line 1218: GPDB4.0 has such problem, For discrete attribute, the split_value is null. For continuous attribute it should have a value. Therefore, I use 'NaN' to represent it for discrete attribute.  please see below:

``` javascript
madlib=# create temp table t2(id int, c1 int) distributed by(id);
CREATE TABLE
madlib=# insert into t2 values(1, 2),(1,null);
INSERT 0 2
madlib=# select max(array[id, c1]) from t2;
ERROR:  null array element where not supported (arrayfuncs.c:872)

```

methods/decision_tree/src/pg_gp/utility.sql_in
aojwang: Since GPDB can't cast boolean to text, we provide a series unified functions to cast int, float, and boolean value to text.

``` javascript
madlib=# select 't'::boolean::text;                                                                                                          
ERROR:  cannot cast type boolean to text
LINE 1: select 't'::boolean::text;
                             ^

```

cwelton: Is this just for "display tree" or is there any other purpose to needing to cast to text?

methods/decision_tree/src/pg_gp/decision_tree.sql_in
renyi533, line 580: Postgres 8.4 has such problem. OK, I will make change so that only postgres 8.4 use this work around.

methods/decision_tree/src/pg_gp/sql/dec_trees.sql_in
renyi533, line 566: I added new test cases. For those additional tests, the accuracy is less than 0.95. 

Actually, we rely on the comparison of displayed tree to ensure the training is OK.

This accuracy check only ensures score function.

methods/decision_tree/src/pg_gp/utility.sql_in
renyi533: HI Caleb,

For greenplum, I tested that 4.2 still do not support cast from boolean to text. So I add one for displaying tree.

I can remove that line of code. If the cast is created in madlib. The statement below will drop it anyway.

DROP FUNCTION IF EXISTS MADLIB_SCHEMA.__to_char
    (
    val BOOLEAN
    ) CASCADE;

But that means 'm4_ifdef(`GREENPLUM',
`CREATE CAST (BOOLEAN AS TEXT) WITH FUNCTION MADLIB_SCHEMA.__to_char(BOOLEAN) AS IMPLICIT;');' can potentially fail.

Do you have any good solution?

Ren Yi

cwelton: Don't create a cast?

Where do you call the cast from?  Perhaps we can address the issue there?

cwelton: Alternately have a function "__create_cast_bool_as_text" which first checks if the cast exists and then creates it if it doesn't.

This query will tell you if the cast exists or not.

```
SELECT count(*) = 1 as cast_bool_as_text_exists from pg_cast where castsource = 'bool'::regtype and casttarget = 'text'::regtype;
```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
cwelton, line 1218: Ah... right.  

Funny enough that bug got introduced because of optimizations made to support sparse vectors (e.g. because of madlib).  

I'm glad we fixed it in 4.1, but that never was fixed in the 4.0 codeline.

methods/decision_tree/src/pg_gp/utility.sql_in
renyi533: OK. We will not create a cast. I can handle this in the display tree function.

aojwang:
The second commit includes below changes:
1) Changes according to Caleb's comments.
2) madlib-310: Too long output information when executing madlib decision tree functions
3) madlib-328: Display error for donation data set
4) madlib-330: Score and Classify function can not work well on dataset which class is null.
5) madlib-332: Train can not work well on special byte value

The third commit includes the changes for c45_genrule in decision_tree.sql_in.
The reason is that PostgresSQL can't recognize the type of a text value if we didn't cast it explicitly, for example:
the type of result after executing "select 'abc'" is unknown
the type of result after executing "select 'abc'::text" is TEXT.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -998,7 +1030,7 @@ RETURNS void AS $$
 DECLARE
 BEGIN
     -- For greenplum, use olap extension: group sets
-    IF (dbid = 10) THEN
+    IF (dbid = 1) THEN
```
cwelton: In your comments above you say that greenplum is dbid = 0, this seems to contradict that.

Wouldn't it be cleaner to use the m4 macros rather than pass this in as a function parameter?

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -1,69 +1,39 @@
...
+ *
+ * @note Greenplum doesn't support boolean to text casting
  */
 DROP FUNCTION IF EXISTS MADLIB_SCHEMA.__to_char
     (
-    val BIGINT
+    val anyelement
     );
-    
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__to_char
```
cwelton, line 11: Other datatypes that do not have conversions to text include:
- bytea
- xml
- point, lseg, path, box, polygon, line, circle
- abstime, reltime,
- bit, varbit
- array types
- madlib.svec

For any of those datatypes this function will fail.  I cannot tell if this is a problem or not.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -1353,6 +1418,70 @@ BEGIN
...
+ * @return true if the table has unsupported type, otherwise return false 
+ *
+ * NOTE: All the supported type are hardcoded here     
+ */
+DROP FUNCTION IF EXISTS MADLIB_SCHEMA.has_unsupported_type
+    (
+    full_table_name  TEXT
+    );
+
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.has_unsupported_type
```
cwelton: Inappropriate function name.  Does not indicate _what_ the type is not supported for.

Is this function intended to be exposed?

Suggest either:
  c45_validate_input_table(), or
  __c45_validate_input_table()

depending on exposure.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -1444,6 +1573,22 @@ DECLARE
...
+    PERFORM MADLIB_SCHEMA.__assert
+        (
+        NOT MADLIB_SCHEMA.has_unsupported_type
+            (
+            MADLIB_SCHEMA.__get_schema_name(input_table_name)   ||
+            '.'                                                 ||
+            MADLIB_SCHEMA.__strip_schema_name(input_table_name)
+            ),
+            'We only support the following types, including 
+             SMALLINT,INT, BIGINT, FLOAT8, REAL, NUMERIC,
```
cwelton: > We only support ... 

Language is too informal.  Error messages should not refer to personal pronouns.

Suggested alternative:
-  "Unsupported data type [%].  Supported types include: "

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -1601,7 +1746,23 @@ DECLARE
...
+
+    PERFORM MADLIB_SCHEMA.__assert
+        (
+        NOT MADLIB_SCHEMA.has_unsupported_type
+            (
+            MADLIB_SCHEMA.__get_schema_name(input_table_name)   ||
+            '.'                                                 ||
+            MADLIB_SCHEMA.__strip_schema_name(input_table_name)
+            ),
+            'We only support the following types, including
```
cwelton: Same comment as above regarding error message.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -2087,14 +2279,18 @@ BEGIN
 		      EXECUTE curstmt;
 		      
             ELSE
-                SELECT MADLIB_SCHEMA.__format(
-                    'UPDATE MADLIB_SCHEMA.% SET key=% WHERE %=''%'';',
+                -- we call "replace" function twice to escapethe string "'" and "\"
```
cwelton: s/escapethe/escape the/

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang: Thanks. Will call them separately. 

``` javascript
        IF (m4_ifdef(`GREENPLUM',`1',`0') = 1) THEN
            PERFORM MADLIB_SCHEMA.__generate_training_instance_greenplum
                (
                training_table_name,
                training_table_meta,
                'training_instance',
                'instance_selection',
                verbosity
                );        
        ELSE
            PERFORM MADLIB_SCHEMA.__generate_training_instance_postgres
                (
                training_table_name,
                training_table_meta,
                'training_instance',
                'instance_selection',
                verbosity
                );        
        END IF;
```

methods/decision_tree/src/pg_gp/utility.sql_in
aojwang, line 11: Thank you for pointing it out. maybe we can remove them from the the supported data type, let's leave it right now, what do you think.
currently supported data types:

``` javascript
SMALLINT,INT, BIGINT, FLOAT8, REAL, NUMERIC,
 INET, CIDR, MACADDR, 
 BIT, BIT VARYING, BOOLEAN, BYTEA,
MONEY, CHAR, VARCHAR, TEXT, "char", 
DATE, TIME, TIMETZ, TIMESTAMP, TIMESTAMPTZ, and INTERVAL
```

methods/decision_tree/src/pg_gp/utility.sql_in
aojwang: Thanks.  Changed to _validate_input_table(). It should not be exposed.
All the private functions were not add the c45 prefix, so we don't want to add the prefix for the function. thanks.

methods/decision_tree/src/pg_gp/utility.sql_in
aojwang: Thanks. Done.

``` javascript
        RAISE EXCEPTION 'Unsupported data type [%]. Supported types include:,
                         SMALLINT,INT, BIGINT, FLOAT8, REAL, NUMERIC,
                         INET, CIDR, MACADDR, 
                         BIT, BIT VARYING, BOOLEAN, BYTEA,
                         MONEY, CHAR, VARCHAR, TEXT, "char", 
                         DATE, TIME, TIMETZ, TIMESTAMP, TIMESTAMPTZ, and INTERVAL', 
                         rec.atttypid::regtype;
```

methods/decision_tree/src/pg_gp/utility.sql_in
aojwang: Done. Thanks.

aojwang:
The 4th commit includes the changes according to Caleb's comments.

methods/decision_tree/src/pg_gp/utility.sql_in
cwelton, line 11: Oh, yeah, I left out money.

Of those datatypes MONEY, BIT, BIT VARYING and BYTEA don't have text casts.

methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -1353,6 +1418,84 @@ BEGIN
...
+            
+    IF (rec IS NOT NULL) THEN
+        -- Print the first unsupported data type, and supported types.
+        RAISE EXCEPTION 'Unsupported data type [%]. Supported types include:,
+                         SMALLINT,INT, BIGINT, FLOAT8, REAL, NUMERIC,
+                         INET, CIDR, MACADDR, 
+                         BIT, BIT VARYING, BOOLEAN, BYTEA,
+                         MONEY, CHAR, VARCHAR, TEXT, "char", 
+                         DATE, TIME, TIMETZ, TIMESTAMP, TIMESTAMPTZ, and INTERVAL', 
+                         rec.atttypid::regtype;
```
cwelton, line 1492: Minor issues: extra comma, missing space
- s/include:,/include:/
- s/SMALLINT,INT/SMALLINT, INT/

aojwang, line 1492: Done. Thanks.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -335,7 +335,7 @@ AGGREGATE MADLIB_SCHEMA.__scv_aggr
     ) 
 (
   SFUNC=MADLIB_SCHEMA.__scv_aggr_sfunc,
-  m4_ifdef( `GREENPLUM_PRE_41',`prefunc=MADLIB_SCHEMA.__scv_aggr_prefunc,')
+  m4_ifdef(`GREENPLUM_PRE_41',`prefunc=MADLIB_SCHEMA.__scv_aggr_prefunc,')
```
fschopp, line 338: Why is the PREFUNC only for versions < 4.1?

renyi533, line 338: For gpdb 4.0, it still does not support ordered aggregate. Thus, we need a prefunc.

For versions beyond 4.0, we use ordered aggregate.

fschopp, line 338: Question: Is the prefunc ever called when used as an aggregate-derived window function? I remember cases where the prefunc has to be specified, but is otherwise not used at all.

PR #100, opened 2012-01-16
Resolve Caleb's comments

renyi533:
The only difference between this one and the request from jianwang is that I removed 'money', 'bit varying', 'bit', and 'bytea' from our supported type list. All other functions are the same.

Since Jianwang is in vacation, he cannot send out a pull request. As a result, I send this one out. Please merge this to main.

As the macro's from Florian still have some problem, I still do not use those macros. We can change them to new macros after solving the known issues.


renyi533:
Even there is no GREENPLUM_PRE_41/POSTGRES_PRE_90 support, current code can run on GREENPLUM 4.1/4.2 and POSTGRES 9.0/9.1. 
With the new MACRO's the code cannot run at all. Thus, I still use the old macros.
We can use the new one after solving the issues.

fschopp:
I don't see problems with the macros. Please elaborate.

fschopp:
Some other comments:
- Is this a superset of pull request 98? 
- Is there a good reason why you merged everything into one commit?
- The changes are significant (at least in size), so it is particularly important to give a meaningful commit message. In other words: Ideally, another person in a year of time should at least have a general idea of what the changes were about.

renyi533:
I will send you emails to clarify the macro's issue.

This is a superset of pull request 98. Jianwang's in vacation. And we have no way to update pull request 98 now. That's why I submitted a new pull request.

OK. Maybe, we should still use pull request 98. There are many review comments here. Once you merged pull request 98 in, I will send our a new small pull request that only include the change for the minor issue of supported list. What's your opinion?

fschopp:
Why did you not pull the changes from Jianwang into your fork, and add your changes from there?

renyi533:
I am very new to github. :)

I asked colleagues here. They also do not know how to do this. Can you give me some instructions?

renyi533:
Actually, I copied the code from Jianwang's branch and changed incrementally. So this pull request is a superset.

fschopp:
The usual thing to do would have been something like:

```
git fetch git://github.com/aojwang/madlib.git decision_tree # Retrieve Jianwang's DT code
git checkout -b my_new_branch FETCH_HEAD # Checkout the tip of the DT branch and start from there
```

This would create a new branch 'my_new_branch' in the fork.

By just copying Jianwang's code you essentially rewrote history. That is sometimes useful (e.g., to combine changes into logical units) but it would usually be done using `git rebase`. In any case, history should not be rewritten if it has already propagates to third parties (i.e., when others have already pulled the changes).

For now I suggest:
- Reapply your changes, this time on top of Jianwang's DT branch. This pull request should be deleted because it make tracking pull request 98 unnecessarily hard.
- Get familiar with git, e.g., by reading http://book.git-scm.com/

renyi533:
Hi Florian,

I noted that you still do not accept pull request 98. I will be in vacation tomorrow. So I will not be able to send new pull request.

Compared with pull request 98, this request of 100 only changes utility.sql_in. And the changes are really minor. They are for two minor CRs, madlib-332 and madlib-339. 

After accepting pull request 98, could you please merge utility.sql_in of this pull request to the main? You can ignore the other three files, including decision_tree.c, decision_tree.sql_in and sql/dec_trees.sql_in.

I am very sorry for the inconveniences for you.

Best regards,
Ren Yi

fschopp:
Pull request has been superseded by 101.

PR #104, opened 2012-01-27
k-means: Fixes for MADLIB-349, MADLIB-365, and others

fschopp:
This is a fix for the following k-means issues:

[MADLIB-349](http://jira.madlib.net/browse/MADLIB-349):
- All pgsql code converted to C code
- array_agg is now only used on "small" relations (it's use induced GroupAggregates which were problematic).
- The initial canopy assignment is now done using an aggregate function where it used to be an expensive cross join between points and centroids (worse, we used to perform a GroupAggregate and not a HashAggregate on this cross product)
- Distance computation now entirely in C functions (in the case of cosine sim. and Tanimoto, the _distance_ caluclation consisted of calling a C function plus some transformation in pgsql).

[MADLIB-366](http://jira.madlib.net/browse/MADLIB-366):
- For canopy clustering, we now assume T1>T2 like in the original paper

Others:
- Fixed bug in canopy merge function where a canopy could be removed if the distance to other canopies was less than .5 \* T1. This could lead to points disappearing (if their distance to the closest canopy was > T2 afterwards, which could happen after just one merge step if T1 > 1.5 \* T2)
- Tanimoto "distance" is now "1 - Tanimoto similarity" and not "acos(Tanimoto similarity)" any more

Note:
This pull request fixes the most urgent issues. It does not address other concerns (of which the author of these lines has many).

### DIFF
--- a/methods/kmeans/src/pg_gp/kmeans.c
+++ b/methods/kmeans/src/pg_gp/kmeans.c
@@ -1,0 +1,186 @@
+#include <postgres.h>
+#include <utils/builtins.h>
+#include "../../../svec/src/pg_gp/sparse_vector.h"
+#include "../../../svec/src/pg_gp/operators.h"
+
+typedef enum {
+    L1NORM = 1,
+    L2NORM,
+    COSINE,
+    TANIMOTO
+} KMeansMetric;
+
+static
+inline
+int
+verify_arg_nonnull(PG_FUNCTION_ARGS, int inArgNo)
+{
+    if (PG_ARGISNULL(inArgNo))
+        ereport(ERROR,
+                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                 errmsg("function \"%s\" called with NULL argument",
+                    format_procedure(fcinfo->flinfo->fn_oid))));
+    return inArgNo;
+}
+
+static
+inline
+void
+get_svec_array_elms(ArrayType *inArrayType, Datum **outSvecArr, int *outLen)
+{
+    deconstruct_array(inArrayType,           /* array */
+                  ARR_ELEMTYPE(inArrayType), /* elmtype */
+                  -1,                        /* elmlen */
+                  false,                     /* elmbyval */
+                  'd',                       /* elmalign */
+                  outSvecArr,                /* elemsp */
+                  NULL,                      /* nullsp -- pass NULL, because we 
+                                                don't support NULLs */
+                  outLen);                   /* nelemsp */
+}
+
+static
+inline
+double
+svec_svec_distance(Datum inVec1, Datum inVec2, KMeansMetric inMetric)
+{
+    PGFunction metric_fn = NULL;
+
+    switch (inMetric)
+    {
+        case L1NORM: metric_fn = svec_svec_l1norm; break;
+        case L2NORM: metric_fn = svec_svec_l2norm; break;
+        case COSINE: metric_fn = svec_svec_angle; break;
+        case TANIMOTO: metric_fn = svec_svec_tanimoto_distance; break;
+        default:
+            ereport(ERROR,
+                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                 errmsg("invalid metric")));
+    }
+    
+    return DatumGetFloat8(DirectFunctionCall2(metric_fn, inVec1, inVec2));    
+}
+
+PG_FUNCTION_INFO_V1(internal_get_array_of_close_canopies);
+Datum
+internal_get_array_of_close_canopies(PG_FUNCTION_ARGS)
+{
+    SvecType       *svec;
+    Datum          *all_canopies;
+    int             num_all_canopies;
+    float8          threshold;
+    KMeansMetric    metric;
+    
+    ArrayType      *close_canopies_arr;
+    int4           *close_canopies;
+    int             num_close_canopies;
+    size_t          bytes;
+    
+    svec = PG_GETARG_SVECTYPE_P(verify_arg_nonnull(fcinfo, 0));
+    get_svec_array_elms(PG_GETARG_ARRAYTYPE_P(verify_arg_nonnull(fcinfo, 1)),
+        &all_canopies, &num_all_canopies);
+    threshold = PG_GETARG_FLOAT8(verify_arg_nonnull(fcinfo, 2));
+    metric = PG_GETARG_INT32(verify_arg_nonnull(fcinfo, 3));
+    
+    close_canopies = (int4 *) palloc(sizeof(int4) * num_all_canopies);
+    num_close_canopies = 0;
+    for (int i = 0; i < num_all_canopies; i++) {
+        if (svec_svec_distance(PointerGetDatum(svec), all_canopies[i], metric)
+            < threshold)
+            close_canopies[num_close_canopies++] = i + 1 /* lower bound */;
+    }
+
+    bytes = ARR_OVERHEAD_NONULLS(1) + sizeof(int4) * num_close_canopies;
+    close_canopies_arr = (ArrayType *) palloc0(bytes);
+    SET_VARSIZE(close_canopies_arr, bytes);
+    ARR_ELEMTYPE(close_canopies_arr) = INT4OID;
+    ARR_NDIM(close_canopies_arr) = 1;
+    ARR_DIMS(close_canopies_arr)[0] = num_close_canopies;
+    ARR_LBOUND(close_canopies_arr)[0] = 1;
+    memcpy(ARR_DATA_PTR(close_canopies_arr), close_canopies,
+        sizeof(int4) * num_close_canopies);
+
+    PG_RETURN_ARRAYTYPE_P(close_canopies_arr);
+}
+
+PG_FUNCTION_INFO_V1(internal_kmeans_closest_centroid);
+Datum
+internal_kmeans_closest_centroid(PG_FUNCTION_ARGS) {
+    SvecType       *svec;
+    ArrayType      *canopy_ids_arr = NULL;
+    int4           *canopy_ids = NULL;
+    ArrayType      *centroids_arr;
+    Datum          *centroids;
+    int             num_centroids;
+    KMeansMetric    metric;
+
+    bool            indirect;
+    float8          distance, min_distance = INFINITY;
+    int             closest_centroid = 0;
+    int             cid;
+
+    svec = PG_GETARG_SVECTYPE_P(verify_arg_nonnull(fcinfo, 0));
+    if (PG_ARGISNULL(1)) {
+        indirect = false;
+    } else {
+        indirect = true;
+        canopy_ids_arr = PG_GETARG_ARRAYTYPE_P(1);
+        canopy_ids = (int4*) ARR_DATA_PTR(canopy_ids_arr);
+    }
+    centroids_arr = PG_GETARG_ARRAYTYPE_P(verify_arg_nonnull(fcinfo, 2));
+    get_svec_array_elms(centroids_arr, &centroids, &num_centroids);
+    if (!PG_ARGISNULL(1))
+        num_centroids = ARR_DIMS(canopy_ids_arr)[0];
+    metric = PG_GETARG_INT32(verify_arg_nonnull(fcinfo, 3));
+
+    for (int i = 0; i < num_centroids; i++) {
+        cid = indirect ? canopy_ids[i] - ARR_LBOUND(canopy_ids_arr)[0] : i;
+        distance = svec_svec_distance(PointerGetDatum(svec), centroids[cid],
+            metric);
+        if (distance < min_distance) {
+            closest_centroid = cid;
+            min_distance = distance;
+        }
+    }
+    
+    PG_RETURN_INT32(closest_centroid + ARR_LBOUND(centroids_arr)[0]);
+}
+
+PG_FUNCTION_INFO_V1(internal_kmeans_canopy_transition);
+Datum
+internal_kmeans_canopy_transition(PG_FUNCTION_ARGS) {
+    ArrayType      *canopies_arr;
+    Datum          *canopies;
+    int             num_canopies;
+    SvecType       *point;
+    KMeansMetric    metric;
+    float8          threshold;
+    
+    canopies_arr = PG_GETARG_ARRAYTYPE_P(verify_arg_nonnull(fcinfo, 0));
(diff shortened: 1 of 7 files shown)


methods/kmeans/src/pg_gp/kmeans.c
```diff
@@ -0,0 +1,186 @@
...
+            canopies_arr, /* array: the initial array object (mustn't be NULL) */
+            1, /* nSubscripts: number of subscripts supplied */
+            &idx, /* indx[]: the subscript values */
+            PointerGetDatum(point), /* dataValue: the datum to be inserted at the given position */
+            false, /* isNull: whether dataValue is NULL */
+            -1, /* arraytyplen: pg_type.typlen for the array type */
+            -1, /* elmlen: pg_type.typlen for the array's element type */
+            false, /* elmbyval: pg_type.typbyval for the array's element type */
+            'd') /* elmalign: pg_type.typalign for the array's element type */
+        );
```
cwelton, line 248: array_set() will allocate a new array.  Given that this is a transition function it would be good to be able to do an update in place when possible (e.g. only when we have verified that the function is being called in an aggregate context).  

This can be addressed in a later patch.

cwelton:
I see no major issues.

methods/kmeans/src/pg_gp/kmeans.c
```diff
@@ -0,0 +1,186 @@
...
+    
+    close_canopies = (int4 *) palloc(sizeof(int4) * num_all_canopies);
+    num_close_canopies = 0;
+    for (int i = 0; i < num_all_canopies; i++) {
+        if (svec_svec_distance(PointerGetDatum(svec), all_canopies[i], metric)
+            < threshold)
+            close_canopies[num_close_canopies++] = i + 1 /* lower bound */;
+    }
+
+    bytes = ARR_OVERHEAD_NONULLS(1) + sizeof(int4) * num_close_canopies;
```
haradh1, line 144: What's the rationale for cases nothing found? I guessed returning NULL was reasonable.

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 248: True. That's an easy fix.

We could also use a doubling-strategy for new allocations, so we would only reallocate logarithmically many times. But then svec is varlena without direct addressing, so there is lots of overhead anyways.

methods/kmeans/src/pg_gp/kmeans.c
```diff
@@ -0,0 +1,186 @@
...
+}
+
+static
+inline
+double
+svec_svec_distance(Datum inVec1, Datum inVec2, KMeansMetric inMetric)
+{
+    PGFunction metric_fn = NULL;
+
+    switch (inMetric)
```
haradh1: Isn't it better to extract the switch out of loop for performance?

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 144: I chose an empty array in order to not have a special case. I.e., I think returning "empty set" is more appropriate here than "I don't know".

src/ports/postgres/modules/kmeans/kmeans.sql_in
```diff
@@ -527,14 +351,15 @@ $$ LANGUAGE plpgsql;
...
-    , SFUNC = MADLIB_SCHEMA.internal_kmeans_canopy_transition
-    m4_ifdef(`GREENPLUM',`,prefunc = MADLIB_SCHEMA.internal_kmeans_canopy_merge')
-    , FINALFUNC = MADLIB_SCHEMA.internal_kmeans_canopy_finalize  
+    , INTEGER           -- distance metric
+    , FLOAT8            -- threshold (T1)
+) (
+    stype = MADLIB_SCHEMA.svec[],
+    sfunc = MADLIB_SCHEMA.internal_kmeans_canopy_transition,
+m4_ifdef(`__GREENPLUM__', `
+    prefunc = array_cat,
```
haradh1, line 359: I see you removed the original merge function, which looks to me like a logic change. Won't this include unnecessary more points than before?

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -386,12 +396,12 @@ def __init_canopy(  madlib_schema, src_points, n, dist_metric, dist_func,
...
         # Calculate the cross distances >> and derive T1 and T2 
-        # T1 >> is the max value from 1st quantile (10 buckets)
-        # T2 >> is the min value from 10th quantile (10 buckets)
+        # T1 >> is the min value from 10th quantile (10 buckets)
+        # T2 >> is the max value from 1st quantile (10 buckets)
         sql = '''
             SELECT 
-                max( CASE WHEN ntile = 1 THEN dist ELSE null END) as t1 
-                , min( CASE WHEN ntile = 10 THEN dist ELSE null END) as t2 
+                max( CASE WHEN ntile = 10 THEN dist ELSE null END) as t1
```
haradh1: I guess you forgot to swap min/max.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp, line 359: Yes. But in the old version, a centroid was removed during merge if it was closer to another centroid than 0.5 \* T2. That means, even after just the first merge, you can only guarantee that each point is no further from the closest centroid than 1.5 \* T2. Now if 1.5 \* T2 > T1, some points would just have been "lost" when filling TempPoints0_Canopies. The problem was even more severe because the number of times a set of centroids is part of a merge operation is unbounded.

methods/kmeans/src/pg_gp/kmeans.c
haradh1, line 144: ARR_NDIM() = 0, then?

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: Indeed, thanks! Good catch!

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 144: I see. PG wants empty arrays to have dimension zero. So we can choose between NULL and construct_empty_array(). I still think that semantically the empty array make more sense. What do you think?

haradh1, line 144: Yeah, it's sometimes discussed in even the community, but no conclusion. I usually avoid empty array and choose NULL since empty array is problematic. But, yeah, it's your choice.

src/ports/postgres/modules/kmeans/kmeans.sql_in
haradh1, line 359: Ah, right. Better than before. I was a little afraid about increasing number of canopies which would cause performance issue.

haradh1:
Done. Looks good!

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 248: For now, I have (only) added the appropriate MemoryContextSwitchTo() calls.

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 144: For now, I decided to return an empty array, so I only needed to change the C code.

methods/kmeans/src/pg_gp/kmeans.c
fschopp: In this case, I would assume the compiler would do that optimization. But you are right. I changed this.

fschopp:
For your review. Please do not pull yet.

src/ports/postgres/modules/kmeans/kmeans.sql_in
cwelton, line 359: The concern I have here is anti-scalability behavior.  If the number of canopies is linearly dependent on the number of database segments then the cost goes up with more segments.

methods/kmeans/src/pg_gp/kmeans.c
cwelton, line 248: The memory context switch doesn't actually matter.

Within the aggnode what occurs is:

```
if (return != input)
{
   switchToMemoryContext(aggContext);
   return = copy(return);
   switchToMemoryContext(rowContext);
}
```

If anything moving it to the agg context may be harmful because then it won't be cleaned up when the row context is released.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp, line 359: True. Some simple options:
- We could remove the merge function and just find the canopies sequentially. _Pro:_ Very easy. Adheres to the McCallum et al. paper. _Con:_ One pass over the data is non-parallelized.
- We could continue dropping centroids (either in the merge step or at the very end) but adjust the internal_get_array_of_close_canopies() function as follows: Return the set of canopies closer than the threshold T1 if this set is non-empty. Otherwise, return the closest canopy, however far away it may be. (Alternatively: Return the set of all canopies.) _Pro:_ Moderate amount of work to implement. Everything runs in parallel. Adheres to the McCallum et al. paper in the sense that the quality of the outcome can intuitively only be better (because all we do is replacing some approximate distances by exact distances). _Con:_ If we have many stray points, the benefits of canopy clustering are low.
- Continue to drop centroids that are close to each other, but afterwards (only on the master) find canopies among those points that currently are contained in no other canopy. _Pro:_ There is a sequential algorithm that would produce the same result. Strictly adheres McCallum et al. paper. _Con:_ Most work to implement (after the sequential phase we would also need, for each point, to append to the list of close canopies). Finding canopies for stray points is non-parallel.
- Continue to drop centroids (in the final function though). Define "closeness" as the maximum distance each point can be away from a centroid (i.e., T1 + T2 currently). _Pro_: Easy to implement _Con:_ Potentially dropping a lot of canopies. Slight deviation from McCallum et al. paper.

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 248: Indeed. For whatever reason, I had in the back of my mind that PG would avoid copying if the memory was allocated in the aggregate context.

Question: Is it OK if VARSIZE(array) is larger than what the array actually uses, i.e., what would be computed using ARR_DIMS() etc.? Or do any parts of PG/GP rely on VARSIZE and the computed size being equal?

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 144: Thinking about it again... Actually, returning NULL is a good thing because internal_kmeans_closest_centroid() interpretes NULL for argument "canopies" as "compare with all centroids". So I will change it to NULL.
See my [other comment below](https://github.com/madlib/madlib/pull/104#r393150).

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 248: It seems this has already been implemented in the form of accumArrayResult(). So I'll use that.

src/ports/postgres/modules/kmeans/kmeans.sql_in
haradh1, line 359: The first choice sounds impossible to me as it doesn't scale at all. The second is in between the current one and the old one, still bogus logic. We need to have the original points information per canopy to really merge them, but I bet it's not worth spending it, since the right approach would be "to calculate inexpensive distance metric" by inverted index as in the paper anyway.

methods/kmeans/src/pg_gp/kmeans.c
cwelton, line 248: I don't think you can use accumArrayResult because user defined aggregates are not allowed to use "internal" datatypes for transition state.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp, line 359: I don't see bogus logic in item no. 2.
We have "approximate distance" = exact distance if in the same canopy and infinity otherwise.
So the suggestion is this: If a point is not close to any canopy, we will always use the exact distance for that point during the k-means phase (as opposed to the approximate distance that we use for all points that do have close canopies).

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -422,50 +432,71 @@ def __init_canopy(  madlib_schema, src_points, n, dist_metric, dist_func,
...
+    # Generate initial centroids using T2 (smaller)
+    __run_quietly( 'DROP TABLE IF EXISTS TempArrayOfCentroids');
+    sql = '''
+        CREATE TEMP TABLE TempArrayOfCentroids AS
+        SELECT 
+            {madlib_schema}.internal_remove_close_canopies(
+                {madlib_schema}.kmeans_canopy(coords, {metric}, {t2}),
+                {metric},
+                {t1}
+            ) AS ccoords
```
fschopp, line 450: [Solution no. 4 as discussed](https://github.com/madlib/madlib/pull/104/files#r393150)

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -422,50 +432,71 @@ def __init_canopy(  madlib_schema, src_points, n, dist_metric, dist_func,
...
         if rv[0]['t2'] is None:
-            plpy.error( 'not enough data to generate canopy T2 threshold (T1<T2)')
+            plpy.error( 'not enough data to generate canopy T2 threshold (T1>T2)')
         else:
             t2 = rv[0]['t2'];
     elif t2 <= 0:
-        plpy.error( 'canopy threshold T2 can not be zero or negative')        
-        
-    # Generate initial centroids using T1 (smaller)
+        plpy.error( 'canopy threshold T2 can not be zero or negative')
```
cwelton: s/can not/cannot/

methods/kmeans/src/pg_gp/kmeans.c
fschopp, line 248: If we really wanted to speed things up then, I guess all we can do then is having our own "ArrayBuildState" that we store in a bytea...

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: Replaced all occurrences of 'can not' by 'cannot' in this file. I have not looked at the error messages, yet. There are probably more typos.

PR #105, opened 2012-01-30
resolve madlib-338

renyi533:
This pull request resolves madlib-338

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.sql_in
+++ b/methods/decision_tree/src/pg_gp/decision_tree.sql_in
@@ -3921,9 +3921,22 @@ BEGIN
                         MADLIB_SCHEMA.__strip_schema_name(tree_table_name)
                     )
                 ),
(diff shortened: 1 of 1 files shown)


methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -3921,9 +3921,22 @@ BEGIN
                         MADLIB_SCHEMA.__strip_schema_name(tree_table_name)
                     )
                 ),
-                'the specified tree table doesn''t exists'
+                'ERROR: relation '||quote_literal(tree_table_name)||' does not exist'
```
cwelton: 'ERROR:' should not be a part of the error message.  This is added automatically.

Same comment applies to both errors raised below as well.

cwelton:
I see no issues with the body of the change.  Once the error messages are fixed it should be good to merge.

renyi533:
I updated based on the comments. Thanks.

PR #109, opened 2012-02-01
kmeans related fixes

agorajek:
Targeted JIRAs:
- http://jira.madlib.net/browse/MADLIB-349
- http://jira.madlib.net/browse/MADLIB-380
- http://jira.madlib.net/browse/MADLIB-384
- http://jira.madlib.net/browse/MADLIB-385
- http://jira.madlib.net/browse/MADLIB-388
- http://jira.madlib.net/browse/MADLIB-389

### DIFF
--- a/src/ports/postgres/modules/kmeans/kmeans.py_in
+++ b/src/ports/postgres/modules/kmeans/kmeans.py_in
@@ -342,7 +342,7 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
 # Centroid initialization using canopy
 # ----------------------------------------
 def __init_canopy(  madlib_schema, src_points, n, dist_metric, dist_func, 
-                    t1, t2, sample_pct):
+                    t1, t2):
     """
     Creates the initial set of centroids using canopy clustering.
 
@@ -355,18 +355,16 @@ def __init_canopy(  madlib_schema, src_points, n, dist_metric, dist_func,
            (e.g. 'cosine_distance')
     @param t1 Larger threshold for canopy clustering 
     @param t2 Smaller threshold for canopy clustering 
-    @param sample_pct Percent of the input data points to use for centroid 
-           seeding
     """
     
     # Find thresholds T1 and T2 (T1 > T2)
     if t1 is None or t2 is None:
     
-        # Allow the user to specify sample size 
-        if sample_pct:
-            sample_size = floor((n*sample_pct)/100);
+        # Set the sample limit for T1/T2 value analysis
+        if n < 1000:
+            sample_size = n;
         else:
-            sample_size = floor(n/100); # default 1%
+            sample_size = 1000;
         
         # Temp table for random points
         random_points = 'kmeans_random_points'
@@ -643,9 +641,7 @@ def kmeans( madlib_schema
     elif init_method == 'canopy':
         k = None;
         cset_count = None;
-        if init_sample_pct is None:
-            init_sample_pct = 1.0;
-        info( ' * init_method = %s (t1=%s, t2=%s, %s pct sample)' % (init_method, t1, t2, init_sample_pct));
+        info( ' * init_method = %s (t1=%s, t2=%s)' % (init_method, t1, t2));
 
     # Validate: init_method & k (other methods)
     elif k > 0:
@@ -865,8 +861,7 @@ def kmeans( madlib_schema
         start = time.time();
         centr_count, t1, t2 = __init_canopy( madlib_schema, 'TempPoints0', 
                                              point_count, dist_metric, 
-                                             dist_func, t1, t2, 
-                                             init_sample_pct);
+                                             dist_func, t1, t2)
         time_sec = round( time.time() - start, 3)
         info( ' * centroids: %s seeded using canopy clustering with t1=%s and t2=%s (%s sec)' 
                % (centr_count, t1, t2, str(time_sec)));
--- a/src/ports/postgres/modules/kmeans/kmeans.py_in
+++ b/src/ports/postgres/modules/kmeans/kmeans.py_in
@@ -163,7 +163,7 @@ def __init_random( madlib_schema, src_points, k, n):
 # ----------------------------------------
 # Centroid initialization using kmeans++
 # ----------------------------------------
-def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
+def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_fraq):
     """
     Creates the initial set of centroids using kmeans++ [1].
 
@@ -181,8 +181,8 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
     @param k Number of initial centroids
     @param n Total number of input data points 
     @param dist_func Name of the distance function (e.g. 'l2norm')
-    @param sample_pct Percent of the input data points to use for centroid 
-           seeding \f$[0,1]\f$
+    @param sample_fraq Fraction of the input data points to use for centroid 
+           seeding, \f$[0,1]\f$
     """
     
     # Temporary data point tables
@@ -208,17 +208,17 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
     );
     
     # Allow user to specify sample size or set the default
-    if sample_pct:
-        sample_size = floor(n*sample_pct);
+    if sample_fraq:
+        sample_size = floor(n*sample_fraq);
     else:
         sample_size = floor(n/100);    # default := 0.01
     
     # Sample size must be greater than k
-    if sample_pct and sample_size < k:
+    if sample_fraq and sample_size < k:
         plpy.error( 'sample size (%s) is smaller than specified k (%s)' % (int(sample_size), k))        
     
     # If sampling on
-    if sample_pct and sample_size > 0 and sample_size < n:
+    if sample_fraq and sample_size > 0 and sample_size < n:
         plpy.execute('''    
             INSERT INTO {initial_sample} (pid, coords, min_distance) 
             SELECT pid, coords::{madlib_schema}.SVEC, NULL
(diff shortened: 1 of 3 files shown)


src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -313,15 +313,26 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
...
+                    FROM 
+                        {temp_target} p, 
+                        (SELECT sum(min_distance^2) as dist_sqrd FROM {temp_target}) total
+                    ) q1
+                    , random() AS x
+                ) q2
+            WHERE
+                x <= cum_weight
+            ORDER BY pid
             LIMIT 1
```
cwelton: We should have a good sense of the performance implications of this change.  The prior plan would generate:
     Limit  (cost=3.51..3.53 rows=1 width=4)
       ->  Gather Motion 2:1  (slice1; segments: 2)  (cost=3.51..3.53 rows=1 width=4)
             ->  Limit  (cost=3.51..3.51 rows=1 width=4)
                   ->  Sort  (cost=3.51..3.61 rows=21 width=4)

While the new plan generates:
    ...
               ->  Window  (cost=3.51..4.53 rows=41 width=4)
                     ->  Gather Motion 2:1  (slice1; segments: 2)  (cost=3.51..4.43 rows=21 width=4)
                           ->  Sort  (cost=3.51..3.61 rows=21 width=4)

Where the most important difference is the lack of a LIMIT node prior to the gather motion.  E.g. it pulls ALL the data back not just a small subset.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -313,15 +313,26 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
...
+                    (
+                    SELECT 
+                        p.pid
+                        , p.coords 
+                        , p.min_distance^2 / total.dist_sqrd as weight
+                    FROM 
+                        {temp_target} p, 
+                        (SELECT sum(min_distance^2) as dist_sqrd FROM {temp_target}) total
+                    ) q1
+                    , random() AS x
```
cwelton: Is the intention to generate only a single random number used for all rows, or a different random number for each row.  Because this is the former, which does not seem correct.

cwelton: Example:
    select \* from generate_series(1,5),random();
    generate_series |      random  
    -----------------+------------------
                   1 | 0.85080617107451
                   2 | 0.85080617107451
                   3 | 0.85080617107451
                   4 | 0.85080617107451
                   5 | 0.85080617107451
    (5 rows)

src/ports/postgres/modules/kmeans/kmeans.py_in
agorajek: "ORDER BY pid LIMIT 1" could be substituted with MIN(pid) and another JOIN with the {temp_target} to get the coords. Perhaps this may be a better approach. I will test it.

src/ports/postgres/modules/kmeans/kmeans.py_in
agorajek: Yes, single random number is required here.

src/ports/postgres/modules/kmeans/kmeans.py_in
agorajek: I find the following "min/join" version to be up to 20% slower than the current "order/limit" one:

```
INSERT INTO {centroids} (cid, coords) 
SELECT {i}, coords
FROM
    (
    SELECT min(pid) as pid
    FROM
        (
        SELECT
            pid
            , sum(weight) OVER (ORDER BY pid) AS cum_weight
            , x
        FROM
            (
            SELECT 
                p.pid
                , p.min_distance^2 / total.dist_sqrd as weight
            FROM 
                {temp_target} p, 
                (SELECT sum(min_distance^2) as dist_sqrd FROM {temp_target}) total
            ) q1
            , random() AS x
        ) q2
    WHERE
        x <= cum_weight
    ) q3
JOIN {temp_target} p USING (pid)
```

cwelton: Yes, I'd expect that to be slower.  The basic design difference between the new query and the old query is the introduction of the window function which is what is forcing the full gather.  Did you compare to the performance of the older query?

agorajek: I did a rough comparison, in which to my surprise the new code turned out to be ~10% faster. But in any case (as Florian correctly pointed out) the old method was not sampling according to kmeans++.

cwelton:
Looks reasonable to me.  Florian do you have any other comments or concerns?

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -313,15 +313,26 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
...
+                ) q2
+            WHERE
+                x <= cum_weight
+            ORDER BY pid
```
fschopp: This can be simplified. Also, I would move the LIMIT into the subquery. Otherwise and AFAIK, the result is strictly speaking undefined, i.e., the data might not be sorted according to pid outside of the subquery any more and LIMIT might filter the wrong tuple.

```
INSERT INTO {centroids} (cid, coords)
SELECT {i}, coords FROM (
    SELECT
        pid, coords, sum(min_distance^2) OVER (ORDER BY pid) AS cum_weight
    FROM
        {temp_target} p,
        (SELECT sum(min_distance^2) * random() AS r FROM {temp_target}) subquery
    WHERE
        r <= cum_weight
    ORDER BY pid
    LIMIT 1
)
```

It shouldn't make a difference, but you never know how smart the query planner is. So you could compare the query plans for "ORDER BY pid" and "ORDER BY cum_weight" (in the subquery, not the window function), and see which one is better. They are both equivalent.

cwelton: That query doesn't work.   cum_weight is an window expression so it cannot occur in the where clause.  If you wanted to guarantee that the optimizer doesn't stomp on your order by you just need to promote it to the outer block.

```
SELECT {i}, coords FROM (
    SELECT
        pid, coords, sum(min_distance^2) OVER (ORDER BY pid) AS cum_weight
    FROM
        {temp_target} p,
        (SELECT sum(min_distance^2) * random() AS r FROM {temp_target}) subquery
) q
WHERE
      r <= cum_weight
ORDER BY pid
LIMIT 1
```

The second ORDER BY shouldn't add any additional cost since the data should have been ordered to resolve the inner OVER (ORDER BY pid), yet it guarantees that the query optimizer can't choose a plan that does not include ordering for the outer query block.

fschopp: You are right, of course. But you need to include r in the SELECT list of the subquery or alternatively move

```
(SELECT sum(min_distance^2) * random() AS r FROM {temp_target}) subquery
```

outside as well.

cwelton: Agreed.

agorajek:
Nice simplification Florian. Updated and ready to merge.

agorajek:
Done. Ready for a review & merge.

cwelton:
Changes look reasonable to me.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -658,8 +651,8 @@ def kmeans( madlib_schema
         elif init_method.lower() == 'kmeans++':
             init_method = 'kmeans++';
             if init_sample_pct is None:
-                init_sample_pct = 1.0;
-            info( ' * init_method = %s (%s pct sample)' % (init_method, init_sample_pct));
+                init_sample_pct = 0.01;
```
fschopp: I think you should rename the variable and remove the '_pct' suffix.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -528,7 +526,7 @@ def kmeans( madlib_schema
     @param init_method Name of the centroid initialization method 
            (e.g. 'kmeans++', 'random')
     @param init_sample_pct Percent of the input data points to use for kmeans++
```
fschopp: 'Percent' should not be used here. Better: 'Fraction of the input data points'

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -182,7 +182,7 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
     @param n Total number of input data points 
     @param dist_func Name of the distance function (e.g. 'l2norm')
     @param sample_pct Percent of the input data points to use for centroid
```
fschopp: Remove '_pct' and replace 'Percent' by 'Fraction'

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -182,7 +182,7 @@ def __init_plusplus( madlib_schema, src_points, k, n, dist_func, sample_pct):
     @param n Total number of input data points 
     @param dist_func Name of the distance function (e.g. 'l2norm')
     @param sample_pct Percent of the input data points to use for centroid 
-           seeding
+           seeding \f$[0,1]\f$
```
fschopp: Just putting \f$[0,1]\f$ here is confusing. If you want to explicitly mention the domain, why not just write 'Fraction (between 0 and 1)' in the line before?

agorajek:
All good comments Florian. I am planning to do the following cleanup:
- use "sample_fraq" for any sample variable defined as fraction (or previously percent)
- use "sample_size" for any sample variable defined as number of rows.

agorajek:
Done.

fschopp:
Was 'sample_fraq' a typo? I'll correct it during merge.

cwelton:
Fraq seems like an unusual choice of abbreviation for fraction.

agorajek:
Yes, it is a typo :) Thanks Florian.

fschopp:
Merged with the above typo fixed.

PR #110, opened 2012-02-03
Small fixes for a variety of issues

fschopp:
Added "utility" module:
- Function version() returns information about MADlib version, build time, git revision, compiler, etc. (Implements MADLIB-309)

Probability functions:
- normalCDF now returns NaN on NaN input, instead of raising an error. (Fixes MADLIB-356)

C++AL:
- Improved exception handling and mapping to SQL error states. Added a new exception for "no solution found"

Logistic regression:
- Some code clean-up: Removed calls to locals() and replace use of keyword arguments by named parameters.
- Return NULL if zero input tuples. This matches the behavior of standard PostgreSQL aggregates when applied to 0 rows. (Fixes MADLIB-284)
- Can now handle any input types that can be converted to BOOLEAN (dependent variable) or FLOAT8[](independent variables). (Fixes MADLIB-283)
- Improved error message in case of non-convergence due to ill-conditioned input. (Fixes MADLIB-345)
- If a negative number is passed for the max # of iterations, an error is displayed. (Fixes MADLIB-355)
- Documentation updated to inform user that a negative convergence threshold essentially disables this particular criterion. (Fixes MADLIB-354)

Linear regression:
- Throw domain_error if input data is not finite.

Decision Trees:
- Documentation is now included, but still needs a lot of work. (Fixes MADLIB-365)


src/ports/postgres/dbconnector/UDF_impl.hpp
```diff
@@ -7,6 +7,10 @@
 // Workaround for Doxygen: Ignore if not included by dbconnector.hpp
 #ifdef MADLIB_DBCONNECTOR_HPP
 
+#define MADLIB_HANDLE_STANDARD_EXCEPTION(err) \
+    sqlerrcode = err; \
+    strncpy(msg, exc.what(), sizeof(msg));
```
haradh1, line 12: To my best knowledge, it's better to wrap this kind of macro with do ... while(0), without semicolon at the end.

src/ports/postgres/modules/regress/logistic.sql_in
```diff
@@ -245,10 +258,11 @@ RETURNS DOUBLE PRECISION[]
 AS 'MODULE_PATHNAME'
 LANGUAGE C IMMUTABLE STRICT;
 
-DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.logreg_cg_step(
-    /*+ y */ BOOLEAN,
-    /*+ x */ DOUBLE PRECISION[],
-    /*+ previous_state" */ DOUBLE PRECISION[]);
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.logregr_igd_step_final(
```
haradh1, line 261: This function is created but I don't see CREATE AGGREGATE change.

src/ports/postgres/modules/regress/logistic.sql_in
```diff
@@ -289,6 +298,23 @@ CREATE AGGREGATE MADLIB_SCHEMA.logregr_irls_step(
...
+ *        method for computing logistic regression
+ */
+CREATE AGGREGATE MADLIB_SCHEMA.logregr_igd_step(
+    /*+ y */ BOOLEAN,
+    /*+ x */ DOUBLE PRECISION[],
+    /*+ previous_state */ DOUBLE PRECISION[]) (
+    
+    STYPE=DOUBLE PRECISION[],
+    SFUNC=MADLIB_SCHEMA.logregr_igd_step_transition,
+    m4_ifdef(`GREENPLUM',`prefunc=MADLIB_SCHEMA.logregr_igd_step_merge_states,')
```
haradh1, line 313: Missing final function here?

src/ports/postgres/modules/regress/logistic.sql_in
fschopp, line 261: Good eyes! Thanks!

cwelton:
Pulled.

src/ports/postgres/dbconnector/UDF_impl.hpp
cwelton, line 12: Agree, that way it works better as a function like macro.

PR #111, opened 2012-02-04
Cast of float array to sparse vector cannot handle null values, take two

haradh1:
I tried to put together aojwang's fix to use makeInplaceSparseData, but it turned out that I was wrong. It is for index/values set and casting float array doesn't have index. I changed my mind to modify underlying arr_to_sdata to let it accept NULL bitmap, so that double memory copies of array isn't needed.

### DIFF
--- a/methods/svec/src/pg_gp/SparseData.c
+++ b/methods/svec/src/pg_gp/SparseData.c
@@ -24,6 +24,8 @@
 #include "access/htup.h"
 #include "catalog/pg_proc.h"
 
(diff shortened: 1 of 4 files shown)


cwelton:
In PR-81 Hitoshi said:

> It turned out I was wrong, makeInplaceSparseData is not appropriate here. Though I sent out PR111, it might be better to merge this fix and put off my fix for later release, to make sure it is correct.

So I am closing this PR for now.

PR #112, opened 2012-02-07
Fixes for madlib-341/381/382

aojwang:
1. madlib-341: Check duplicated id values in training/classification table. (utility.sql_in).
2. madlib-381/382: Out of memory FOR C45_TRAIN on GPDB4.1. (decision_tree.sql_in).
3. Add an additional star (/**) to the classify, score, display, clean API.
4. Refine documentation for public API.

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.sql_in
+++ b/methods/decision_tree/src/pg_gp/decision_tree.sql_in
@@ -1,14 +1,11 @@
-/* ----------------------------------------------------------------------- *//** 
+/*
  *
  * @file decision_tree.sql_in
  *
  * @brief decision tree APIs and main controller written in PL/PGSQL
+ *
  * @date Dec. 22 2011
- *
- * @sa For a brief introduction to decision trees, see the
- *     module description \ref grp_dectree.
- *
- *//* ----------------------------------------------------------------------- */
+ */
 
 m4_include(`SQLCommon.m4')
 
@@ -28,9 +25,7 @@ m4_ifelse(
     `m4_define(`__POSTGRESQL_PRE_9_0__')'
 )
 
-/**
-@addtogroup grp_dectree
-
+/*
 @about
 
 This module provides an implementation of the C4.5 decision tree algorithm. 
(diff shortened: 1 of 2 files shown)


cwelton:
I would be fine taking the fix for madlib-341, this looks low risk.  

However the fixes for madlib-381/382 are larger in scope than we can take at this stage in the release.  We do not have time to test and address any fallout from these changes.  Recommend waiting until after the release and inclusion in a later patch.

aojwang:
Ok, i will create another pull request to include the fix for 341 and documentation issue.

aojwang:
Hi Caleb,

If time is allowed, we still prefer to push the changes to the master. The reason is as following.
1) After the fixes, the node id will be changed and trained tree table will also be changed(jump column changed to lmc_nid and lmc_fval). Therefore, I think it's better to push to master, so that they look the same between this release and next release(patch) for dev users.  

2) No jump column (int[]) in trained tree any more, since its max size is equal to the number of distinct value of the best split feature represented by the row. We introduce two columns:  lmc_nid (leftmost child node id) and lmc_fval(the feature value which leads to the leftmost child node id). Other child nodes' ids can be deduced by lmc_nid and lmc)_fval, since the values for a feature and child nodes' ids are continuous. Therefore, most changes are related to change jump column to lmc_nid and lmc_fval although the fixes are large in scope. The risk should be low.

3) Delete some unnecessary code, such as reassign serial ids for tree node (__generate_final_tree). Now, it's very simple. Nodes' ids will not be changed. no aggregate functions to update lmc_nid and lmc_fval (just by simple SQL statement). Therefore, the risk should be low.

After discussed with Jiali, they may take 21 hours to run the minimum tests, and 29 hours for whole tests. Currently, they are working on k-means tests. Next night, they have time to run DT tests. We may get the result on Friday afternoon or next monday.

What do you think? thanks very much.

cwelton:
Hard deadline is Thursday evening here, which gives aproximately 48 hours, this is not sufficient time to pull a change like this.

aojwang:
Hi Caleb,

The pull request, https://github.com/madlib/madlib/pull/113, include all the bug fixes except for madlib-381/382. Thanks.

cwelton:
Closing per discussion based on the understanding that this is included in later PR.

PR #114, opened 2012-02-08
resolve madlib-373

renyi533:
This fix resolved madlib-373.

I have tested all the datasets supplied by Jiali. The results in PG8.4 are consistent with other version.

I also tested up to twenty other datasets. The results are also consistent.

The fix is straightforward.


cwelton:
Looks good to me too.  Change is straight forward, only effects PG8.4, and fixes wrong results.  Good to merge.

PR #115, opened 2012-02-08
kmeans: fix for handling NaN values (MADLIB-364)

agorajek:
Documentation update is included in Florian's latest doc commit.

### DIFF
--- a/src/ports/postgres/modules/kmeans/kmeans.py_in
+++ b/src/ports/postgres/modules/kmeans/kmeans.py_in
@@ -787,14 +787,16 @@ def kmeans( madlib_schema
     ''';
     __run_quietly( sql);
     # Create the Temporary Point table
-    # Remove any NULL values
+    # Remove any NULL, NaN, and any non-finite values
     sql = '''
         INSERT INTO TempPoints0 
         SELECT pid, coords, 0, null
         FROM ''' + src_view + '''
-        WHERE ''' + madlib_schema + '.svec_dot(coords,coords) IS NOT NULL';
+        WHERE abs( 
+                  coalesce(''' + madlib_schema + '''.svec_elsum(coords), 'Infinity'::FLOAT8)
+                 ) < 'Infinity'::FLOAT8''';
     plpy.execute( sql);        
-    # Find the number of points after skipping NULL values
+    # Find the number of points after cleanup
     sql = 'SELECT count(*) as cnt FROM TempPoints0';
     rv = plpy.execute( sql);    
     orig_point_count = point_count;      
@@ -810,7 +812,7 @@ def kmeans( madlib_schema
         plpy.error( "input points must have the same dimensions\n")            
     info( ' * points: ' + str(orig_point_count) + ' (' \
             + str(rv[0]['min_dim']) + ' dimensions), kept ' + str(point_count) + \
-            ' after removing NULLs');     
+            ' after removing non-finite values');     
 
     # Initialize centroids
     info( '... generating initial centroids')
--- a/src/ports/postgres/modules/kmeans/test/kmeans.sql_in
+++ b/src/ports/postgres/modules/kmeans/test/kmeans.sql_in
@@ -19,6 +19,11 @@ SELECT * FROM MADLIB_SCHEMA.kmeans_sample_data(
     'km_testdata'
 );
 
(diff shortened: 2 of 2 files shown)


src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -787,14 +787,14 @@ def kmeans( madlib_schema
...
     __run_quietly( sql);
     # Create the Temporary Point table
-    # Remove any NULL values
+    # Remove any NULL, NaN, and any non-finite values
     sql = '''
         INSERT INTO TempPoints0 
         SELECT pid, coords, 0, null
         FROM ''' + src_view + '''
-        WHERE ''' + madlib_schema + '.svec_dot(coords,coords) IS NOT NULL';
+        WHERE abs(''' + madlib_schema + '.svec_elsum(coords)) >= 0';
```
cwelton: -- What about vectors containing negative values?  I don't think el_sum >= 0 is the right condition here. --

Which is, of course, why you have the abs() there.  Never mind earlier comment.

cwelton: However I get this result:

```
 select 'NaN'::float8 >= 0;
 column
 ----------
  t
 (1 row)
```

Which means that I'm not convinced that this test will exclude the NaN's correctly.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -810,7 +810,7 @@ def kmeans( madlib_schema
         plpy.error( "input points must have the same dimensions\n")            
     info( ' * points: ' + str(orig_point_count) + ' (' \
             + str(rv[0]['min_dim']) + ' dimensions), kept ' + str(point_count) + \
-            ' after removing NULLs');     
+            ' after removing non-finite values');
```
cwelton, line 815: 'Infinity' is a non-finite value that would pass your >= 0 check as well.

src/ports/postgres/modules/kmeans/kmeans.py_in
agorajek: NaN is greater than zero ?:

``` sql
test=# select 'NaN'::float8 > 0;
 ?column? 
----------
 t

test=# select 'NaN'::float8 = 0;
 ?column? 
----------
 f

test=# select 'NaN'::float8 < 0;
 ?column? 
----------
 f
```

Isn't this a bug?

fschopp: Yikes! GP is not IEEE 754 compliant.

agorajek:
This commit should fix the problems. Also added a test.

cwelton:
As stands I'd like to hold off on this until after the release.

agorajek:
Why?

cwelton:
Because it has to release tomorrow and the solution provided here does not look complete.

cwelton:
Plus, I'm fine release noting this issue.

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: None of this is an efficient way of checking finiteness of all elements. It's also quite convoluted. That said, the correct (but still ugly) way would be something like

```
WHERE abs(coalesce(''' + madlib_schema + '.svec_elsum(coords), 'Infinity'::FLOAT8)) < 'Infinity'::FLOAT8
```

The above does not catch Infinity.

agorajek:
1) Fine that it's a bug, but we would need a workaround in the interim.
2) Really?

``` sql
test=# select madlib.svec_elsum('{1,1}:{1,NaN}'::madlib.svec);
 svec_elsum 
------------
        NaN
```

3) Nice! It looks like it will work for NaNs as well.

agorajek:
Fixed again.

src/ports/postgres/modules/kmeans/kmeans.py_in
cwelton: Regarding IEEE 754 compliance, from the postgres documentation:

> Note: In most implementations of the “not-a-number” concept, NaN is not considered equal to any other numeric value (including NaN). In order to allow numeric values to be sorted and used in tree-based indexes, PostgreSQL treats NaN values as equal, and greater than all non-NaN values.

cwelton:
Pulling after review

PR #118, opened 2012-02-15
New Statistics Module

fschopp:
_This pull request should be merged only after pull request 139. Some minor changes will be necessary for that. However, I'd like to open the discussion already now because I'll be offline for two days._


src/modules/stats/t_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    typename HandleTraits<Handle>::ReferenceToDouble parameter;
+};
+
+/**
+ * @brief Update the corrected sum of squares
+ *
+ * For numerical stability, we should not compute the sample variance in the
+ * naive way. The literature has many examples where this gives bad results
+ * even with moderately sized inputs.
```
fschopp, line 82: In fact, the implementation below is like in Chan et al. (and not a naive one).

cwelton:
Current plan is to schedule this for pull.

src/config/Version.yml
```diff
@@ -1 +1 @@
-version: 0.3
+version: 0.4dev
```
cwelton, line 1: Good, I was going to file a jira for this.   I didn't expect to see it in the inferential statistics PR, but I'm happy to see it changing.

We should probably still have a jira to remove the "dev" prior to release.

src/modules/stats/chi_squared_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+// Workaround for Doxygen: A header file that does not declare namespaces is to
+// be ignored if and only if it is processed stand-alone
+#undef _DOXYGEN_IGNORE_HEADER_FILE
+#include "chi_squared_test.hpp"
+
+/**
+ * @brief Transition state for chi-squared functions
+ *
+ * Note: We assume that the DOUBLE PRECISION array is initialized by the
+ * database with length 8, and all elemenets are 0.
```
cwelton: s/elemenets/elements/

cwelton: The assumption that the state will be an array of length 8 is somewhat suspect.  This will be the case when it is properly initialized by an aggregate, but individual functions of can be directly callable.  When called directly the parameters can be anything that a user provides.  

For instance, what is the behavior if a user calls:  `select chi2_gof_test_merge_states(array[1],array[2]);`?

If this will reference uninitialized memory then it needs to be fixed.

src/modules/stats/chi_squared_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+    double &ioLeftSumObsSquareOverExp, double &ioLeftSumObs,
+    double &ioLeftSumSquaredDeviations,
+    double inRightNumRows, double inRightSumExp,
+    double inRightSumObsSquareOverExp, double inRightSumObs,
+    double inRightSumSquaredDeviations) {
+    
+    if (inRightNumRows <= 0)
+        return;
+    
+    // FIXME: Use compensated sums for numerical stability
```
cwelton, line 80: File jiras for unresolved issues/improvements and refer to the jira number in the comment.

src/modules/stats/chi_squared_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+    ioLeftSumObsSquareOverExp += inRightSumObsSquareOverExp;
+    ioLeftSumObs += inRightSumObs;
+}
+
+AnyType
+chi2_gof_test_transition::run(AnyType &args) {
+    Chi2TestTransitionState<MutableArrayHandle<double> > state = args[0];
+    int64_t observed = args[1].getAs<int64_t>();
+    double expected = args[2].isNull() ? 1 : args[2].getAs<double>();
+    int64_t df = args[3].isNull() ? -1 : args[3].getAs<int64_t>();
```
cwelton: Slight discrepancy from the documented behavior.  

Documented:

>  @param df Degrees of freedom. This is the number of events reduced by the  degree of freedom lost by using the observed numbers for defining the expected number of observations. If this parameter is \c NULL, the degree  of freedom is taken as \f$ (k - 1) \f$.

Actual:

>  @param df Degrees of freedom. This is the number of events reduced by the  degree of freedom lost by using the observed numbers for defining the expected number of observations. If this parameter is \c NULL or < 0, the degree  of freedom is taken as \f$ (k - 1) \f$.

src/modules/stats/chi_squared_test.cpp
```diff
@@ -0,0 +1,188 @@
...
+    
+    if (state.uniformDist != args[2].isNull()) {
+        if (state.numRows > 0)
+            throw std::invalid_argument("Expected number of observations must "
+                "be given for all events or must be NULL for all events, in "
+                "which case a discrete uniform distribution is assumed.");
```
cwelton: Not technically correct since NULL and -1 are treated as equivalent, but not a huge deal.

src/modules/stats/kolmogorov_smirnov_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+// Workaround for Doxygen: A header file that does not declare namespaces is to
+// be ignored if and only if it is processed stand-alone
+#undef _DOXYGEN_IGNORE_HEADER_FILE
+#include "kolmogorov_smirnov_test.hpp"
+
+/**
+ * @brief Transition state for Kolmogorov-Smirnov-Test functions
+ *
+ * Note: We assume that the DOUBLE PRECISION array is initialized by the
+ * database with length 7, and all elemenets are 0.
```
cwelton: Like chi_squared I am concerned about the assumption of what the input looks like for user callable functions.

src/modules/stats/kolmogorov_smirnov_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    if (state.expectedNum != expectedNum) {
+        if (state.num.sum() > 0)
+            throw std::invalid_argument("Number of samples must be constant "
+                "parameters.");
+        
+        state.expectedNum = expectedNum;
+    }
+    
+    if (state.last > value && state.num.sum() > 0)
+        throw std::invalid_argument("Must be used as an ordered aggregate.");
```
cwelton: Not just as an ordered aggregate, but also ordered by the same expression as the input parameter.  

E.g. this is good: `select agg(x order by x) ...`, whereas these are not `select agg(x order by x desc)`, `select agg(x order by y)`.

It would be good if the database had a better way to express the relevant properties so that a user can call the function simply, unfortunately it does not.

src/modules/stats/mann_whitney_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+// Workaround for Doxygen: A header file that does not declare namespaces is to
+// be ignored if and only if it is processed stand-alone
+#undef _DOXYGEN_IGNORE_HEADER_FILE
+#include "mann_whitney_test.hpp"
+
+/**
+ * @brief Transition state for Mann-Whitney-Test functions
+ *
+ * Note: We assume that the DOUBLE PRECISION array is initialized by the
+ * database with length 7, and all elemenets are 0.
```
cwelton: Same comment regarding user supplied parameters.

src/modules/stats/mann_whitney_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+    int sample = args[1].getAs<bool>() ? 0 : 1;
+    double value = args[2].getAs<double>();
+    
+    if (state.last < value) {
+        state.numTies.setZero();
+    } else if (state.last == value) {
+        for (int i = 0; i <= 1; i++)
+            state.rankSum(i) += state.numTies(i) * 0.5;
+    } else if (state.num.sum() > 0) { // also satisfied here: state.last > value
+        throw std::invalid_argument("Must be used as an ordered aggregate.");
```
cwelton: Same comment regarding use as an ordered aggregate.

src/modules/stats/kolmogorov_smirnov_test.cpp
cwelton: Additionally ordered aggregates don't scale particularly well due to the need of establishing a single global ordering.  Not that I see any immediate alternatives.

src/modules/stats/kolmogorov_smirnov_test.cpp
cwelton: Also s/elemenets/elements/

src/modules/stats/one_way_anova.cpp
```diff
@@ -0,0 +1,293 @@
...
+// Workaround for Doxygen: A header file that does not declare namespaces is to
+// be ignored if and only if it is processed stand-alone
+#undef _DOXYGEN_IGNORE_HEADER_FILE
+#include "one_way_anova.hpp"
+
+/**
+ * @brief Transition state for one-way ANOVA functions
+ *
+ * Note: We assume that the DOUBLE PRECISION array is initialized by the
+ * database with length 1, and all elemenets are 0.
```
cwelton: Same comment regarding user supplied parameters and elemenets

src/modules/stats/one_way_anova.cpp
```diff
@@ -0,0 +1,293 @@
...
+     * We define this function so that we can use TransitionState in the argument
+     * list and as a return type.
+     */
+    inline operator AnyType() const {
+        return mStorage;
+    }
+    
+    /**
+     * @brief Return the index (in the num, sum, and coorected_square_sum
+     *     fields) of a group value
```
cwelton: s/coorected/corrected/

src/modules/stats/one_way_anova.cpp
```diff
@@ -0,0 +1,293 @@
...
+    }
+    return pos;
+}
+
+template <>
+uint32_t
+OWATransitionState<AbstractionLayer::MutableArrayHandle<double> >::idxOfGroup(
+    const Allocator& inAllocator, uint16_t inValue) {
+    
+    // FIXME: Think of using proper iterators. Add some safety. Overflow checks.
```
cwelton, line 118: Jiras for known issues remaining.

Edit: I don't see any explicit issues in the code below related to overflow issues, but the comment concerns me.  If you do not believe this code is safe then it isn't ready for merge.

src/modules/stats/one_way_anova.cpp
```diff
@@ -0,0 +1,293 @@
...
+            
+            num.segment(0, oldSelf.numGroups) << oldSelf.num;
+            sum.segment(0, oldSelf.numGroups) << oldSelf.sum;
+            corrected_square_sum.segment(0, oldSelf.numGroups) << oldSelf.corrected_square_sum;
+        }
+    }
+    return posToIndices[pos];
+}
+
+// FIXME: Same function used for t_test. Factor out.
```
cwelton, line 169: Jiras for known issues remaining

src/modules/stats/one_way_anova.cpp
```diff
@@ -0,0 +1,293 @@
...
+template <>
+uint32_t
+OWATransitionState<AbstractionLayer::ArrayHandle<double> >::idxOfGroup(
+    const Allocator&, uint16_t inValue) {
+    
+    uint16_t pos = std::lower_bound(
+        groupValues, groupValues + numGroups, inValue) - groupValues;
+
+    if (pos >= numGroups || groupValues[pos] != inValue) {
+        // Did not find this group value. We have to start a new group.
```
cwelton, line 106: I'm a bit confused by the "we have to start a new group" comment in combination with throwing a runtime_error.

src/modules/stats/t_test.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+struct internal : public AbstractionLayer {
+    static AnyType tStatsToResult(double inT, double inDegreeOfFreedom);
+};
+
+/**
+ * @brief Transition state for t-Test functions
+ *
+ * Note: We assume that the DOUBLE PRECISION array is initialized by the
+ * database with length 6, and all elemenets are 0.
```
cwelton, line 40: Same comment about user supplied arguments and elemenets

src/modules/stats/t_test.cpp
cwelton, line 82: Good to add to the comments, would also be good to note that further improvement would be desirable, e.g. Ogita et al.

src/modules/stats/wilcoxon_signed_rank_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    WSRTestTransitionState<MutableArrayHandle<double> > state = args[0];
+    double value = args[1].getAs<double>();
+    
+    // Ignore values of zero.
+    if (value == 0)
+        return state;
+    
+    int sample = value > 0 ? 0 : 1;
+    
+    // FIXME: The following epsilon is hard-coded
```
cwelton: Yes it is.  What fix would you propose?

src/modules/stats/mann_whitney_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+ */
+AnyType
+mw_test_transition::run(AnyType &args) {
+    MWTestTransitionState<MutableArrayHandle<double> > state = args[0];
+    int sample = args[1].getAs<bool>() ? 0 : 1;
+    double value = args[2].getAs<double>();
+    
+    if (state.last < value) {
+        state.numTies.setZero();
+    } else if (state.last == value) {
```
cwelton: Is there a reason you perform floating point equality here, but allow for an epsilon in wsr_test_transition?

cwelton:
My review is complete.

1) I have concerns about direct invocation of the aggregate sub-functions due to assumptions about the transition state that are never verified.

2) For outstanding issues it would be good to have jiras tracking them so that they do not become lost and forgotten.

3) Some inconsistencies between behavior and documentation.

4) A couple small typos.

src/modules/stats/one_way_anova.cpp
```diff
@@ -0,0 +1,293 @@
...
+        
+        uint16_t numGroupsReserved = utils::nextPowerOfTwo(
+            static_cast<uint16_t>(numGroups));
+        if (numGroupsReserved > numGroups) {
+            // We have enough reserve space allocated.
+            std::copy(groupValues + pos, groupValues + numGroups,
+                groupValues + pos + 1);
+            groupValues[pos] = inValue;
```
haradh1, line 132: This is an insertion sort, which may become a bottleneck.

src/modules/stats/t_test.cpp
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/**
+ * @brief Perform the pooled (i.e., assuming equal variances) two-sample t-Test
+ *     final step
+ */
+AnyType
+t_test_two_pooled_final::run(AnyType &args) {
+    TTestTransitionState<ArrayHandle<double> > state = args[0];
+
```
haradh1, line 204: For t_test_two_pooled_final, t_test_two_unpooled_final, and t_test_final, is there no need to check to return NULL as in t_test_one_final?

src/modules/stats/kolmogorov_smirnov_test.cpp
haradh1: And looks like the required length is 6.

src/ports/postgres/modules/stats/hypothesis_tests.sql_in
```diff
@@ -0,0 +1,808 @@
...
+ *    ORDER BY <em>value</em>
+ *)).* FROM <em>source</em></pre>
+ *
+ * @note
+ *     This aggregate must be used as an ordered aggregate
+ *     (<tt>ORDER BY \em value</tt>) and will raise an exception if values are
+ *     not ordered.
+ */
+CREATE
+m4_ifdef(`__GREENPLUM__',`ORDERED')
```
haradh1, line 544: If we support stats in GPDB 4.0, this will fail.

haradh1:
Done my quick review.  Assign back to Florian.

src/ports/postgres/modules/stats/hypothesis_tests.sql_in
cwelton, line 544: Good point.  We should document and not install the unsupported module on GPDB 4.0.  Or... have some way of identifying which portions of the module are not supported and skip only those portions.

src/modules/stats/chi_squared_test.cpp
fschopp: This assumption here is meant as a precondition for the correct outcome (in a mathematical sense). If this precondition is not satisfied, e.g., because the function is called directly, the result will not be correct. The C++ AL will raise an exception when an out-of-bounds array access occurs.

That said, your concern raises a valid point in that there are cases where the code takes the address of an array element (with bounds checking), but then accesses memory relative to that pointer (no bounds checking any more). This is the case when Eigen objects are mapped to different locations in a single array. I'll check those cases and fix if necessary.

src/config/Version.yml
fschopp, line 1: Yes, we might want to factor this out into a separate commit.

src/modules/stats/chi_squared_test.cpp
fschopp, line 80: MADLIB-500, MADLIB-501

src/modules/stats/chi_squared_test.cpp
fschopp: Probably the right thing to do is to raise an error if `df <= 0`. Also, we should make `df` an optional argument (which currently means defining an overloaded UDA).
_Update_: Implemented as I suggested.

src/modules/stats/kolmogorov_smirnov_test.cpp
fschopp: Unfortunately, anything that requires the rank of values also requires a sort. The only alternative would be online approximation algorithms. Though I am doubtful that that would be easy to parallelize. Research on anything like "online multi-stream Kolmogorov-Smirnov test approximation algorithms" is probably limited or non-existent...

src/modules/stats/one_way_anova.cpp
fschopp, line 132: True, but I don't think the number of groups will ever become significant (but rather be in the single-digit range, or at most two-digit range). Otherwise, you are right, and I made a wrong assumption. Are you aware of use cases with more groups?

src/modules/stats/one_way_anova.cpp
fschopp, line 106: Good catch. This is comment is a left-over from copied code, and it is clearly confusing and inappropriate. The comment should read sth. like: "Did not find this group value. Since the underlying state is immutable, we will raise a runtime error. Obviously, this error should never occur."

src/modules/stats/t_test.cpp
fschopp, line 82: Yes, this should point to MADLIB-499 and subtasks.

src/modules/stats/wilcoxon_signed_rank_test.cpp
fschopp: I'd rather express it as a function of the machine epsilon and give an explanation for this expression.

src/modules/stats/mann_whitney_test.cpp
fschopp: Not anything I remember. The WSR epsilon test might have been motivated by me testing some sample data. I wrote all this in one week, so there was not much time, and this just seems inconsistent. I'll check again before merging.

src/ports/postgres/modules/stats/hypothesis_tests.sql_in
fschopp, line 544: I assume ordered aggregates are the only unsupported feature in GPDB 4.0. We thus could omit all rank-based tests, or we simply omit all hypothesis tests. Alternatively, we could leave them in, but the user would still have to ensure that the aggregate is seeing the values in order (which may or may not be possible using workarounds).

src/modules/stats/t_test.cpp
fschopp, line 204: True, that should be done.

src/modules/stats/one_way_anova.cpp
haradh1, line 132: No, actually I don't have any idea of the expected number of groups.  If it's only two-digits at most, it may be ok.   It is always true my assumption doesn't work when some crazy man shows up, which bothers me :).  Especially performance issues.

src/modules/stats/kolmogorov_smirnov_test.cpp
fschopp: True. This is, BTW, an example where directly calling the function using an array of length 5 would lead to undefined behavior.

src/modules/stats/mann_whitney_test.cpp
fschopp: No problem here, bounds will be checked.

src/modules/stats/one_way_anova.cpp
fschopp: Out-of-bounds access possible here. Needs fix.

src/modules/stats/one_way_anova.cpp
fschopp, line 118: The issue here is that the transition state contains pointers within the transition-state array, but no bounds-checking is currently performed. Probably not the only such place in MADlib, and there is MADLIB-499 anyway. But yes, some input validation should be added for now.

src/modules/stats/one_way_anova.cpp
fschopp, line 169: [MADLIB-500](http://jira.madlib.net/browse/MADLIB-500)

src/modules/stats/t_test.cpp
fschopp, line 40: No issue here. Bounds will be checked.

src/modules/stats/chi_squared_test.cpp
fschopp: For args[2], the expected number of observations, this error msg is correct. You are thinking about args[3], the degree of freedom.

src/modules/stats/mann_whitney_test.cpp
fschopp: Replaced by almost-equal testing.

src/modules/stats/mann_whitney_test.cpp
fschopp: Fixed:

```
    throw std::invalid_argument("Must be used as an ordered aggregate, "
        "in ascending order of the second argument.");
```

src/modules/stats/one_way_anova.cpp
fschopp: Fixed.

src/modules/stats/wilcoxon_signed_rank_test.cpp
fschopp: Fixed. Added a new function utils::almostEqual().

src/ports/postgres/modules/stats/hypothesis_tests.sql_in
fschopp, line 544: For now, module stats is deactivated on GP4.0.

src/modules/stats/wilcoxon_signed_rank_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    const double epsilon = 1e-10;
+    
+    if (std::fabs(state.last) + epsilon < std::fabs(value)) {
+        state.numTies.setZero();
+    } else if (std::fabs(std::fabs(state.last) - std::fabs(value)) < epsilon) {
```
fschopp: Replaced by:

```
// For almostEqual, we choose a precision of 2 * 3 units in the last place.
// This is because we assume that value is the result of adding up to three
// values (in the dependent paired test, value may be computed as
// "first - second - mu_0").
if (utils::almostEqual(std::fabs(state.last), std::fabs(value), 3) {
    [...]
```

_Update_: Please disregard this comment.

fschopp: Hmm, this one is actually problematic. If `value` is the result of a subtraction, then we are facing the problem of loss of significance. (We obviously know the difference with much less precision than we knew the original values.) From a numerical perspective, the only clean solution is to change the signature and have three parameters `first`, `second`, and `mu_0`. Alternatively, the user could provide a different argument that tells us the precision with which the differences are known.

fschopp:
_I have made the following modifications before merging. The biggest change involves a change in functionality and is shown in bold font below. I'll merge already now to facilitate QA, but I suggest a quick post-merge review of the latest commit as well._

Changes resulting from discussion around pull request 118.
Hypothesis tests:
- Out-of-bounds checks for transition states
- t-test/f-test: UDAs return NULL in case of insufficient data
- **Wilcoxon-signed-rank test/Mann-Whitney test: Updated testing of ties. For WSR tests, I added a new precision parameter. This parameter has to be used in the unit test.**
- Disabled module on GPDB4.0 because ordered aggregates are required
- If UDAs are not properly used as ordered aggregates, the error messages are now more informative.
- UDA chi2_gof_test now has default arguments

k-means:
- Fixed a wrong schema name in test

PR #119, opened 2012-02-17
Fixes for madlib-381/382

renyi533:
Fix for madlib-381/382: Out of memory FOR C45_TRAIN on GPDB4.1.

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.sql_in
+++ b/methods/decision_tree/src/pg_gp/decision_tree.sql_in
@@ -1874,11 +1874,11 @@ BEGIN
                  SET lmc_nid = g.lmc_nid, lmc_fval = g.lmc_fval
                  FROM  
                     (
-                    SELECT DISTINCT ON(parent_id) parent_id,
-                           id as lmc_nid, 
-                           tree_location[array_upper(tree_location,1)] as lmc_fval
+                    SELECT parent_id,
+                           min(id) as lmc_nid, 
+                           min(tree_location[array_upper(tree_location,1)]) as lmc_fval
                     FROM %
-                    ORDER BY parent_id, id
+                    GROUP BY parent_id
                     ) g
                 WHERE k.id = g.parent_id',
                 ARRAY[
@@ -1955,7 +1955,7 @@ DECLARE
     ret                         MADLIB_SCHEMA.c45_train_result;
     selection_visit             BOOLEAN[];
     selection_index             INT := 1;
-    nid                   		INT := 1;
+    nid               		INT := 1;
     best_time_cmp               BOOLEAN := 'f';
 BEGIN   
     exec_begin = clock_timestamp();
@@ -2155,7 +2155,7 @@ m4_changequote(>>>`<<<, >>>'<<<)
                             '], 0, 1, 1, 1, 1, '                || 
                             answer.node_id                      ||
                           ');';
-		        nid = nid + 1;
+		    nid = nid + 1;
                 END LOOP;
                 
                 SELECT MADLIB_SCHEMA.__format
@@ -2190,8 +2190,8 @@ m4_changequote(>>>`<<<, >>>'<<<)
                             array_to_string(temp_location, ',') ||
                             '], 0, 1, 1, 1, 1, '                || 
                             answer.node_id                      ||
-                      ');';
-                nid = nid + 1;
+                         ');';
+                    nid = nid + 1;
                 END LOOP;
 
                 SELECT MADLIB_SCHEMA.__format
@@ -3358,8 +3358,6 @@ DECLARE
     ret                     TEXT := '';
     tree_location           INT[];
     feature                 INT;
-    lmc_nid                 INT;
-    lmc_fval                INT;
     maxclass                INT;
     cat_size                INT;
     is_feature_cont         BOOLEAN;
@@ -3370,6 +3368,7 @@ DECLARE
     max_child_nid           INT;
     probability             FLOAT;
     curstmt                 TEXT;
+    child_nid               INT;
 BEGIN
     IF (id IS NULL OR id <= 0) THEN
         RETURN ret;
@@ -3377,7 +3376,7 @@ BEGIN
 
     SELECT MADLIB_SCHEMA.__format
             (
-                'SELECT tree_location, feature, lmc_nid, lmc_fval,is_feature_cont, 
+                'SELECT tree_location, feature, is_feature_cont, 
                         split_value, maxclass,cat_size,probability
                  FROM %
                  WHERE id = %',
@@ -3388,7 +3387,7 @@ BEGIN
              )
     INTO curstmt;
     
-    EXECUTE curstmt INTO tree_location, feature, lmc_nid, lmc_fval, is_feature_cont, 
+    EXECUTE curstmt INTO tree_location, feature, is_feature_cont, 
                          temp_split_value, maxclass, cat_size, probability;  
 
     curr_value = tree_location[array_upper(tree_location,1)];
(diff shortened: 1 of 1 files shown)


aojwang:
The changes are below.
1) Remove the jump column, which is an array, from the result tree table, since it maybe very big if the best split attribute has a large number of distinct values.

2) Generate the node ID sequentially on master so that we will not depend on the serial type.

3) As the child nodes' IDs of a node is continuous, we just need to remember the left most child node id (lmc_nid) and  the feature value which goes to the left most child node(lmc_fval). 
For example, assume that the left most child node id is i (lmc nid = i), and the feature value which leads to the lmc is j (lmc fval = j). Given a feature value x, then it will lead to the node whose id is i + x - j.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -183,16 +183,16 @@ sql> SELECT * FROM MADLIB.c45_train(
...
-  6 | {0,2,2}       |       4 |                 1 |         1 |        1 |                 0 |    0 |        2 |         3 |               | f               |            
-  7 | {0,3,1}       |       4 |                 1 |         1 |        2 |                 0 |    0 |        2 |         4 |               | f               |            
-  8 | {0,3,2}       |       4 |                 1 |         1 |        1 |                 0 |    0 |        3 |         4 |               | f               |            
+ id | tree_location | feature |    probability    | ebp_coeff | maxclass |    split_gain     | live | cat_size | parent_id | lmc_nid | lmc_fval | is_feature_cont | split_value 
+----+---------------+---------+-------------------+-----------+----------+-------------------+------+----------+-----------+---------+----------+-----------------+-------------
+  1 | {0}           |       3 | 0.642857142857143 |         1 |        2 | 0.171033941880327 |    0 |       14 |         0 |       2 |        1 | f               |
```
haradh1, line 188: I don't see why lmc_fval here is ok with 1.  From the delete line above, it seems 2, no?

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -1901,43 +1856,38 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__generate_final_tree
...
+    
+    -- for each node, find the left most child node id and the feature value, and
+    -- update the node's lmc_nid and lmc_fval column     
+    SELECT MADLIB_SCHEMA.__format
+            (
+                'UPDATE % k
+                 SET lmc_nid = g.lmc_nid, lmc_fval = g.lmc_fval
+                 FROM  
+                    (
+                    SELECT DISTINCT ON(parent_id) parent_id,
```
haradh1: It seems this query can be optimized as

``` sql
SELECT parent_id, MIN(lmc_id), MAX(lmc_fval) ... GROUP BY parent_id
```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -3468,20 +3436,40 @@ BEGIN
...
+                    tree_table,
+                    MADLIB_SCHEMA.__to_char(id)
+                    ]
+            )
+    INTO curstmt;   
+    EXECUTE curstmt INTO max_child_nid;
+    
+    -- if current node has children, then recursively display
+    node_count = 1;
+    WHILE (lmc_nid IS NOT NULL) AND (0 < node_count AND lmc_nid <= max_child_nid) LOOP
```
haradh1: Alternatively you can fetch all the children node IDs at one time so that you don't need to EXECUTE the the end of loop repeatedly.  In addition, I don't understand why you need to check node_count every time, since the loop condition should be FOR lmc_nid IN min_lmc_nid .. max_chlid_nid.  Am I missing something?

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -3823,7 +3815,35 @@ BEGIN
...
+        -- if the node (whose id is "jump") doesn't exist, 
+        -- then insert them into result table 
+        -- (be classified to max_class of its corrsponding node)
+        SELECT MADLIB_SCHEMA.__format(
+            'INSERT INTO %(id, jump, class, prob, parent_id, leaf_id) 
+             SELECT id, 0, class, prob, parent_id, leaf_id
+             FROM %
+             WHERE jump NOT IN (SELECT id FROM %)',
+            ARRAY[
+                result_table_name,
```
haradh1, line 3819: What is the column "jump" here for?  Isn't it related to jump in tree_table?

By the way, where is the result table's comments/descriptions?  I don't find it even in the madlib document.

haradh1:
Overall the change is straightforward, but there is not justification why this change fixes MADLIB-381.  Can you explain how many elements were in the jump array when the issue occurred?  And are we sure that the jump array actually has only continuous values and we can transform it to the set of min and max?

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang, line 188: The correct value for lmc_fval is 1. 
The jump array is organized as following. 
1) jump[i] =x represents the value (i -1) will lead to the xth node.
2) value 0 means a missing value. As the array subscript of pl/pqsql starts from 1, we add 1 to the feature's value.

Therefore, assume that the best attribute for the root node has three distinct values: 1, 2, and 3; and value 1 will go to the second node; value 2 will go to the 3rd node; value 3 will go to the 4th node.

haradh1, line 188: That's an explanation for the jump array.  My question is about lmc_fval definition.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang: Thanks. As the lmc_nid and lmc_fval are the minimum node id and feature value, i will change it to the following code. 

``` javascript
SELECT parent_id,
               min(id) as lmc_nid,
               min(tree_location[array_upper(tree_location,1)]) as lmc_fval
FROM %
GROUP BY parent_id
```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang: You are right. Done. Thanks very much.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang, line 3819: This jump column is not related to the tree_table, it means which node the case will go in the next level. For example, the jump of all the cases is 1 (root node ID), then each case will go to different node with our level-by-level algorithm.

If you see the comments of c45_classify (see below), we give the description of the classification result table. As we store the encoded values for class column (its type is INT), users need to join the table with encoded table to get the real class value. This maybe not convenient. Thus, in our latest code, the type of class column is the same as that of class column in training table. Furthermore, remove the prob column, since it's useless. Thanks. 

``` javascript
 * @return The columns of classified table:
 *  - <tt>id INT</tt>     - Record id.
 *  - <tt>class INT</tt>  - Predicted class.
 *  - <tt>prob FLOAT</tt> - Probability of the predicted class.
 */
DROP FUNCTION IF EXISTS MADLIB_SCHEMA.c45_classify

```

aojwang:
For discrete attribute, we encode all the distinct values with continuous numbers. For continuous, we only have two values 1 (<=split value) and 2 ( > split value).  Therefore the size of jump array equals to (the number of distinct values  + 1) for discrete attribute ( jump[1] is for missing value), and equals to 2 for continuous attribute.

Assume that the current best split attribute is a discrete attribute, which has 1000 distinct values (1, 2, ..., 1000). The worse case is, that the values feed to the aggregate is in ascending order (1, 2, ..., 1000). Furthermore, assume that the 1000 values will lead to the nodes whose IDs are from 10 to 1010. Then we can get:
jump[2] = 10            -- I am not sure how many size of memory will be allocate if we didn't specify the array size, but each time we will have an assignment for the array (temp = $1;), then  add a new element to the array ( temp[$2+1] = $3;)
jump[3] = 11            -- should enlarge the array size
...
jump[1001] = 1010 -- should enlarge the array size

Therefore, I think there are a lot of memory allocation and free when we call this aggregate. And if the best attribute has a large number of distinct values, then the memory allocation should be very large from the experiment.  We think that use plc to write this function which can be more efficiently. However, on one hand, we need to use an array to store the jump IDs, if there is a lot of distinct values, maybe it's not unacceptable. On the other hand, consider our tree node IDs for a parent node are continuous. Therefore, we just store the left most child node ID and the value lead to this node. Thanks.

``` javascript
CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__jump_aggr_sfunc
(
    INT[],
    INT,
    INT
)
RETURNS INT[] AS $$
DECLARE
        temp INT[];
BEGIN
        temp = $1;
        temp[$2+1] = $3;

        RETURN temp;
END
$$ LANGUAGE PLPGSQL;

```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
aojwang, line 188: Sorry, maybe it's not very clear. As mention in jump array. In the old code, as we will add 1 to the feature value, the lmc_fval should be (2 -1) = 1. In the new code, lmc_fval is the real feature value which leads to the left most child node. Therefore, the lmc_fval is 1. Thanks.

haradh1, line 188: Ah, now I think I understand.  I thought it is for the length of the child nodes.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
haradh1, line 3819: Oh, I see.  I thought other modules would have the table definition of the result in the "about" page, but it seems no modules have ones. Now it's clear.

haradh1:
While attempting merge this, I found this fails in install-check.

<pre>
psql:/tmp/madlib.7Qm47C/decision_tree/test/dec_trees.sql_in.tmp:805: ERROR:  array subscript in assignment must not be null
</pre>


I'm testing it against postgresql 9.1.  Holding this merge until it's solved.

haradh1:
Just got the updated version of decision_tree.sql_in from aojwang, and the test passed.  The request is pulled to the master with editorialization by me.

PR #120, opened 2012-02-17
Fix for madlib-317: DT will create global table and it has privelege issue

renyi533:
1. Add an additional parameter "target_schema_name" to the c45_train function. 

```
CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.c45_train
    (
    split_criterion             TEXT,
    training_table_name         TEXT, 
    result_tree_table_name      TEXT,
    validation_table_name       TEXT, 
    continuous_feature_names    TEXT, 
    feature_col_names           TEXT, 
    id_col_name                 TEXT, 
    class_col_name              TEXT, 
    confidence_level            FLOAT,
    how2handle_missing_value    TEXT, 
    target_schema_name          TEXT
    max_num_iter                INT, 
    max_tree_depth              INT, 
    min_percent_mode            FLOAT,
    min_percent_split           FLOAT, 
    verbosity                   INT
    )
```
1. Put the "training_info" table to the MADLIB_SCHEMA. And store the "target_schema_name" in the training_info table.
2. If the result tree table name contains schema name, then it must be the same with the target schema name.
   Otherwise, the result tree table will be under the target schema.
3. Put the trained tree table and auxiliary tables (such as encoded table, data info table, key-value tables, etc.) to the "target_schema_name" schema.

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.sql_in
+++ b/methods/decision_tree/src/pg_gp/decision_tree.sql_in
@@ -695,7 +695,7 @@ m4_changequote(>>>`<<<, >>>'<<<)
                         ) AS t1
                     CROSS JOIN
                     (
-                        SELECT key FROM MADLIB_SCHEMA.%
+                        SELECT key FROM %
                     ) as t2
                 ) AS n1 
                 LEFT JOIN % n2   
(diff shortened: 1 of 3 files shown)


renyi533:
This solution is not acceptable. I will close this pull request. We need a better solution for madlib-317

PR #121, opened 2012-02-24
Additions to probability module in support of statistical hypothesis tests.

fschopp:
Most significant changes:
- Changed order of argument: Parameters come last now because they often have reasonable defaults and thus are optional
- Now active and being installed
- Include Fisher F distribution function
- Inlcude Kolmogorov distribution function (from CERN Root project)
- All CDFs now return 0 for -infinity and 1 for +infinity.

And other smaller but related changes.


src/modules/prob/student.cpp
```diff
@@ -120,35 +123,41 @@
...
+ * @param nu Degree of freedom (>0)
  *
  * Note: The running time of calculating the series is proportional to nu.
  * We therefore use the normal distribution as an approximation for large nu.
  * Another idea for handling this case can be found in reference [8].
  */
 
-double studentT_CDF(int64_t nu, double t) {
-	double		z,
+double studentT_CDF(double inT, double inNu) {
```
haradh1, line 134: The variable names are different from the comment above and UDF declaration.

src/modules/prob/kolmogorov.cpp
```diff
@@ -0,0 +1,159 @@
...
+// END Copied from CERN ROOT, math/mathcore/src/TMath.cxx
+
+// BEGIN Copied from CERN ROOT, math/mathcore/src/TMath.cxx
+// (SVN Rev. 41830, ll. 657-715)
+Double_t TMath::KolmogorovProb(Double_t z)
+{
+// Calculates the Kolmogorov distribution function,
+   //Begin_Html
+   /*
+   <img src="gif/kolmogorov.gif">
```
haradh1, line 99: Is this included in the commit?

haradh1, line 99: Ah, it's just a copy and we never see it?

fschopp, line 99: You mean math/mathcore/src/TMath.cxx? There is a link at the top of the source file if you want to see it. I currently don't know if this is the wrong way to embed an excerpt of LGPL'd code. Do you?

haradh1, line 99: No, I mean just kolmogorov.gif.  I thought it'd be able to be seen as doxygen and so we needed it in the tree, but this part is out of doxygen, right?

fschopp, line 99: Yes, the CERN Root project is using some other documentation generation. Doxygen will not pick up or generate anything here.

src/ports/postgres/dbconnector/main.cpp
```diff
@@ -18,4 +18,5 @@
 // Now export the symbols
 #undef DECLARE_UDF
 #define DECLARE_UDF DECLARE_UDF_EXTERNAL
+#define DECLARE_LIBRARY_EXPORTS
```
haradh1, line 21: Why do we need this?  It seems to me you can always export them.

haradh1, line 21: Or you can just switch #ifdef DECLARE_UDF in the other headers if it's not from main.cpp.  I assume DECLARE_UDF is void if DECLARE_LIBRARY_EXPORTS is not defined.

src/modules/prob/kolmogorov.cpp
```diff
@@ -0,0 +1,159 @@
...
+   // This function returns the confidence level for the null hypothesis, where:
+   //   z = dn*sqrt(n), and
+   //   dn  is the maximum deviation between a hypothetical distribution
+   //       function and an experimental distribution with
+   //   n    events
+   //
+   // NOTE: To compare two experimental distributions with m and n events,
+   //       use z = sqrt(m*n/(m+n))*dn
+   //
+   // Accuracy: The function is far too accurate for any imaginable application.
```
cwelton, line 117: I'm amused by the "far too accurate" comment.

src/modules/prob/student.cpp
```diff
@@ -120,35 +123,41 @@
...
-	else if (nu >= 1000000)
-		return normal_cdf(t);
-	else if (nu >= 200)
-		return studentT_cdf_approx(nu, t);
+	else if (inNu >= 1000000)
+		return normalCDF(t);
+	else if (inNu >= 200)
+		return studentT_cdf_approx(t, inNu);
+    
+    /* inNu is non-negative here, so nu will be the closest integer */
```
cwelton, line 154: Indentation here is misleading, we are no longer in the else if block, but our indentation is the same.

src/modules/prob/student.cpp
```diff
@@ -120,35 +123,41 @@
...
+    int64_t    nu = std::floor(inNu + 0.5);
+	
+    if (std::fabs(inNu - nu)/inNu > 0.01)
```
cwelton, line 158: Is 0.01 the right threshold?

src/modules/prob/test/student.sql_in
```diff
@@ -19,4 +19,4 @@ select MADLIB_SCHEMA.student_t_cdf(2147483647, 1.79e308)::REAL;
 -- select MADLIB_SCHEMA.student_t_cdf(-2147483648, 0.000000000001)::REAL;
 
 select MADLIB_SCHEMA.student_t_cdf(1, 2000)::REAL - (1./2 + 1/pi() * atan(2000))::REAL AS shouldBeZero;
-select MADLIB_SCHEMA.student_t_cdf(2,-3)::REAL - (1./2 * (1. + (-3.)/sqrt(2. + pow(-3, 2))))::REAL AS shouldBeZero;
\ No newline at end of file
+-- select MADLIB_SCHEMA.student_t_cdf(2,-3)::REAL - (1./2 * (1. + (-3.)/sqrt(2. + pow(-3, 2))))::REAL AS shouldBeZero;
\ No newline at end of file
```
cwelton: Need tests for the other distribution functions.

src/modules/prob/test/student.sql_in
```diff
@@ -9,9 +9,9 @@ select MADLIB_SCHEMA.student_t_cdf(NULL, 12)::REAL;
 -- select MADLIB_SCHEMA.student_t_cdf(0, 0)::REAL; 
 -- select MADLIB_SCHEMA.student_t_cdf(-2, 1000000)::REAL;
 -- select MADLIB_SCHEMA.student_t_cdf(0, -2)::REAL;
```
cwelton: It would, of course, be highly preferable to have a test suite that is able to run the negative tests.

src/modules/prob/student.cpp
```diff
@@ -232,11 +227,11 @@ double studentT_cdf_approx(int64_t nu, double t)
  */
 AnyType
 student_t_cdf::run(AnyType &args) {
-    int64_t nu = args[0].getAs<int64_t>();
-    double t = args[1].getAs<double>();
-        
+    double t = args[0].getAs<double>();
+    double nu = args[1].getAs<double>();
```
cwelton, line 232: I'm slightly concerned about the change in parameter order here since its a change in API.  However I do acknowledge that this is more consistent.

If we are going to include this then we should be sure to ultimately release note it, which leads me to think that we need a specific Jira on the change of parameter order.

src/modules/prob/student.cpp
fschopp, line 154: For whatever reason, there is a mixture of tabs and spaces here. Don't know how that happened. I'll fix that.

src/modules/prob/student.cpp
fschopp, line 232: Luckily, it was not exposed before. This function existed, but we never did a CREATE FUNCTION.

src/modules/prob/test/student.sql_in
fschopp: True, the tests are completely outdated.

cwelton:
Overall looks good.  Things I'd like to see:
- Tests for distribution functions other than student's T
- Jiras for new functions and changes in API for existing functions, targeted to the 0.4 release.

General question:
- Boost also supplies pdf and quantile functions for every distribution it supports, wouldn't this be a good addition to our own functionality here?  Is there a reason we are only exposing the cumulative form at this time?

src/modules/prob/student.cpp
cwelton, line 232: Ah, yes.  All is well then.

cwelton:
Incidentally the comment on the change request is "Probabiliy module".   s/Probabiliy/Probability/

cwelton:
Oops, didn't mean to close the PR, reopening.

src/modules/prob/student.cpp
fschopp, line 134: Thanks. Fixed.

src/ports/postgres/dbconnector/main.cpp
fschopp, line 21: No, DECLARE_UDF is a macro for declaring the C++ UDF implementation first, and then a macro for declaring the C export afterwards.

src/modules/prob/student.cpp
fschopp, line 158: Intuitively, this seems fine. But I have done no testing. I added a

```
// FIXME: Add some justification/do some tests.
```

src/modules/prob/test/student.sql_in
fschopp: There is now a function called check_if_raises_error() in the utilities module. This is used for the modified Student-t test. Eventually, we might want to make this function more fine-grained and make sure a particular exception is raised.

fschopp:
Tests redone. Should serve as an example for other probability functions for now. Eventually, the tests should cover even more, but basic testing is better than no (useful) testing at all.

PR #122, opened 2012-03-05
Fix for madlib-416: The decision tree algorithm does not accept temporary tables as training set

renyi533:
The decision tree algorithm currently do not accept temporary tables as training sets.

This pull request corrected this behavior.

### DIFF
--- a/methods/decision_tree/src/pg_gp/utility.sql_in
+++ b/methods/decision_tree/src/pg_gp/utility.sql_in
@@ -785,26 +785,11 @@ BEGIN
     str_val = btrim(split_part(full_name, '.', 2), ' ');
 
     IF( str_val is null or str_val = '' ) THEN
-        -- If user does not explicitly specify schema name,
-        -- tentatively set the schema name to public
         ret_val = 'public';
-
-        -- If there is no table belonging to public,
-        -- set the schema name to temporary schema.
-        IF (
-             NOT MADLIB_SCHEMA.__table_exists
-                (
-                    ret_val,
-                    MADLIB_SCHEMA.__strip_schema_name(full_name)
-                )
-           )  THEN
-           SELECT nspname FROM pg_catalog.pg_namespace 
-                WHERE oid=pg_my_temp_schema() INTO ret_val;
-        END IF;
     ELSE
         ret_val = btrim(split_part(full_name, '.', 1), ' ');
     END IF;
-    RETURN lower(ret_val);
+    RETURN ret_val;
 end
 $$ LANGUAGE PLPGSQL;
 
--- a/methods/decision_tree/src/pg_gp/utility.sql_in
+++ b/methods/decision_tree/src/pg_gp/utility.sql_in
@@ -297,111 +297,101 @@ $$ LANGUAGE PLPGSQL;
 
 /*
  * @brief Test if the specified table exists or not
- * @param schema_name   schema name of the table
- * @param table_name    the short table name without schema name
+ *
+ * @param full_table_name    The full table name
  *
  */
-DROP FUNCTION IF EXISTS MADLIB_SCHEMA.__table_exists
-    (
-    schema_name  TEXT,
-    table_name   TEXT
-    );
-    
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__table_exists
     (
-    schema_name  TEXT,
-    table_name   TEXT
+    full_table_name    TEXT
     ) 
 RETURNS BOOLEAN AS $$
 DECLARE
-    curstmt         TEXT := '';
-    result          INT := 0;
-    schema_name_c   TEXT := '';
-    table_name_c    TEXT := '';
+    --toid              OID;
+    table_name          TEXT;
+    temp                TEXT[];
+    len                 INT;
+    schema_names        TEXT[];
+    temp_schema         TEXT;
 BEGIN
-    PERFORM MADLIB_SCHEMA.__assert
-        (schema_name IS NOT NULL, 'Schema name must not be null!');
-        
-    PERFORM MADLIB_SCHEMA.__assert
-        (table_name IS NOT NULL, 'Table name must not be null!');
-    
-    IF (length(btrim(lower(schema_name), ' ')) = 0) THEN
-        schema_name_c = 'public';
-    ELSE
-        schema_name_c = schema_name;
+    IF (full_table_name IS NULL) THEN
+        RETURN 'f';
     END IF;
+
+    temp = string_to_array(full_table_name, '.');
+    len = array_upper(temp, 1);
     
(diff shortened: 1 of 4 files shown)


methods/decision_tree/src/pg_gp/utility.sql_in
```diff
@@ -785,11 +785,26 @@ BEGIN
...
+                )
+           )  THEN
+           SELECT nspname FROM pg_catalog.pg_namespace 
+                WHERE oid=pg_my_temp_schema() INTO ret_val;
+        END IF;
     ELSE
         ret_val = btrim(split_part(full_name, '.', 1), ' ');
     END IF;
-    RETURN ret_val;
+    RETURN lower(ret_val);
```
cwelton: This is wrong, per my comments in the Jira.

renyi533: Sometimes, we would like to call this  function with non-existent tables. For example, 'madlib.table_in_future' does not exist. But we would like to get the schema name, which is 'madlib'. The solution of letting the system resolve the name and determine the schema will not work.

If users specify the full name, our old function works well. But if users omit the schema name, the old function may return incorrect result. Is there any way for me to look up the search path? Then, I can return the correct result.

haradh1:
Why can't we just do

``` sql
SELECT table_name::regclass
```

for the table existence check?

haradh1:
OK.  I fount the jira and replied there.

renyi533:
Yes. We can use the below code to check the existense of table. But since we only allow 100 subtransactions, there will be error after calling this function 100 times. 

Even if we resolve this issue in 4.2.2, the other version still has this problem. Therefore, we find a solution that can work for all versions.

```
-- Currently, we only allow 100 subtransactions within a transaction.
-- we will not use this way
--BEGIN
--    SELECT full_table_name::regclass::oid INTO toid;
--EXCEPTION
--    WHEN undefined_table THEN
--    RETURN 'f';
--END;
```

renyi533:
The issue related with this bug can be reproduced below.
1. create temp table golf_temp as select \* from golf;
2. Use the temporary table as training set. (Do not specify the temporary table's schema, like pg_temp_10).
3. Since the original table_exists function does not check the tables under temporary schema, it will report error.

After I change the logic of __table_exists, it now can handle temporary tables.

renyi533:
There are three commits for this fix. Please ignore the first two commits. For the first two commits, I take a unsatisfactory solution by changing __get_shema_name instead of __table_exists.

In the second commit, I reverted the change for __get_schema_name. In the third commit, I changed __table_exists to resolve this issue.

haradh1:
Does it work with double-quoted names?

haradh1:
It should not necessarily be plpgsql.  With pl/python you can use try block and it'd be a few lines with C function.  I don't care which language it is, just correctness.

renyi533:
I verified it works with double-quoted names. Please see below.

testdb=# create temp table "TestTemp" (id int);
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column named 'id' as the Greenplum Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
CREATE TABLE
testdb=# 
testdb=# \d "TestTemp"
 Table "pg_temp_149.TestTemp"
 Column |  Type   | Modifiers 
--------+---------+-----------
 id     | integer | 
Distributed by: (id)
testdb=# select madlib.__table_exists('TestTemp');

##  __table_exists 

 t
(1 row)

testdb=# select madlib.__table_exists('testtemp');

##  __table_exists 

 f
(1 row)

renyi533:
Yes. We can use other language. But since we currently use plpgsql extensively for decision tree, it is preferable for us to use plpgsql. And I am sure our solution currently can work well.

haradh1:
But it doesn't with

<pre>
__table_exists('"TestTemp"')
</pre>


while it works with regclass cast.

renyi533:
The parameter of __table_exists is a string. If you want upper case letters, you just write 'TestTemp' not '"TestTemp"'.

Please see below.

testdb=# select 'TestTemp';

##  ?column? 

 TestTemp
(1 row)

testdb=# select '"TestTemp"';

##   ?column?  

 "TestTemp"
(1 row)

When you issue the command of CREATE TEMP TABLE "TestTemp". You create the table of 'TestTemp' not '"TestTemp"'. The table of '"TestTemp"' really does not exist so that the function returns false.

testdb=# select madlib.__table_exists('"TestTemp"');

##  __table_exists 

 f
(1 row)

testdb=# select madlib.__table_exists('TestTemp');

##  __table_exists 

 t
(1 row)

haradh1:
Double quote is not only for the upper/lower cases.  Think about CREATE TABLE public."foo.bar".  I don't like to point out the tiny corner cases for too long on this, and what I'm saying is let's use the existing mechanism as much as possible so that we don't need to be bothered with this kind of things.

haradh1:
More specifically, my suggestion is: (didn't test it)

``` c
PG_FUNCTION_ARGS(table_exists);
Datum
table_exists(PG_FUNCTION_ARGS)
{
    text            input;
    List           *name;
    Oid             relid;

    if (PG_ARGISNULL(0))
        PG_RETURN_BOOL(false);
    text =  PG_GETARG_TEXT_P(0);
    names = stringToQualifiedNameList(input, "regclassin");
    relid = RangeVarGetRelid(makeRangeVarFromNameList(names), true);
    PG_RETURN_BOOL(OidIsValid(relid));
}
```

``` sql
CREATE FUNCTION __table_exists(text) RETURNS bool AS
'$libdir/foo', 'table_exists' LANGUAGE C IMMUTABLE;
```

cwelton:
Hitoshi's suggestion looks reasonable.  Note that the second parameter to stringToQualifiedNameList is the "caller" of the function, so I would actually recommend:   `stringToQualifiedNameList(input, "table_exists");`

However the parameter isn't actually used, so it doesn't make much difference.

Note that the same technique would also provide a much better way to perform the schema name extraction as well.

haradh1:
Ah, right, for the second argument.  I wish we could have this kind of "common" utility in the common area, unlike dt_utility so that other methods like kmeans can use it.

cwelton:
Absolutely agree.  This is general architecture functionality and does not belong as DT/RF specific code.

renyi533:
Many thanks for your comments. I implemented it with C now. 

Please review it again.

renyi533:
Yes. This is a very general function. And some other functions are also very general.

We will have a dedicated pull request related with code refactoring to address such issues. It it OK for us to resolve all such issues then?

haradh1:
Pushed to master.

PR #123, opened 2012-03-07
fix madlib-434

renyi533:
Currently, the decision tree algorithm returns the result table of {Id, Class, Possibility}. For the column of 'Class', it contains the encoded information.

From the point of end users, the 'Class' column should contain information before encoding.

This pull request fix this issue.

### DIFF
--- a/methods/decision_tree/src/pg_gp/decision_tree.sql_in
+++ b/methods/decision_tree/src/pg_gp/decision_tree.sql_in
@@ -3871,6 +3871,9 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.c45_classify
 RETURNS MADLIB_SCHEMA.c45_classify_result AS $$
 DECLARE
     encoded_table_name  TEXT := '';
+    temp_result_table   TEXT := '';
+    metatable_name      TEXT;
+    result_rec          RECORD;
     begin_time          TIMESTAMP;
     ret                 MADLIB_SCHEMA.c45_classify_result;
 BEGIN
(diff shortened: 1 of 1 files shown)


renyi533:
This pull request includes two commits. 

The second commit only make it compatible with pull request 120. 

If pull request 120 is rejected, please just accept the first commit.

If pull request 120 is accepted before this one, please merge both of the two commits.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -3880,20 +3883,32 @@ BEGIN
...
+            ' WHERE column_type=''c'' LIMIT 1;'
+            INTO result_rec;
+    result_rec.table_name = 'MADLIB_SCHEMA.'||result_rec.table_name;
+
+   -- translate the encoded class information back
+    EXECUTE 'CREATE TABLE '||result_table_name||' AS SELECT n.id, m.'||result_rec.column_name
+        ||' as class,n.prob from '||temp_result_table||' n,'||result_rec.table_name
+        ||' m where n.class=m.key m4_ifdef(`__GREENPLUM__', `DISTRIBUTED BY (id)');';
+        
     EXECUTE 'DROP TABLE IF EXISTS ' || encoded_table_name || ';';
```
haradh1, line 3910: What is the encoded_table?? Isn't it result_table?

haradh1:
I think it's good to go, at least for this issue.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
renyi533, line 3910: We preprocess/encode the original training table and get the encoded table, which is just a temporary table. It is not result table.

haradh1, line 3910: The returned value of "classify_internal" is just confusing.  There's no description anywhere.

renyi533, line 3910: We can change or add comments in code refactoring. I think this issue is not closely related with this defect. 

Let's resolve it in a separate thread. OK?

haradh1, line 3910: Yes, this is off topic.

haradh1:
Is the comment for the Class column in the classify function still valid?

renyi533:
The comment of "<tt>class INT</tt> - Predicted class." is not valid.

It should be "<tt>class (The same type as the class column in training set)</tt> - Predicted class.".

I cannot access my machine now. I will modify next Monday.

renyi533:
I just modified the comments. Please push in the change.

haradh1:
Thanks.  Closing.

PR #126, opened 2012-04-05
MADLIB-351

fschopp:
Build system:
- Now uses {PG|GP}_VERSION_NUM instead of {PG|GP}_VERSION if available. (Main benefit: Now builds fine against development builds of Greenplum. Unfortunately, GP_VERSION does not contain the major build version but only "main build dev" by default.)

Madpack:
- When DBMS version cannot be determined because of a parsing error, default to the newest supported version of this DBMS. (Main benefit: Now installs fine on development build of Greenplum. Unfortunately "SELECT version()" does not return a major version number of development builds by default.)

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -268,14 +268,15 @@ def __get_madlib_dbver(schema):
 ## # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 def __get_dbver():
     try:
-        versionString = __run_sql_query("""SELECT version()""",
+        versionStr = __run_sql_query("""SELECT version()""",
             True)[0]['version']
         if portid == 'postgres':
-            return re.search("PostgreSQL[a-zA-Z\s]*(\d+\.\d+)",
-                versionString).group(1)
+            match = re.search("PostgreSQL[a-zA-Z\s]*(\d+\.\d+)",
+                versionStr)
         elif portid == 'greenplum':
-            return re.search("Greenplum[a-zA-Z\s]*(\d+\.\d+)",
-                versionString).group(1)
+            match = re.search("Greenplum[a-zA-Z\s]*(\d+\.\d+)",
+                versionStr)
+        return None if match is None else match.group(1)
     except:
         __error("Failed reading database version", True)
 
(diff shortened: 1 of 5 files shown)


src/madpack/madpack.py
```diff
@@ -802,16 +803,30 @@ def main(argv):
...
+        else:
+            __info("Detected %s version %s." % (ports[portid]['name'], dbver),
+                True)
+            if not os.path.isdir(os.path.join(portdir, dbver)):
+                __error("This version is not among the %s versions for which "
+                    "MADlib support files have been installed (%s)." % 
+                    (ports[portid]['name'],
+                    ", ".join([name for name in os.listdir(portdir)
+                                    if os.path.isdir(os.path.join(portdir, name))
+                                    and re.match("^\d+\.\d+", name)])
```
fschopp: Grumble. This was supposed to be "supportedVersions".

haradh1:
It looks good.  I wonder if we couldn't use pg_config at run time, too.

<pre>
haradh1-mac:cdb-pg haradh1$ grep GP_VERSION `pg_config --includedir`/pg_config.h
#define GP_VERSION "main build dev"
#define GP_VERSION_NUM 50000
</pre>

src/ports/postgres/cmake/FindPostgreSQL.cmake
```diff
@@ -106,16 +109,31 @@ if(${PKG_NAME}_PG_CONFIG AND ${PKG_NAME}_CLIENT_INCLUDE_DIR)
...
+                ".*#define ${_PG_CONFIG_VERSION_MACRO} \"([^\"]+)\".*" "\\1"   
+                ${PKG_NAME}_VERSION_STRING "${_PG_CONFIG_HEADER_CONTENTS}")
+            string(REGEX REPLACE "([0-9]+).*" "\\1"
+                ${PKG_NAME}_VERSION_MAJOR "${${PKG_NAME}_VERSION_STRING}")
+            string(REGEX REPLACE "[0-9]+\\.([0-9]+).*" "\\1"
+                ${PKG_NAME}_VERSION_MINOR "${${PKG_NAME}_VERSION_STRING}")
+            string(REGEX REPLACE "[0-9]+\\.[0-9]+\\.([0-9]+).*" "\\1"
+                ${PKG_NAME}_VERSION_PATCH "${${PKG_NAME}_VERSION_STRING}")
+        endif(${PKG_NAME}_VERSION_NUM MATCHES "^[0-9]+$")
+    endif(EXISTS "${${PKG_NAME}_CLIENT_INCLUDE_DIR}/pg_config.h")
```
haradh1, line 136: What about "else?"

fschopp, line 136: No need for else. In the "else" case some required variables will not be set, and cmake will give an appropriate error message. That's handled in find_package_handle_standard_args() below.

fschopp:
Some arguments against using pg_config.h at runtime:
- Permissions in file system need not to be congruent with in-db permissions
- Currently, we only need to connect via psql. Using pg_config.h would require to find out which pg_config belongs to the running PG instance.
- Some time in the future, madpack might be called remotely.

src/madpack/madpack.py
```diff
@@ -802,16 +803,30 @@ def main(argv):
...
+                                if os.path.isdir(os.path.join(portdir, dirItem))
+                                and re.match("^\d+\.\d+", dirItem)]
+        if dbver is None:
+            dbver = ".".join(map(str, max([map(int, versionStr.split('.'))
+                    for versionStr in supportedVersions])))
+            __info("Could not parse version string reported by {DBMS}. Will "
+                "default to newest supported version of {DBMS} "
+                "({version}).".format(
+                    DBMS = ports[portid]['name'],
+                    version = dbver
```
cwelton, line 817: So, if for some reason we gave '-p greenplum', but were pointing at a postgres instance it would fail the regex and conclude that it was the most recent version of Greenplum?

fschopp, line 817: The platform check is independent of the version check. Currently we have:

```
schopf$ src/bin/madpack -p greenplum -c :54322 version
Password for user schopf:
madpack.py : INFO : Could not parse version string reported by Greenplum DB. Will default to newest supported version of Greenplum DB (4.2).
madpack.py : ERROR : Invalid database platform specified.
```

We might want to swap the version and the platform check.

fschopp, line 817: In any case, I think madpack needs to be rewritten. We should reword [MADLIB-264](http://jira.madlib.net/browse/MADLIB-264) accordingly.

cwelton, line 817: Ah, fair enough.

PR #127, opened 2012-04-05
Fix for MADLIB-495: UnboundLocalError: local variable 'rv' referenced before assignment

cwelton:


src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -64,6 +64,7 @@ def __run_quietly(sql):
     except:
         # If there was a problem do this anyway
         plpy.execute("SET client_min_messages = %s;" % prev_msg_level)
+        raise
```
cwelton, line 67: Otherwise:
a) We exit this exception block without having assigned rv and the return statement will fail.
b) If, somehow, the return succeeded we would have consumed the error without any indication to the user, which would be wrong.

=> the correct thing to do is re-raise the exception after restoring client_min_messages.

fschopp:
Manual rebased merge.

PR #128, opened 2012-04-13
Support EXTENSION in PG 9.1+.

haradh1:
In madpack, if the target platform supports extension, it switches
to create extension files instead of installing with the old way.
EXTENSION in postgres requires two files in its install directory
under $SHAREDIR/extension, and madpack has been avoiding interaction
with platform's directory.  It is still arguable to switch the
install process automatically, maybe leaving this choice another
command is the better answer.

madlib.control_in is put in madpack directory, per some design
discussion with Florian.  Also this includes a small change in
madpack to show python stack trace on database object creation
error when -v option is specified, which is handy for the diagnosis
of the script problem.

MADLIB-316

### DIFF
--- a/src/ports/postgres/madpack/madlib.control_in
+++ b/src/ports/postgres/madpack/madlib.control_in
@@ -1,0 +1,3 @@
+default_version = '@MADLIB_VERSION_STRING@'
+comment = 'A scalable in-database analytics library'
+relocatable = false
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -57,17 +57,65 @@ dbver = None        # DB version
 con_args = {}       # DB connection arguments
 verbose = None      # Verbose flag
 
-## # # # # # # # # # # # # # # # # # # # # # # # # # # # #
-# Create a temp dir 
-# @param dir temp directory path
-## # # # # # # # # # # # # # # # # # # # # # # # # # # # #
-def __make_dir(dir):
-    if not os.path.isdir(dir):
-        try:
-            os.makedirs(dir)
-        except:
-            print "ERROR: can not create directory: %s. Check permissions." % dir
-            exit(1)    
+class ModuleItem:
+    def __init__(self, name):
+        self.name = name
+
+    # Find the Python module dir (platform specific or generic)
+    def pydir(self):
+        pathname = os.path.join(maddir, "ports", portid, dbver, "modules", self.name)
+        if os.path.isdir(pathname):
+            return os.path.dirname(pathname)
+        else:
+            return os.path.join(maddir, "modules")
+
+    # Find the SQL module dir (platform specific or generic)
+    # Return None if this platform is not supported.
+    def sqldir(self):
+        specific_path = os.path.join(maddir, "ports", portid, "modules", self.name)
+        generic_path = os.path.join(maddir, "modules", self.name)
+        if os.path.isdir(specific_path):
+            return os.path.dirname(specific_path)
+        elif os.path.isdir(generic_path):
+            return os.path.dirname(generic_path)
+
+    # Generator for each masked file under its sql directory
+    def sqldir_files(self, mask):
+        full_mask = os.path.join(self.sqldir(), self.name, mask)
+        files = glob.glob(full_mask)
+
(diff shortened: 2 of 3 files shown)


src/ports/postgres/madpack/madlib.control_in
```diff
@@ -0,0 +1,3 @@
+default_version = '@MADLIB_VERSION_STRING@'
+comment = 'A scalable in-database analytics library'
+relocatable = false
```
fschopp, line 3: We should be relocatable. All my code is (I think). What is not relocatable?

haradh1, line 3: Relocatable means this extension can be moved from one schema to another (ALTER EXTENSION madlib SET SCHEMA another).  As far as I can tell, the hard-coded (with substitution of m4) schema is all over the code.

fschopp, line 3: Yes, but only as a schema qualifier. I think PG would update that properly?

fschopp, line 3: In fact, madpack renames the old madlib schema if it exists. As far as I remember, the old MADlib still works afterwards.

haradh1, line 3: Really?  Hmm, I'll test it later...

jhellerstein:
Nice!  I'm excited to see EXTENSION support.  When it's mature we should make sure it's on pgxn.org

J

src/ports/postgres/madpack/madlib.control_in
haradh1, line 3: This is not what I expected, but svec operator is not schema-qualified.

<pre>
haradh1=# alter extension madlib set schema madlib3;
ERROR:  extension "madlib" does not support SET SCHEMA
DETAIL:  operator &lt;(madlib.svec,madlib.svec) is not in the extension's schema "madlib"
</pre>

haradh1, line 3: Hmm, it's qualified.  In turn, it is not relocatable because it's schema-qualified.

haradh1, line 3: I think commutator and negator should also be schema qualified.  There's something weird.

``` sql
haradh1=# select oprname, oprnamespace from pg_operator where oprleft = 'madlib.svec'::regtype and oprright = 'madlib.svec'::regtype;
 oprname | oprnamespace 
---------+--------------
 ||      |        92641
 -       |        92641
 +       |        92641
 /       |        92641
 %*%     |        92641
 *       |        92641
 ^       |        92641
 =       |        92641
 <       |        92641
 <=      |        92641
 >=      |         2200
 >       |         2200
 <>      |        92641
 =       |         2200
 ==      |        92641
 <>      |         2200
 >=      |        92641
 >       |        92641
 <       |         2200
 <=      |         2200
(20 rows)
```

haradh1, line 3: Florian,

RENAME schema / ALTER EXTENSION SET SCHEMA changes object-schema relationship correctly, but doesn't change procedure source.  For example, logregr is implemented in plpgsql with this source.

``` sql
            SELECT
                MADLIB_SCHEMA.$sql$ || fnName || $sql$(_madlib_state) AS result
                FROM _madlib_iterative_alg
                WHERE _madlib_iteration = $sql$ || theIteration || $sql$
            ) subq
```

And invoking this after moving schema fails because this part points to the old schema name.  The right answer to support relocatable is to remove schema qualifier from everywhere, which seems difficult at the moment.

fschopp, line 3: Hmm, that is an oversight indeed. Can a PL/pgSQL function find out its own schema name? In PL/Python we do that.

haradh1, line 3: Where in PL/Python for example?

cwelton:
There is an outstanding question of what the right design here should be from an installation perspective.  How should it interact with MADpack, etc?

Current thought is not to pull this as it stands but to continue working to get it to a more stable state.

haradh1:
The design was changed and sent another pull request 141.  Closing this.

PR #129, opened 2012-04-13
Rf review

renyi533:
This is a preliminary review of the random forest code. I do not intend to push it in immediately. Any comments are welcome.

Since C4.5 share a lot of common code with random forest, we also changed c4.5 code. Please also review that part.

In this pull request, it includes all other pending pull request for decision tree.


aojwang:
The implementation of RF supports:
- Building random forests
- Multiple split critera, including:
  . Information Gain
  . Gini Coefficient
  . Gain Ratio
- Random forest Classification/Scoring
- Random forest Display
- Continuous and Discrete features
- Equal frequency discretization for continuous features
- Missing value handling
- Sampling with replacement

aojwang:
For C4.5 decision tree, the change is below.
1) C45_Train API
- remove the parameter "max_num_iter". As we grow the tree level by level, this parameter doesn't make sence.

2) Bugs 
- Fix for madlib-381/382: out of memory. PR: https://github.com/madlib/madlib/pull/119
- Fix for madlib-317. PR: https://github.com/madlib/madlib/pull/120

3) Performance Improvement
- Features used by ancestor nodes will be excluded when finding the best split for the node.
- Calculate the ACS set feature by feature.
- Parallize the operations after finding the best split.
- Format function was changed to C code from PLPGSQL

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+ *
+ * Parameters:
+ *      s     The pointer points to the string
+ *      len	  The length of the substring to be converted to the text
+ * Return:
+ *      A string represented by the text structure
+ */
+static
+text*
+dt_string_to_text
```
haradh1, line 1891: Use cstring_to_text

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+		{
+			dt_check_error_value
+				(
+					(str - 1) && ('\\' == *(str - 1)),
+					"The char immediately before a %s must be a \\",
+					"%"
+				);
+
+			/*
+			 * The char immediately
```
haradh1, line 2019: Why are we continuing to use plpgsql with such reinvention of wheel??  If you have time to invest in this, simply convert driver UDFs to plpython.

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+	float8 num_errors,
+	float8 conf_level,
+	float8 coeff
+	);
+
+
+/*
+ * The functionality of this function is the same with palloc0, except that
+ * we explicitly set the size to (1 * unit_size) if it's 0, and check
+ * if the memory allocation is failed or not.
```
haradh1, line 116: palloc/palloc0 never return NULL.

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+ * Parameters:
+ *      num_units	The number of memory units should be allocated
+ *      unit_size   	The number of bytes per each memory unit
+ *
+ * Return:
+ *      The pointer points to the allocated memory which size is
+ *      (num_units * unit_size)
+ */
+void*
+dt_palloc0
```
haradh1, line 127: We don't need this.  Such interpretation should be done in the specific application context.

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+	void *ptr = palloc((num_units) * (unit_size));
+	dt_check_error(ptr, "memory allocation failure");
+
+	return ptr;
+}
+
+
+/*
+ * Calculates the total errors used by Error Based Pruning (EBP).
+ * This will be wrapped as a plc function.
```
haradh1, line 171: What is `plc` function?

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+        array_length 	 = max_num_of_classes + 1;
+        rebuild_array 	 = true;
+
+    }
+    else
+    {
+        if (fcinfo->context && IsA(fcinfo->context, AggState))
+            pg_class_count = PG_GETARG_ARRAYTYPE_P(0);
+        else
+            pg_class_count = PG_GETARG_ARRAYTYPE_P_COPY(0);
```
haradh1, line 408: Is it expected to call this function as a non-aggregate function?

haradh1, line 408: And if so, the right check is AggCheckCallContext.

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+ * Parameters:
+ *      sample_size     Number of records to be sampled
+ *      low             Low limit of sampled values
+ *      high            High limit of sampled values
+ *      seed            Seed for random number
+ * Return:
+ *      A set of integer values sampled randomly between [low, high].
+ */
+Datum
+dt_sample_within_range
```
haradh1, line 1617: It seems this could be done by generate_series() and random().

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+{
+    ArrayType*	scv_state_array	= PG_GETARG_ARRAYTYPE_P(0);
+	dt_check_error
+		(
+			scv_state_array,
+			"invalid aggregation state array"
+		);
+
+    int	 array_dim		= ARR_NDIM(scv_state_array);
+    dt_check_error_value
```
haradh1, line 1431: I strongly recommend to separate error check code and actual logic.  This (and other functions around here) is just unreadable.

methods/CART/src/pg_gp/dt.c
```diff
@@ -0,0 +1,2219 @@
...
+ * Cast a text to c-style string.
+ *
+ * Parameters:
+ *      t     The pointer points to the text
+ * Return:
+ *      A c-style string
+ */
+static
+char*
+dt_text_to_string
```
haradh1, line 1858: Use text_to_cstring

methods/CART/src/pg_gp/c45.sql_in
```diff
@@ -0,0 +1,1391 @@
...
+                ),
+                'the specified tree table'      || 
+                coalesce('<'                    || 
+                tree_table                      || 
+                '> does not exists', ' is NULL')
+            );   
+
+    FOR str IN SELECT * FROM 
+m4_changequote(`>>>', `<<<')
+m4_ifdef(>>>__HAS_ORDERED_AGGREGATES__<<<, >>>
```
haradh1, line 1129: I don't see why we need two of with_ordered and no_ordered versions.  You can define only one and encapsulate the ordered/non-ordered switching logic inside the function.

haradh1:
Overall this is unnecessarily too messy.  First, we really need to use pl/python instead of pl/pgsql.  The code around __format is too verbose and unmaintainable.  Second, we should progres the big feature step by step but this contains everything at once (utility, code relocation, new feature, refactoring, etc.), which is not able to be reviewed / given feedback. Thanks.

methods/CART/src/pg_gp/c45.sql_in
renyi533, line 1129: Postgres 8.4 does not support ordered aggregate. Therefore, we need a different solution for it. Since the solution without ordered aggregate is much slower, we will only use it for PG 8.4.

Anyway, we need a switch point. I can define another function named MADLIB_SCHEMA.__treemodel_display and embed the switching logic inside this new function. Does this meet your suggestions?

methods/CART/src/pg_gp/dt.c
renyi533, line 1617: This function can surely be implemented with random() and some other existing PLPGSQL facilities. Actually, for demo purpose, I originally wrote a version leveraging random()/generate_series() with PLPGSQL. But that function's performance is much worse than this C implementation. (Ten times slower)

methods/CART/src/pg_gp/dt.c
renyi533, line 408: Actually, in our code, we never call this function as a non-aggregate function. We add this check to ensure the correctness even if it is called as a non-aggregate function.

renyi533, line 408: AggCheckCallContext" is available for PG 9.0/9.1. And I just searched madlib's code, it seems we have a copy of this function in the file of GPCompatibility.cpp.

For PG 8.4, is it OK for us to use it directly?

I think our original implementation is correct. Its logic is the same as "AggCheckCallContext".

methods/CART/src/pg_gp/dt.c
renyi533, line 1891: Though this function is static, we think naming collision is not good. Thus, we added the prefix of "dt_" to avoid potential collision. What's the value of changing it to "cstring_to_text". Don't you think naming collision is bad?

methods/CART/src/pg_gp/dt.c
renyi533, line 408: If we can use it safely for all versions of databases, I am very happy to use it instead. Please confirm for us. Many thanks.

methods/CART/src/pg_gp/dt.c
renyi533, line 1858: It is the same as above. How do you think about the potential naming collision problem?

aojwang:
1) For performance consideration, we use c to write the format function. The code looks long, but the code path is very clear. What do you mean unmaintainable here? thanks.

2) The PR really contains a lot of changes, maybe we can schedule a code walk through, so that we are in the same page. What do you think, thanks.

methods/CART/src/pg_gp/dt.c
aojwang, line 1858: Actually, i notice the existing function text_to_cstring. However, there are some error when linkage. I try use text_to_cstring again, it works well. Maybe i have something wrong before. I will change to the existing functions: text_to_cstring and cstring_to_text.  Thanks.

methods/CART/src/pg_gp/dt.c
aojwang, line 1891: Actually, i notice the existing function text_to_cstring. However, there are some error when linkage. I try use text_to_cstring again, it works well. Maybe i have something wrong before. I will change to the existing functions: text_to_cstring and cstring_to_text. Thanks.

methods/CART/src/pg_gp/dt.c
aojwang, line 2019: You're right. Plpython has a lot of string utility, and easy to use. However, for the historical reason, we use the PL/PGSQL. We have some discussion on this issue in the meeting before. Thanks.

methods/CART/src/pg_gp/dt.c
aojwang, line 116: Ok, will use palloc/palloc0 directly. Thanks

methods/CART/src/pg_gp/dt.c
aojwang, line 171: What we mean here is that we will create a PL/PGSQL function for this c function, as you can see from the following code.
{code}
CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__ebp_calc_errors
    (
    total                       FLOAT8,
    prob                      FLOAT8,
    confidence_level    FLOAT8
    ) RETURNS FLOAT8 
AS 'MODULE_PATHNAME', 'dt_ebp_calc_errors'
LANGUAGE C STRICT IMMUTABLE;
{code}

methods/CART/src/pg_gp/dt.c
haradh1, line 2019: I know the discussion that it's because of the historical reason.  I see what you mean, you didn't have time to convert it in plpython because there is too much existing plpgsql.  That's why I'm saying "if you have time, use the time to make it plpython"  I think inventing the new stuff in c and plpgsql is much more than transforming it to plpython.  To write a new stuff like __format is not only to write it once, but we'll need it read, maintained, and looked back if there's some bug.

methods/CART/src/pg_gp/dt.c
haradh1, line 171: The sentence is confusing because I don't get what is plc.  Do you mean plpgsql function, or C function?

methods/CART/src/pg_gp/dt.c
haradh1, line 408: Yes, we have the compatibility layer and use it.

haradh1:
> 1) For performance consideration, we use c to write the format function. The code looks long, but the code path is very clear. What do you mean unmaintainable here? thanks.

It may be clear to you, the writers, but is not to us, the readers.  What I'd like to see is to separate algorithm and miscellaneous works so that the intention of the code can be seen clearly.  Currently it is difficult to tell even where the algorithm is described and what is the other ones.

methods/CART/src/pg_gp/dt.c
renyi533, line 1891: I don't know "text_to_cstring" is standard library function. Then, we should use it definitely.

cwelton:
My understanding is that this should be closed in preference for PR-134.  Please confirm.

aojwang:
Yes. This PR should be discarded. Please review the RF implementation in [PR-134](https://github.com/madlib/madlib/pull/134)

renyi533:
Yes. This PR is obsolete. Please review PR-134.

cwelton:
Closing obsolete pull request.

PR #131, opened 2012-04-16
Added boost probability functions

KevinYL:
As requested in [MADLIB-412](http://jira.madlib.net/browse/MADLIB-412), probability functions from boost now have a MADlib wrapper.


src/ports/postgres/modules/prob/test/prob.sql_in
```diff
@@ -53,3 +53,50 @@ SELECT assert(
...
     'Student-t CDF: Non-positive degree of freedom (nu) does not raise error.'
 );
+
+
+
+CREATE OR REPLACE FUNCTION _test_normal_pdf(t float8, mu float8, sigma float8) RETURNS float8 AS $$
+BEGIN
+    return exp(-1*(t - mu)^2/(2*(sigma^2)))/sqrt(2*pi()*sigma^2);
+END;
+$$LANGUAGE 'plpgsql' STRICT IMMUTABLE;
```
fschopp: This should be a plain SQL function, not PL/pgSQL.

src/ports/postgres/modules/prob/test/prob.sql_in
```diff
@@ -53,3 +53,50 @@ SELECT assert(
...
+    normal_cdf(0, NULL, 1) IS NULL AND
+    normal_cdf(0, 0, NULL) IS NULL,
+    'Normal CDF: Wrong handling of NULLs.' 
+);
+
+select assert(
+    isnan(normal_cdf('NaN', 0, 1)) AND
+    isnan(normal_cdf(0, 'NaN', 1)) AND
+    isnan(normal_cdf(0, 0, 'NaN')),
+    'Normal CDF: Wrong handling of NULLs.'
```
fschopp: This is a copied typo. s/NULLs/NaN/

src/modules/prob/students_t.cpp
```diff
@@ -0,0 +1,160 @@
...
+	STUDENTS_T_DOMAIN_CHECK(df);
+	
+	
+	if ( x == -std::numeric_limits<double>::infinity() ) {
+		return 0.0;
+	}
+	if ( x == std::numeric_limits<double>::infinity() ) {
+		return 1.0;
+	}
+	return boost::math::cdf(boost::math::students_t_distribution<>(df), x);
```
cwelton: This replaces Florian's Student T implementation with the Boost implementation.  In previous discussions with him my impression was that he believed that his implementation should outperform the boost implementation.  

Florian, could you comment here?

KevinYL:
The relating jira is available via: http://jira.madlib.net/browse/MADLIB-448

src/modules/prob/students_t.cpp
```diff
@@ -0,0 +1,160 @@
...
+namespace modules {
+
+namespace prob {
+// Workaround for Doxygen: A header file that does not declare namespaces is to
+// be ignored if and only if it is processed stand-alone
+#undef _DOXYGEN_IGNORE_HEADER_FILE
+#include "students_t.hpp" 
+
+
+#define STUDENTS_T_DOMAIN_CHECK(df)\
```
fschopp: This should not be a macro but a proper function.

src/modules/prob/students_t.cpp
```diff
@@ -0,0 +1,160 @@
...
+			return std::numeric_limits<double>::quiet_NaN();\
+		}\
+		if ( !(df > 0) ) {\
+			throw std::domain_error("Students_t distribution is undefined when df doesn't conform to (df > 0).");\
+		}\
+	} while(0)
+
+
+inline double 
+STUDENTS_T_CDF(double x, double df)
```
fschopp: Please don't capitalize function names. Capitalization should be reserved for macros.

src/modules/prob/students_t.cpp
fschopp: In my experiments, my implementation of the Student's-t-CDF was more accurate (see the comments in my implementation, i.e., in student.cpp). However, it only works for natural (as in natural number) degrees of freedom, not for rational ones. That's why my implementation resorted to boost as well in non-natural cases. However, a natural degree of freedom is far more common.

At this point, I do not see a reason to replace my existing Student's t implementation.

src/modules/prob/students_t.cpp
```diff
@@ -0,0 +1,160 @@
...
+}
+
+double
+students_t_PDF(double x, double df) {
+	double res = 0;
+
+	try {
+		res = STUDENTS_T_PDF(x, df);
+	}
+	catch (...) {
```
fschopp: Please don't use catch alls. There are only very few places where their use is legitimate.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -37,14 +67,49 @@ IMMUTABLE STRICT;
...
  * @brief Fisher F cumulative distribution function.
  *
- * @param x
- * @param d_1 Degree of freedom in numerator
- * @param d_2 Degree of freedom in denominator
- * @return \f$ \Pr[X \leq x] \f$ where \f$ X \f$ is a Fisher F distributed random
+ * @param x Fisher F distributed random variable
+ * @param d_1 Degree of freedom in numerator \f$ d_1 > 0 \f$
+ * @param d_2 Degree of freedom in denominator \f$ d_2 > 0 \f$
+ * @return \f$ \Pr[X \leq x] \f$ where \f$ X \f$ is a fisher F distributed random
```
fschopp: fisher is a name, so it should be written with a capital F.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -37,14 +67,49 @@ IMMUTABLE STRICT;
...
+LANGUAGE C
+IMMUTABLE STRICT;
+
+/**
+ * @brief Fisher F probability density function.
+ *
+ * @param x Fisher F distributed random variable
+ * @param d_1 Degree of freedom in numerator \f$ d_1 > 0 \f$
+ * @param d_2 Degree of freedom in denominator \f$ d_2 > 0 \f$
+ * @return the relative frequency of \f$ x \f$ where \f$ x \f$ is a fisher F distributed random
```
fschopp: fisher -> Fisher

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -55,14 +120,14 @@ IMMUTABLE STRICT;
 /**
  * @brief Kolmogorov cumulative distribution function.
  *
- * @param x
- * @return \f$ \Pr[X \leq x] \f$ where \f$ X \f$ is a Kolmogorov distributed random
+ * @param x Kolmogorov distributed random variable
```
fschopp: No, small x is not a random variable. It is just some value x. Please remove that description again.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -55,14 +120,14 @@ IMMUTABLE STRICT;
...
+ * @return \f$ \Pr[X \leq x] \f$ where \f$ X \f$ is a kolmogorov distributed random
```
fschopp: kolmogorov -> Kolmogorov

Please just remove the change consisting of these two lines.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -55,14 +120,14 @@ IMMUTABLE STRICT;
...
  *     variable
  *
  * @sa Kolmogorov-Smirnov test: ks_test()
  */
 CREATE FUNCTION MADLIB_SCHEMA.kolmogorov_cdf(
-    x DOUBLE PRECISION
+    "x" DOUBLE PRECISION
```
fschopp: What's the benefit of this change? I think we should use unquoted identifiers.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -71,31 +136,96 @@ IMMUTABLE STRICT;
...
+IMMUTABLE STRICT;
+
+/**
+ * @brief Normal probability density function
+ *
+ * @param x normal distributed random variable
+ * @param mu Mean \f$ \mu \f$
+ * @param sigma Standard deviation \f$ \sigma > 0 \f$
+ * @return the relative frequency of \f$ x \f$ where \f$ x \f$ is a normally distributed random
+ *     variable with mean \f$ \mu \f$ and variance \f$ \sigma^2 \f$
```
fschopp, line 176: This does not make sense. See above. Also, I don't think we need to explain what a density function is by using a term like "relative frequency". Right here, that is arguably more confusing than helpful. Therefore, it should be enough to write:

```
@return \f$ f(x) \f$ where \f$ f \f$ is the probability density function of a normally
    distributed random variable with mean \f$ \mu \f$ and variance \f$ \sigma^2 \f$
```

src/modules/prob/chi_squared.hpp
```diff
@@ -1,34 +1,31 @@
...
+DECLARE_UDF(chi_squared_quantile)
 
 #if !defined(DECLARE_LIBRARY_EXPORTS)
 
-double chiSquaredCDF(double t, double nu);
-double fisherF_CDF(double t, double df1, double df2);
-double normalCDF(double t, double mu = 0, double sigma = 1);
-
-#endif // !defined(DECLARE_LIBRARY_EXPORTS)
+double chi_squared_CDF(double x, double df);
```
fschopp, line 26: What is the benefit of renaming chiSquaredCDF to chi_squared_CDF?

src/ports/postgres/modules/prob/prob.sql_in
cwelton: No effect at all from this change.

src/modules/prob/bernoulli.cpp
```diff
@@ -0,0 +1,166 @@
...
+		res = std::numeric_limits<double>::quiet_NaN();
+	}
+
+	return res;
+}
+
+
+
+inline double 
+BERNOULLI_PDF(double x, double succ_prob)
```
fschopp: We should keep our terminology clear. This is a probability mass function, not a probability density function.

src/modules/prob/bernoulli.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+	BERNOULLI_DOMAIN_CHECK(succ_prob);
+	if ( (int)x != x && !std::isinf(x) ) {
```
fschopp: We should not accept double arguments but then bail if a value is not integer. The argument should simply be changed to integer.

src/modules/prob/bernoulli.cpp
```diff
@@ -0,0 +1,166 @@
...
+double
+bernoulli_PDF(double x, double succ_prob) {
+	double res = 0;
+
+	try {
+		res = BERNOULLI_PDF(x, succ_prob);
+	}
+	catch (...) {
+		res = std::numeric_limits<double>::quiet_NaN();
+	}
```
fschopp, line 112: I commented already elsewhere: Don't use a catch all. Don't capitalize BernoulliPDF.

src/modules/prob/bernoulli.cpp
```diff
@@ -0,0 +1,166 @@
...
+	
+	if ( x < 0 ) {
+		return 0.0;
+	}
+	if ( x > 1 ) {
+		return 1.0;
+	}
+	x = floor(x);
+	return boost::math::cdf(boost::math::bernoulli_distribution<>(succ_prob), x); 
+}
```
fschopp, line 49: Please follow the style put forward by my existing C++ code. E.g., indent with 4 spaces, place functions opening { on previous line, etc. Thanks.

src/modules/prob/bernoulli.cpp
```diff
@@ -0,0 +1,166 @@
...
+	BERNOULLI_DOMAIN_CHECK(succ_prob);
+	
+	
+	if ( x < 0 ) {
+		return 0.0;
+	}
+	if ( x > 1 ) {
+		return 1.0;
+	}
+	x = floor(x);
```
fschopp, line 47: Why would that be necessary?

src/modules/prob/beta.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+
+
+inline double 
+BETA_QUANTILE(double x, double alpha, double beta)
+{
+	BETA_DOMAIN_CHECK(alpha, beta);
+	
+	if ( x < 0 || x > 1 ) {
+		throw std::domain_error("Beta distribution is undefined for CDF out of range [0, 1].");
```
fschopp: Unclear error message.

src/modules/prob/bernoulli.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+	BERNOULLI_DOMAIN_CHECK(succ_prob);
+	if ( (int)x != x && !std::isinf(x) ) {
+		throw std::domain_error("Bernoulli distribution is a discrete distribution, random variable can only be interger.");
+	}
+	
+	if ( x < 0 ) {
+		return 0.0;
+	}
+	if ( x > 1 ) {
```
fschopp: I'd suggest an `else if` here for clarity.

src/modules/prob/chi_squared.hpp
```diff
@@ -1,34 +1,31 @@
...
- *//* ----------------------------------------------------------------------- */
+ *//* ----------------------------------------------------------------------- */ 
 
 // Workaround for Doxygen: A header file that does not declare namespaces is to
 // be ignored if and only if it is processed stand-alone
 #if !defined(_DOXYGEN_IGNORE_HEADER_FILE)
-
 /**
- * @brief Chi-squared cumulative distribution function
+ * @brief chi_squared cumulative distribution function
```
fschopp: Questionable change

fschopp:
General remarks:
- All comments I made so far apply to most if not all probability functions.
- Support for discrete distributions needs more work (see my comments).
- Some error messages should be stated more clearly
- Doxygen comments need some work. E.g.: Don't write SQL function names in math mode. Ensure that the terminology is correct. None of these functions accept a "random variable" as parameter.

I'll need to have a second look at the corner-case checks.

src/modules/prob/bernoulli.cpp
cwelton: Agreed.  Additionally "integer" is misspelled.

src/modules/prob/bernoulli.cpp
KevinYL, line 47: Boost will throw a exception if x is not an integer for discrete distributions.

src/modules/prob/chi_squared.hpp
KevinYL, line 26: To be uniform. I don't think there is meaning to create ourself function names. I make all the function names conforms to boost.

src/modules/prob/bernoulli.cpp
KevinYL, line 112: Then catch what? All the domain errors was checked to return NaN in your code? What else exceptions will throw? Only overflow. But overflow only happens on border conditions in which we must handle ourselves. If we can't handle these border conditions I think it will be a bug.

src/modules/prob/students_t.cpp
KevinYL: Sorry, I will revert cdf to yours version. In the past before I don't know why you implemented your own version. Fixed

src/ports/postgres/modules/prob/prob.sql_in
KevinYL: Just to be uniform. 
This is a piece of code copy from main. Why d_1 and d_2 are quoted, and x are not quoted? I make them uniform. 
CREATE FUNCTION MADLIB_SCHEMA.fisher_f_cdf(
    x DOUBLE PRECISION,
    "d_1" DOUBLE PRECISION,
    "d_2" DOUBLE PRECISION
) RETURNS DOUBLE PRECISION
AS 'MODULE_PATHNAME'
LANGUAGE C
IMMUTABLE STRICT;

src/modules/prob/bernoulli.cpp
KevinYL: fixed

src/modules/prob/bernoulli.cpp
KevinYL: fixed

src/modules/prob/bernoulli.cpp
KevinYL: fixed

src/modules/prob/chi_squared.hpp
KevinYL: fixed

src/modules/prob/chi_squared.hpp
KevinYL, line 26: To be honest, there is no any good reason. I do for this because I generated most of the code by script from configuration file with the input list below:
chi_squared             ;   chi_squared_distribution             ;    c#0                 ; df>@<df > 0 

Renaming to chi_squared_CDF makes it more simple to process by script. 

So if this is not a serious problem, I want to keep this name convention. 

Almost 95 percent of the codes are the same for different distributions.

src/modules/prob/students_t.cpp
KevinYL: This piece of code appears in very source file and is used by all the three functions.  If any of the args is NaN, I want the functions return NaN directly, which can't be done by function. 

So if this is not a serious problem. Could I just keep it?

src/modules/prob/students_t.cpp
KevinYL: fixed. Using _ as prefix to indicate internal function.

src/modules/prob/students_t.cpp
KevinYL: These functions are all calculation functions. I think there are only two exception. One is domain_error, according to the original code, this kind of error should be caught and return NaN. The another is overflow exception, but for distribution functions, there is no reason to overflow, we should process these conditions.

src/ports/postgres/modules/prob/prob.sql_in
KevinYL: fixed

src/ports/postgres/modules/prob/prob.sql_in
KevinYL: fixed

src/ports/postgres/modules/prob/prob.sql_in
KevinYL: fixed, remove all quotes.

src/ports/postgres/modules/prob/prob.sql_in
KevinYL: fixed

src/ports/postgres/modules/prob/prob.sql_in
KevinYL: fixed

src/ports/postgres/modules/prob/prob.sql_in
KevinYL, line 176: Hi, Florian, I may need more your help about the comments on all the distribution functions. I need much more study to express as clearly as you do. How about you edit these comments directly after the source code is ok?

src/modules/prob/bernoulli.cpp
KevinYL, line 49: For opening {, fixed. But for indenting, why do not we used tab? GPDB requires indenting with tab. Madlib requires indenting with 4 spaces. One group, two standard, is there any good reason?

src/modules/prob/beta.cpp
KevinYL: fixed

src/modules/prob/students_t.cpp
fschopp: Please don't start names with an underscore. See, e.g., [here](http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier) for reasons why. I still don't see the need to change the existing name, which was `studentT_CDF`. You said you used a script for generating code, but the script could be changed, couldn't it?

src/modules/prob/chi_squared.hpp
fschopp, line 26: Would it make sense to add the script to the repository? It could be invoke by the CMake build script.

KevinYL, line 26: OK. You are the boss.  I'll change the rename back. Even I don't know the differences. But if there is no good reason for other functions' name, I refuse to change. 

About the script. I can just generate 95% percent of the codes. About half of the files need manual adjust. So It's meaningless to make it into the repository.

src/modules/prob/bernoulli.cpp
fschopp, line 112: Honestly, I would not catch anything here. We check all conditions upfront that we anticipate to cause issues. Conversely, any exception raised during the boost function call is unexpected. I think, we want to see those unexpected errors, and we also want users to report them.

src/modules/prob/chi_squared.hpp
fschopp, line 26: Or in other words: Is your script ready for inclusion? If not, how much work is it to polish it?

KevinYL, line 26: Could you get online? 

```
if ( x < 0 ) {
    return 0.0;
}
else if ( std::isinf(x) ) {
    return 0.0;
}
else if ( 0 == x ) {
    /// FIXME
    return df1 > 2 ? 0 : (df1 == 2 ? 1 : std::numeric_limits<double>::infinity());
}
```

This piece of code was picked from fisher_f_pdf.  The FIXME items need manual adjust. 
Because not every function can call boost directly. And conditions like this is not fit to 
be generated by script. About half of the distribution functions need do manual work like this.

fschopp:
I just took a closer look at the Boost [Math Toolkit](http://www.boost.org/doc/libs/1_49_0/libs/math/doc/sf_and_dist/html/index.html). The way errors are handled can be fine-tuned by choosing appropriate [Policies](http://www.boost.org/doc/libs/1_49_0/libs/math/doc/sf_and_dist/html/math_toolkit/policy.html). So there exists a much more elegant way of argument validation and we should:
- Use policies instead of wrapping boost function calls with ugly checks
- Use boost error messages whenever possible
- Omit any domain validation whenever boost already does so (e.g., the boost normal cdf actually **does** return 0 for x = -infinity and 1 for x = +infinity). I don't know why we are doing a redundant handling of these cases. Maybe a copy&paste leftover.
- Given the flexibility of policies, I see no need to make our wrapped probability functions available to C++ code. C++ code should just call boost directly.
- The students-t code should be modified to conform to the boost API. We should also have some numbers that back up my claim that my implementation provides better accuracy than boost's implementation. The right place for such numbers is either the design document or doxygen comments.

Given the above items, we are not yet ready for merge. I am happy to receive feedback on the above. If not, I'll go ahead and make a concrete proposal. I'll also update our JIRA as needed.

fschopp:
I have added Kevin's SQL changes and unit tests to my fork, and will update pull request 156 as soon as all tests pass. I will then close this pull request.

fschopp:
Pull request 156 has been updated with Kevin's changes. Unfortunately, there are a lot of bugs in the boost code when it comes to corner cases, and writing the workarounds is tedious. Therefore, pull request 156 does not yet pass all of the unit tests.

In any case, pull request 131 is now obsolete.

cwelton:
Agreed, closing this PR.  Thanks for the work in this area Kevin.

PR #132, opened 2012-04-23
Fixes for madlib-382

aojwang:
In current version, there is no tree ID in the result tree table. Therefore, remove the related code. Now, it can run successfully in Postgres8.4 and series version. Thanks.


haradh1:
Verified with this version PG8.4 runs without failure. Closing.

PR #134, opened 2012-04-28
Random Forest Implementation For MADlib (MADLIB-419)

aojwang:
Random forest (RF) is an ensemble classifier that consists of multiple decision trees (DTs) and outputs the class that is voted by the majority of the individual trees. A prominent advantage of RF is its unexcelled accurary. This is the main reason why we support RF in MADlib. MADlib RF is based on Leo Breiman and Adele Cutler’s RF algorithm and runs directly within databases. 

Our RF implementation is based on MADlib DT without error based pruning and reduced error pruning.  But we add a parameter, named max_split_point, to control how many feature values will be treated as split values. Using this approach, in most cases, the total time of training an RF can be decreased a lot.

Our RF functionality can be described as following.
- Training              (rf_train)
- Displaying          (rf_display)
- Classification      (rf_classify)
- Scoring               (rf_score)
- Cleaning             (rf_clean)


methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -4030,11 +4044,9 @@ $$ LANGUAGE PLPGSQL;
  *        since gpdb's performance is not good for small data sets.
  *
  * @param classification_table_name  The full name of the table containing the 
- *                                   classification set
```
aojwang, line 4033: Remove the unnecessary comments.
We don't want to open a JIRA and pull request for this issue, since we only remove the comments. Therefore, we combine them together, and the changes will be applied to the latest MADlib version.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+    ret                         MADLIB_SCHEMA.rf_train_result;
+BEGIN   
+    ret = MADLIB_SCHEMA.rf_train
+            (
+                split_criterion,
+                training_table_name,
+                result_rf_table_name,
+                10, 
+                null,
+                0.632,
```
aojwang, line 880: This is a well-known bootstrap method, called 0.632 bootstrap. According to the book "Data mining concepts and techniques, 3rd Edition", "suppose we are given a data set of D tuples. Each tuple has a probability 1/d of being selected, so the probability of not being chosen is 1 − 1/d. We have to select D times, so the probability that a tuple will not be chosen during this whole time is (1−1/d)^d. If D is large, the probability approaches e^−1. Thus, 36.8% of tuples will not be selected for training and thereby end up in the test set, and the remaining 63.2% will form the training set.”

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
+/* ----------------------------------------------------------------------- *//**
```
aojwang, line 1: As RF and C4.5 share the same tree growing engine, this file provides all the APIs of RF, and validates the input parameters' value.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -2225,16 +2226,28 @@ BEGIN
...
-    EXECUTE 'DROP TABLE IF EXISTS tr_association CASCADE';
-    curstmt = MADLIB_SCHEMA.__format
-                (
-                    'CREATE TEMP TABLE tr_association AS 
-                     SELECT id, 1 as tid, 1 as nid, 1 as weight 
-                     FROM %
-                     m4_ifdef(`__GREENPLUM__', `DISTRIBUTED BY (id)')',
-                     training_table_name
-                 );
-    EXECUTE curstmt;
```
renyi533, line 2237: Add one parameter to control whether do data sampling here.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -5005,3 +5013,135 @@ MADLIB_SCHEMA.__dt_get_node_split_fids(INT4, INT4, INT4, INT4[])
...
 LANGUAGE C VOLATILE;
+
+
+/*
+ * @brief The function samples a set of integer values between low and high.
+ *
+ * @param sample_size     The number of records to be sampled_rd.
+ * @param low             The low limit of sampled valuesn_fid.
+ * @param high            The high limit of sampled valuesnid.
+ * @param seed            The seed for random number generator.
```
renyi533: Add the subroutines for data sampling.

methods/decision_tree/src/pg_gp/rf.sql_in
renyi533, line 1: This file contains the API of random forest. It also calls the tree growing engine, which is at decision_tree.sql_in.

After this pull request is accepted, we will submit another pull request and separate decision_tree.sql_in into two files, namely c45.sql_in and dt.sql_in. c45.sql_in is for the API of c4.5. And dt.sql_in is for the tree growing engine.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+			tid = tids[i];
+			PERFORM MADLIB_SCHEMA.__assert
+					(
+						tid IS NOT NULL AND
+						tid > 0			AND
+						tid <= max_tid, 
+						'the ID of the tree in the array must be in range [1, ' || 
+						max_tid 												|| 
+						']'              
+					);
```
renyi533, line 971: Check the tree_id is within a valid range.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ * 2) We sort all values in ascending order and divides the range into K  
+ *    intervals so that each interval contains the same number of sorted values.
+ * 3) We only consider the values in the intervals' boundaries as potential
+ *    split points.
+ *
+ * We discretize continuous features on local regions during training rather 
+ * than discretizing on the whole dataset prior to training because local 
+ * discretization takes into account the context sensitivity.
+ *
+ * @param split_criterion_name      The split criterion used for tree construction.
```
aojwang: Should be split_criterion according to the code.

methods/decision_tree/src/pg_gp/sql/dec_trees.sql_in
```diff
@@ -554,15 +623,28 @@ declare
...
+    -- classify should work.
+    PERFORM MADLIB_SCHEMA.c45_classify(tree_name,score_set,classify_result,0);
+    -- Check the classification result is OK
+    IF ( MADLIB_SCHEMA.__table_exists(classify_result) ) THEN
+        EXECUTE 'DROP TABLE '||classify_result;
+    ELSE
+        result = 'FAIL';
+        RAISE EXCEPTION 'Install check failed.';
+    END IF;
+
```
renyi533, line 647: Add the test for c45_classify.

methods/decision_tree/src/pg_gp/sql/dec_trees.sql_in
```diff
@@ -535,6 +535,75 @@ INSERT INTO MADLIB_SCHEMA.crx_dt_test VALUES (486, 'b', 74.829999999999998, 19,
...
+        EXECUTE 'SELECT COUNT(*) FROM '||classify_result||' t1,'||classify_result2||
+            ' t2 WHERE t1.id=t2.id AND t1.class=t2.class;' INTO size_consistent;
+
+        EXECUTE 'DROP TABLE '||classify_result;
+        EXECUTE 'DROP TABLE '||classify_result2;
+        
+        -- The results from the two classification method should be consistent.
+        IF (size <> size_consistent) THEN
+            result = 'FAIL';
+            RAISE EXCEPTION 'Install check failed.';
```
renyi533, line 588: If the two classification results differ, there exists a bug.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ * 2) We sort all values in ascending order and divides the range into K  
+ *    intervals so that each interval contains the same number of sorted values.
+ * 3) We only consider the values in the intervals' boundaries as potential
+ *    split points.
+
+ * We discretize continuous features on local regions during training rather 
+ * than discretizing on the whole dataset prior to training because local 
+ * discretization takes into account the context sensitivity.
+ *
+ * @param split_criterion_name      The split criterion used for tree construction.
```
aojwang: Should be split_criterion according to the code.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+
+The <b>data to classify</b> is expected to be 
+of the same form as <b>training data</b>, except
+that it does not need a class column.
+
+@usage
+
+- Run the training algorithm on the source data:
+  <pre>SELECT * FROM \ref rf_train(
+    '<em>split_criterion_name</em>',
```
aojwang: Should be split_criterion.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ *
+ * @return A rf_classify_result object.
+ *
+ */
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.rf_classify
+    (
+    rf_table_name               TEXT, 
+    classification_table_name   TEXT, 
+    result_table_name           TEXT, 
+    is_serial_classification    BOOLEAN,
```
renyi533, line 1085: User can specify whether to classify with all trees together or one by one.

For very large datasets, serial classification is quicker as some operation in gpdb is not linear. Otherwise, serial classification is inferior.

For example, for the kddcup dataset, which is of 5 million records, if trained model is of more than 40 trees, serial classification is superior.

renyi533, line 1085: I tested on a 40 segments configuration running on gpdb[1-5].delta.sm.greenplum.com

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+       0.7,                                  -- sampling_percentage
+	   'temperature,humidity',               -- continuous feature names
+	   'outlook,temperature,humidity,windy', -- feature column names
+	   'id',                                 -- id column name
+	   'class',                              -- class column name
+	   'explicit',                           -- how to handle missing value
+	   10,                                   -- max tree depth
+	   0.001,                                -- min percent mode
+	   0.001,                                -- min percent split
+       'f'                                   -- optimize continuous feature
```
renyi533: This comment should be max_split_point. And its type is integer.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+    <em>feature3</em> SUPPORTED_DATA_TYPE,
+    ....................
+    <em>featureN</em> SUPPORTED_DATA_TYPE,
+    <em>class</em>    SUPPORTED_DATA_TYPE,
+    ...
+)</pre>
+
+The detailed list of SUPPORTED_DATA_TYPE is: 
+SMALLINT, INT, BIGINT, FLOAT8, REAL, NUMERIC,
+DECIMAL, INET, CIDR, MACADDR, BOOLEAN,
```
haradh1, line 76: decimal and numeric are identical.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ * discretization takes into account the context sensitivity.
+ *
+ * @param split_criterion           The split criterion used for tree construction. 
+ *                                  The valid values are infogain, gainratio, or
+ *                                  gini.
+ * @param training_table_name       The name of the table/view with the training data.
+ * @param result_rf_table_name      The name of the table where the resulting. 
+ *                                  trees will be stored.
+ * @param num_trees                 The number of trees to be trained. 
+ * @param features_per_node         The number of features to be considered
```
haradh1: There's no description about NULL for this parameter, while NULL is passed in the example.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ * continuous features. The discretization process is as follows:
+ * 1) We determine the minimum and maximum values of a continuous feature. 
+ * 2) We sort all values in ascending order and divides the range into K  
+ *    intervals so that each interval contains the same number of sorted values.
+ * 3) We only consider the values in the intervals' boundaries as potential
+ *    split points.
+ *
+ * We discretize continuous features on local regions during training rather 
+ * than discretizing on the whole dataset prior to training because local 
+ * discretization takes into account the context sensitivity.
```
haradh1, line 483: I'd like to see more comments here.  You only give the explanation about how to discrete the continuous values, but other parameters and behaviors should be explained as well.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ * continuous features. The discretization process is as follows:
+ * 1) We determine the minimum and maximum values of a continuous feature. 
+ * 2) We sort all values in ascending order and divides the range into K  
+ *    intervals so that each interval contains the same number of sorted values.
+ * 3) We only consider the values in the intervals' boundaries as potential
+ *    split points.
+
+ * We discretize continuous features on local regions during training rather 
+ * than discretizing on the whole dataset prior to training because local 
+ * discretization takes into account the context sensitivity.
```
haradh1: Move the comments which is applicable for general parts of RF to somewhere common.  Duplicated same comments here and there is not good.

methods/decision_tree/src/pg_gp/rf.sql_in
haradh1, line 1085: I'd love to make the decision automatically inside this function.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+ * @param how2handle_missing_value  The way to handle missing value. The valid 
+ *                                  values are 'explicit' and 'ignore'.
+ * @param max_tree_depth            The maximum tree depth. 
+ * @param node_prune_threshold      The minimum number of records required in a 
+ *                                  child node.
+ * @param node_split_threshold      The minimum number of records required in a 
+ *                                  node in order for a further split to be possible.
+ * @param max_split_point           The upper limit of sampled split points for 
+ *                                  continuous features.
+ * @param verbosity                 A flag indicating whether we should run in
```
haradh1: Make the comment for verbosity consistent.

methods/decision_tree/src/pg_gp/rf.sql_in
```diff
@@ -0,0 +1,1314 @@
...
+    (   
+    input_set_size        BIGINT,   
+    classification_time   INTERVAL
+    );
+
+
+/**
+ * @brief This API is defined for training a random forest.  
+ *
+ * For continuous features, each possible value could be used as a potential
```
haradh1: Which function is responsible for the discretization?  I'm looking for that in existing code, but it's fairly difficult for me.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1536,6 +1536,77 @@ enum DT_SCV_FINAL_ARRAY_INDEX
...
+		dt_check_error
+			(
+				low<=high && low>=0,
+				"The low margin must not be greater than the high margin. "
+           		"And negative numbers are not accepted"
+			);
+
+        /* total number of samples to be returned */
+        funcctx->max_calls = PG_GETARG_INT64(0);
+        srand(time(NULL)+seed);
```
haradh1: Again, I do ask why you need srand(), as you want just random numbers.  Calling random() without srand() gives you just random numbers.  srand(volative_value) also gives you random numbers.

methods/decision_tree/src/pg_gp/rf.sql_in
aojwang, line 76: Done. Thanks.

methods/decision_tree/src/pg_gp/rf.sql_in
aojwang: Done. Thanks.

methods/decision_tree/src/pg_gp/rf.sql_in
aojwang, line 483: Done. Thanks.

methods/decision_tree/src/pg_gp/rf.sql_in
aojwang: Removed and add new comments.

methods/decision_tree/src/pg_gp/rf.sql_in
aojwang: Done. Thanks.

methods/decision_tree/src/pg_gp/rf.sql_in
aojwang: the function `__create_training_instance`  in `decision_tree.sql_in`will do the job. After we get the ACC set, we will do the discretization for continuous features, and finally get the ACS set used to find the best splits. About the ACC and ACS set, please refer to the following description.

```
Attribute value and Class label Counts (ACC) set, which includes the following counts information for both continuous and discrete attribute. 
• Counts of a combination of attribute value and class label occurs 
• Counts of an attribute value occurs 
• Counts of a class label occurs 
• The number of records in the training table 
• Is an attribute continuous or discrete

```

```
Attribute value and Class label Statistics (ACS) set, calculated based on the ACC set, which includes the following statistics information for both continuous and discrete attribute. 
• For a <aij, ck> pair, the counts of  <aix, ck> pairs where aix is less/greater than aij, if ai is a continuous attribute 
• For attribute value aij, the counts of aix where aix is less/greater than aij, if ai is a continuous attribute 
• Counts of a combination of aij and ck occur, if ai is a discrete attribute 
• Counts of a class label occurs 
• Counts of an attribute value occurs, if the attribute is discrete 
• The number of records in the training table 
• Is an attribute continuous or discrete

NOTE: ai means the ith attribute, aij means the jth value of the ith attribute, and ck is the kth class label
```

methods/decision_tree/src/pg_gp/rf.sql_in
renyi533, line 1085: Yes. If we can determine automatically inside the function, it is perfect.

But for different hardware configurations and different data sets, the results differ greatly based on my experiments. It is hard to work out a formula for that.

methods/decision_tree/src/pg_gp/decision_tree.c
renyi533: If you don't call srand, the random number sequence is the same for each call. right?

I want different random number sequence for each call. Please refer to http://cplusplus.com/reference/clibrary/cstdlib/srand/.

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -5005,3 +5012,135 @@ MADLIB_SCHEMA.__dt_get_node_split_fids(INT4, INT4, INT4, INT4[])
...
+        SET tree_id=((sample_id-1)%('||num_of_tree||')+1)::INT;';
+
+    -- remove those invalid samples with join operation
+    -- add the weight field
+    EXECUTE 'DROP TABLE IF EXISTS '||target_table;
+    EXECUTE 'CREATE TEMP TABLE '||target_table||'(id, tid, nid, weight) AS
+        SELECT record_id, tree_id, tree_id, count(*) as weight  
+        FROM (sample_result_temp t join '||src_table||' k on t.record_id=k.id) l
+        GROUP BY record_id, tree_id
+    m4_ifdef(`__GREENPLUM__', `DISTRIBUTED BY(id)');';
```
aojwang: The update operation is very expensive, especially for large dataset. I think we can use the following SQL statement to create the target table so that we can avoid the updating. And we can remove the tree_id column from the sample_result_temp.  Although we need to calculate the tid and nid twice, i think it can do better than updating.

```
CREATE TEMP TABLE target_tale(id, tid, nid, weight) AS
SELECT record_id, (sample_id-1)%num_of_tree + 1 as tid, (sample_id-1)%num_of_tree + 1 as nid,  count(*) as weight
FROM sample_result_temp t, src_table k
WHERE  t.record_id=k.id
GROUP BY record_id, tid, nid;
```

methods/decision_tree/src/pg_gp/decision_tree.sql_in
haradh1, line 4033: Agreed.

methods/decision_tree/src/pg_gp/rf.sql_in
haradh1, line 880: This description is nice to include as a comment.

methods/decision_tree/src/pg_gp/rf.sql_in
haradh1, line 1085: I'm not saying let's do it completely automatic way.  We can have a short form where it infers the better way and users can also input their preference.

aojwang:
Hi Hitoshi,

I update the latest changes to the PR. As the QA wants to do testing base on the code in the GitHub (After this PR was merged, all the functionality were done. File reorg of next PR will not affect the functions), could you please review it and merge them ASAP.  Thanks very much.

methods/decision_tree/src/pg_gp/decision_tree.c
haradh1: > If you don't call srand, the random number sequence is the same for each call. right?

The random number sequence is same.  The generated random numbers are random. Why do you need different sequence for each call?

methods/decision_tree/src/pg_gp/decision_tree.sql_in
```diff
@@ -2225,16 +2225,28 @@ BEGIN
...
+    IF (sampling_needed) THEN
+        PERFORM MADLIB_SCHEMA.__sample_with_replacement
+            (
+            num_trees,
+            round(sampling_percentage * total_size)::INT,
+            training_table_name,
+            'tr_association'
+            );
+    ELSE
+        EXECUTE 'DROP TABLE IF EXISTS tr_association CASCADE';
```
haradh1: Don't you need to DROP tr_association when sampling_needed = true?

methods/decision_tree/src/pg_gp/rf.sql_in
haradh1: In that function, why do you need self join, with one side cross join?

haradh1: In that function, why when max_split_points is less than 1 do we use the same table?  Can we truncate the instance table at the beginning in that case? Can we INSERT INTO one table FROM the same table?

methods/decision_tree/src/pg_gp/decision_tree.c
cwelton: This is just wrong.  You shouldn't reset the seed at all.   Please remove this.

methods/decision_tree/src/pg_gp/rf.sql_in
cwelton, line 880: Agreed, special values like this that have mathematical roots need to be properly documented in the code so that people know why those values were chosen.

methods/decision_tree/src/pg_gp/decision_tree.c
cwelton: For some additional context, the database has a function srandom() that it calls during backend startup to provide a random seed.for the process as well as functionality to allow users to set random seeds.  It is not appropriate for you to mess with the random seed value in this way.   If, for some reason, we were to want to set the random seed directly it should be done via the database porting layer and call srandom() instead of calling srand() directly.

To that end it is more appropriate to use the random() function supplied by the database porting layer than to call rand() directly.

methods/decision_tree/src/pg_gp/decision_tree.c
```diff
@@ -1536,6 +1536,77 @@ enum DT_SCV_FINAL_ARRAY_INDEX
...
+    funcctx   = SRF_PERCALL_SETUP();
+    call_cntr = funcctx->call_cntr;
+    max_calls = funcctx->max_calls;
+ 
+    /* when there is more records to return */
+    if (call_cntr < max_calls)
+    {
+        int64       low			= PG_GETARG_INT64(1);
+        int64       high		= PG_GETARG_INT64(2);
+        float8      rand_num	= (rand()/(float8)(RAND_MAX+1.0));
```
cwelton: Please use random() from the database porting layer instead of calling rand() directly.

methods/decision_tree/src/pg_gp/rf.sql_in
renyi533, line 1085: Initially, my solution is quite similar to your suggestion above. During internal review in Beijing, we think it is not appropriate as the used formula is not fully validated in various conditions. So the short form statically call the parallel classification.

I will check with others to see whether we should add this again. Thanks.

haradh1:
I really suggest adding comments on these tables to explain what they are for and where they are used.
- training_instance
- training_instance_temp
- training_instance_aux
- node_size_aux
- tr_association
- sf_association
- classified_instance_ping
- classified_instance_pong

methods/decision_tree/src/pg_gp/rf.sql_in
renyi533: We use the join to add the missing <class, value> pairs. 

If one <class, value> pair does not exist in the original data, the grouping set operation will not generate it.

renyi533: If max_split_points is less than 1, then we will not enable the continuous feature discretization. That's why we have to check this.

We cannot truncate the instance table at the beginning. Otherwise, we cannot compute the ACS.

We do not insert into one table from the same table. We insert into a different table actually.

methods/decision_tree/src/pg_gp/decision_tree.c
renyi533: If the random number sequence is the same, each time we will sample the same data set. 

But we should get different set for each sampling operation.

renyi533: Based on Caleb's comments, I will use random() instead. It seems the backend already provide a random seed. Then, I do not need set the seed.

I will change and verify.

methods/decision_tree/src/pg_gp/decision_tree.c
renyi533: OK.

methods/decision_tree/src/pg_gp/rf.sql_in
gpgavin, line 880: The API doc already includes this as a footnote. However, I agree that we could put as comments.

methods/decision_tree/src/pg_gp/rf.sql_in
renyi533, line 1085: It is very difficult for us to find a good strategy in a tight time frame. After talking with Gavin, we can add this in next release.

I have filed a jira of MADLIB-487. What do you think?

methods/decision_tree/src/pg_gp/decision_tree.sql_in
renyi533: I will add the drop when sampling_needed = true. Thanks.

renyi533:
I will add comments for these tables.

methods/decision_tree/src/pg_gp/rf.sql_in
renyi533, line 880: I will add the comments.

renyi533:
I updated the code based on your comments. Please help to review again. Thanks.

haradh1:
Merged.  Thanks for your work!

gpgavin:
Hitoshi, thanks for the extensive efforts on reviewing this.

renyi533:
Thanks HItoshi and Caleb. Your comments are invaluable for us.

PR #135, opened 2012-04-28
fix madlib-481

renyi533:
This pull request fixes madlib-481.

We should set the auxiliary table as temporary.


haradh1:
Done. Thanks.

PR #136, opened 2012-05-03
Feature Extraction: - Improvement per review and test.

haradh1:
This rearranges assorted code that was making it inconvenient from user
perspective.  viterbi_mtbl and viterbi_rtbl are now created inside
text_feature_extraction() and index is created at the same time.
viterbi_mtbl was hard-coded in text_feature_extraction(), and now
it accepts argument.  vcrf_label() took an argument for a schema name
but it didn't use it in a good way, so now the argument is deprecated.
Also some tables in the install_test() that seemed to not be used are
removed entirely from the source.


haradh1:
Hi Kun,

I'd like you to review my additional improvement on the textfex.

haradh1:
Pushed with Kun's suggestion.

PR #138, opened 2012-05-03
Madpack: - Add -t option to specify cases to test.

haradh1:
This option is effective only in install-check at the moment.

MADLIB-482

### DIFF
--- a/src/madpack/madpack.py
+++ b/src/madpack/madpack.py
@@ -989,7 +989,8 @@ def main(argv):
         # 2) Run test SQLs 
         __info("> Running test scripts for:", verbose)   
         
-        caseset = set(args.testcase.split(',')) if args.testcase != "" else set()
+        caseset = set([test.strip()
+            for test in args.testcase.split(',')]) if args.testcase != "" else set()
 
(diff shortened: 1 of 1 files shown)


src/madpack/madpack.py
```diff
@@ -986,11 +989,18 @@ def main(argv):
         # 2) Run test SQLs 
         __info("> Running test scripts for:", verbose)   
         
+        caseset = set(args.testcase.split(',')) if args.testcase != "" else set()
```
fschopp: Is it worth adding a `strip()`?

```
[test.strip() for test in args.testcase.split(',')]
```

fschopp:
Good idea. Will merge once I have an answer re: my other question.

src/madpack/madpack.py
haradh1: I was just lazy because people will specify it without quotes in command line and without quotes you cannot contain spaces, but it seems nice.

haradh1:
Done.

fschopp:
Merged with commit d371f07c.

PR #139, opened 2012-05-04
New Types in C++ AL: Sparse Vectors and Function Pointers

fschopp:
_Please confirm with me before actually pulling these changes. I have not done tests, yet, but would like to open the discussion before being offline for the next two days._

Significant other changes:
- All files in the PostgreSQL connector now properly declare namespaces (as opposed to assuming that they will be #include'd in the right namespaces). The previous ugly workaround for Doxygen is no longer necessary. (Accounts for many modified files)
- The AbstractionLayer class was used like a namespace. This has been corrected in that the AbstractionLayer class has been completely removed. Any occurrences where inheritence was used to import names have been accordingly modified (and corrected).
- The caching of database metadata has been encapusalted in the new SystemInformation, FunctionInformation, and TypeInformation classes.
- Many backend functions have been wrapped by a madlib_\* function that puts a PG_TRY()/PG_CATCH() block around the function. This is in the new backend.hpp
- Simplified the files containing UDF declarations


src/ports/postgres/dbconnector/SparseVector_impl.hpp
```diff
@@ -0,0 +1,124 @@
...
+    }
+    add_run_to_sdata(reinterpret_cast<char*>(&runValue),
+        size - lastIndex, kValueLength, sdata);
+
+	// Add the final tallies
+	sdata->unique_value_count = sdata->vals->len / kValueLength;
+	sdata->total_value_count = size;
+
+	return svec_from_sparsedata(sdata, true /* trim */);
+}
```
haradh1, line 116: Is it possible that Eighen Sparse Vector may have 0-size vector?

haradh1:
What is the main use case in your mind?  I assume you are planning to add more code on top of this, but I don't see it here.  What I thought was to store ESV (index and value arrays) directly in the disk and when it's used mapping pointer to disk to Eigen's internal index values so that no expensive conversion happens.

src/ports/postgres/dbconnector/SparseVector_impl.hpp
fschopp, line 116: After default initialization, an Eigen sparse vector has `indices == NULL && values == NULL && nnz == 0`.

fschopp:
Yes, eventually we would like to also replace the on-disk representation. But for now, I am only interested in making sparse vectors generally available to C++ code. The storage aspect will be addressed later, and it will be completely transparent to the higher-level C++ code.

src/ports/postgres/dbconnector/dbconnector.hpp
```diff
@@ -58,107 +93,167 @@
...
+        _pgfunc _passedlist; \
+    } MADLIB_PG_DEFAULT_CATCH_AND_END_TRY; \
+}
+
+// The maximum number of arguments that can be passed to a function
+// Note that PostgreSQL defines FUNC_MAX_ARGS, which we could use here. However,
+// this can be a fairly large number that might exceed BOOST_PP_LIMIT_REPEAT.
+// In fmgr.h, PostgreSQL provides support functions (FunctionCall*Coll) for only
+// up to 9 arguments.
+#define MADLIB_FUNC_MAX_ARGS 9
```
haradh1: I'm not sure if this is for only internal (inside C/C++) invocation, but if it's for SQL invoked functions, too, 9 is too few.

src/ports/postgres/dbconnector/SystemInformation_proto.hpp
```diff
@@ -0,0 +1,237 @@
...
+    Oid getReturnType(FunctionCallInfo fcinfo);
+    TupleDesc getReturnTupleDesc(FunctionCallInfo fcinfo);
+    FmgrInfo* getFuncMgrInfo();
+    const char* getFullName();
+};
+
+/**
+ * @brief Cached information about the PostgreSQL system catalog
+ *
+ * In order to guarantee tyoe-safety through reflection/type-introspection, the
```
haradh1: typo: tyoe-safety

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
```diff
@@ -0,0 +1,558 @@
...
+
+namespace {
+
+/**
+ * @brief Initialize an OID-to-data hash table
+ */
+inline
+void
+initializeOidHashTable(HTAB*& ioHashTable, MemoryContext inCacheContext,
+    size_t inEntrySize, const char* inTabName, uint32_t inMaxNElems) {
```
haradh1, line 24: It looks like you create a hash table every time the function gets called (i.e., per one SELECT func()), which seems to me expensive.  Why do you need a hash table instead of plain memory chunk to store those information?

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
```diff
@@ -0,0 +1,558 @@
...
+                    /* typmod */ -1);
+                MemoryContextSwitchTo(oldContext);
+            } else {
+                cachedTypeInfo->tupdesc = NULL;
+            }
+            madlib_ReleaseSysCache(tup);
+        }
+    }
+    
+    return cachedTypeInfo;
```
haradh1, line 179: And if SystemInformation is for longer period than one function invocation (i.e., multiple SELECT func()), found entry should be rechecked if it is still valid.

src/ports/postgres/dbconnector/dbconnector.hpp
fschopp: Good point. The comment does not give an answer to this question. So here it is: This macro is only used to allow the usual `funcPtr(arg1, ..., argN)` syntax for a funcPtr of type FunctionHandle. Search for "BOOST_PP_REPEAT" in files AnyType_*.hpp.

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
fschopp, line 24: The hash table is created whenever fn_extra is NULL. So at most once per query and per function. So except for the first time it is called, a transition function always reuses the existing hash table.

Regarding the question why a hash table: Because we store catalog information for multiple OIDs (type information for each argument/return type, and function information for each function that our C++ code might invoke, including function information that called functions might invoke, i.e., in a transitive sense).

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
fschopp, line 179: We make the assumption that while fn_extra is non-NULL, it is still valid. I think backend code makes the same assumption at certain places. See, e.g., init_MultiFuncCall(), fmgr_sql(), etc.

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
fschopp, line 24: BTW: I use a PostgreSQL hash table here (as opposed to a C++ hash map) because it will eventually be garbage collected when its PG memory context is deleted. Therefore, no C++ destructors would be called and using C++ objects (unless they are [plain old data (POD)](http://en.wikipedia.org/wiki/Plain_old_data_structure)) would result in undefined behavior. The SystemInformation, FunctionInformation, TypeInformation classes are therefore POD.

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
```diff
@@ -0,0 +1,558 @@
...
+ */
+inline
+void
+getSystemInformationAndMemoryContext(FmgrInfo* inFmgrInfo,
+    SystemInformation*** outSysInfo, MemoryContext* outMemCtxt) {
+    
+    if (inFmgrInfo->fn_retset) {
+        // A set-returning function uses fn_extra to store cross-call
+        // information. See, e.g., init_MultiFuncCall() in funcapi.c.
+        // Fortunately, it stores a point to a struct FuncCallContext in
```
fschopp, line 63: s/point/pointer/

src/ports/postgres/dbconnector/SystemInformation_impl.hpp
haradh1, line 24: Yes, I've been looking for a good way to register C++ objects in fn_extra, but so far the only way to do it is to remember objects somewhere in the heap and hook transaction callback; ugly.

haradh1, line 24: For the hash table rationale, I know that there is a lot of objects you want to cache, but in most of the case it's almost empty.  I'm just afraid it's too expensive for simple functions.

fschopp, line 24: Why do you think creating a hash table is expensive? I haven't looked at the PG implementation too much, but in general this is a simple data structure. Also, for each function the hash table is created only once per query, so not a big deal I think. When would the execution time of a code path that is run only once per query, and that is probably in the order of nanoseconds, be relevant?

haradh1, line 24: I don't have a good sense how expensive it is.  I just thought it'd be nice if we could avoid when it's not needed.  May not be worth looking further.

src/dbal/EigenIntegration/HandleMap_proto.hpp
```diff
@@ -48,7 +63,12 @@ class EigenTypes<MapOptions>::HandleMap
      */
     HandleMap(const Handle &inHandle, Index inNumElem)
       : Base(const_cast<Scalar*>(inHandle.ptr()), inNumElem),
-        mMemoryHandle(inHandle) { }
+        mMemoryHandle(inHandle) {
+    
+//        madlib_assert(inHandle.size() >= inNumElem * sizeof(Scalar),
+//            std::runtime_error("Expected HandleMap to be larger than it "
+//                "actually is."));
```
cwelton, line 70: Commented out because...?

Would it make sense to have a compile option to enable heavier weight sanity checks?

fschopp, line 70: Hmm, I wanted to make all Handles aware of their size, including `TransparentHandle`s, which are currently just pointers. But this requires a few code changes, so I commented it out temporarily. I could obviously remove this change for now.

src/ports/postgres/dbconnector/SystemInformation_proto.hpp
```diff
@@ -0,0 +1,237 @@
...
+     * is known to be implemented on top of the C++ abstraction layer. NULL
+     * if not or unknown. If non-NULL, we this function pointer will be
+     * called directly, thus circumventing any detour through the backend.
+     */
+    UDF::Pointer cxx_func;
+    
+    /**
+     * Holds system-catalog information that must be looked up before a
+     * function can be called through fmgr. We store it here, because the
+     * information can be resued for multiple function calls.
```
cwelton: s/resued/reused/

src/ports/postgres/dbconnector/SystemInformation_proto.hpp
```diff
@@ -0,0 +1,237 @@
...
+     * An uninitialized flinfo is indicated by <tt>flinfo.fn_oid</tt> not
+     * being equal to \c oid.
+     *
+     * FIXME: Currently, no expression parse tree is stored. In fact, there
+     * is none, and we would have to artificially generate a default
+     * expression parse tree. This implies that function that are called
+     * from C++ functions currently do not support the following features:
+     * - Polymorphism
+     * - Anonymous return tuples
+     */
```
cwelton, line 104: Do we detect and error out in the event of Polymorphism and anon return tuples?  Or what is the behavior?

cwelton:
Aside from the couple small issues, this looks reasonable to me.  Lets try to get this in soon.  I'd like to wrap up "code complete".

Hitoshi, any final thoughts?

src/ports/postgres/dbconnector/AnyType_impl.hpp
```diff
@@ -508,77 +353,118 @@ T AbstractionLayer::AnyType::getAs() const {
...
 }
 
-#endif // MADLIB_DBCONNECTOR_HPP (workaround for Doxygen)
+/**
+ * @brief Return an AnyType object representing Null.
+ *
+ * @internal
+ *     An object representing Null is not guaranteed to be unique. In fact, here
+ *     we simply return an AnyType object initialized by the default
+ *     constructor.
```
haradh1, line 451: Can't it be a singleton?

src/ports/postgres/dbconnector/SystemInformation_proto.hpp
```diff
@@ -0,0 +1,237 @@
...
+};
+
+/**
+ * @brief Cached information about PostgreSQL functions
+ *
+ * For explanations, see struct FmgrInfo in fmgr.h,
+ * struct FuncCallContext in funcapi.h, enum TypeFuncClass in funcapi.h,
+ * and struct FormData_pg_proc in pg_proc.h.
+ */
+struct FunctionInformation {
```
haradh1, line 77: Do we support SRF?

haradh1:
I think it's ok.

src/ports/postgres/dbconnector/SystemInformation_proto.hpp
fschopp, line 104: Yes, it's currently detected implicitly. Since polymorphic/anonymous types cannot be resolved, the type conversion will fail. E.g., the C++ AL does not know how to convert from double to typeoid ANYELEMENT, so it will complain: "Invalid type conversion. Backend expects type ID 2283 ('anyelement') but supplied type ID is 701 ('float8')."

src/ports/postgres/dbconnector/AnyType_impl.hpp
fschopp, line 451: It could be, but what is the benefit of having (essentially) a global variable if it can be avoided?

src/ports/postgres/dbconnector/SystemInformation_proto.hpp
fschopp, line 77: No, currently not. But we might in the future.

src/ports/postgres/dbconnector/AnyType_impl.hpp
haradh1, line 451: It reduces a small amount of memory.  Not a big deal.  But I see this pattern in a lot of other software.

PR #141, opened 2012-05-07
Support CREATE EXTENSION, take two.

haradh1:
Instead of the previous attempt to support this in madpack time, this time
we try to make the build system to create EXTENSION files so that upcoming
PGXN support can handle it gracefully.  For users who want to install it
as an extension, the process is to download source, configure, and
"make extension-install."  Still the extension is not relocatable, users
can specify schema name on CREATE EXTENSION with SCHEMA option.

MADLIB-316

### DIFF
--- a/src/madpack/sort-module.py
+++ b/src/madpack/sort-module.py
@@ -1,0 +1,34 @@
+#!/usr/bin/python
+
+"""
+sort-module.py
+
+Given number of strings, sort them based on the module name which the
+string contains resolving the dependencies.  Example:
+
+  $ sort-module.py a/regress/b a/svec/b a/array_ops/b
+  > a/svec/b a/regress/b a/array_opts/b
+
+Note this script assumes to be run at the exact current directory,
+so you need change directory first to run it.
+"""
+
+import configyml
+
+
+portspecs = configyml.get_modules("../config")
+
+def find_order(path):
+  for i, moduleinfo in enumerate(portspecs['modules']):
+    modname = moduleinfo['name']
+    if modname in path:
+      return i
+  # return as the last if not found.
+  return len(portspecs['modules'])
+
+def main(args):
+  print " ".join(sorted(args, key = find_order))
+
+if __name__ == '__main__':
+  import sys
+  main(sys.argv[1:])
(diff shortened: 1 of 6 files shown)


haradh1:
Not seeing objection, I'm going to merge it today.

fschopp:
Let me take another look. I'll post here once I'm done reviewing.

src/madpack/sort-module.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python
+
+"""
+sort-module.py
+
+Given number of strings, sort them based on the module name which the
+string contains resolving the dependencies.  Example:
```
fschopp, line 7: I don't understand that comment. Could you rephrase that?

src/ports/postgres/CMakeLists.txt
```diff
@@ -236,6 +236,99 @@ macro(add_current_${PORT_LC}_version)
...
+    # We use X.Y.Z style for version string for it is recommended in PGXN.
+    set(EXTENSION_SQL "${CMAKE_CURRENT_BINARY_DIR}/extension/madlib--${MADLIB_VERSION_MAJOR}.${MADLIB_VERSION_MINOR}.${MADLIB_VERSION_PATCH}.sql")
+
+    # The workhorse for extension target
+    # It is hard to interpret Modules.yml in this cmake file, thus we use
+    # sort-module.py to sort modules based on the dependencies.
+    add_custom_command(OUTPUT ${EXTENSION_SQL}
+        COMMAND cat ${EXTENSION_SQL_BASE}
+            `cd ${CMAKE_BINARY_DIR}/src/madpack &&
+            python sort-module.py ${MODULE_SQL_POST_FILES}` >${EXTENSION_SQL}
```
fschopp, line 301: The uber-correct way would be using cmake commands like execute_process. I am not sure if this works if the user is using a shell we are not expecting.

fschopp:
Looks fine for now.

src/madpack/sort-module.py
haradh1, line 7: Draft.

<pre>
Sort input strings based on the module name in them such that the module dependencies are resolved.
</pre>

src/ports/postgres/CMakeLists.txt
haradh1, line 301: I tried execute_process to sort the file names, then the dependency order gets broken.  I leave it for now.

haradh1:
Thanks for your inputs!  Merged to master with the feedbacks.

PR #144, opened 2012-05-09
Fix for MADLIB-486

aojwang:
The PR includes the fix for [MADLIB-486](http://jira.madlib.net/browse/MADLIB-486).


cwelton:
How does this relate to the array length error described in MADLIB-486?

aojwang:
If we use DBL_EPSILON, then the split criterion value(SCV, such as infogain, gainratio, or gini) will not be treated as zero although they are really zero (6.66134E-16, 4.44089E-16 or -4.44089E-16). Therefore, we will grow the tree (create child nodes), since the SCV is not zero (If it's zero, we will treat it as leaf node). At this point, although we only have two discrete features, the tree will grow again and again (The SCV is not zero). Therefore, the number of parent discrete features (the array length) will be exceeded. Thanks.

PR #145, opened 2012-05-09
Fix for MADLIB-488

aojwang:
The PR adds the check for the parameter _max_tree_depth_ ([MADLIB-488](http://jira.madlib.net/browse/MADLIB-488)).


cwelton:
Fix looks safe to merge.

PR #146, opened 2012-05-10
Fix madlib-493

renyi533:
Currently, there are two existing issues related with missing value handling for continuous features

Firstly, even if users set 'IGNORE', we still replace the missing values.

Secondly, we replace the missing value with the existing minimum value. It is better for us to use average value instead.

We fixed this issue in this pull request.


methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -1235,40 +1235,46 @@ RETURNS void AS $$
...
     curstmt         TEXT := '';
     num             INT := 0;
-    min_val         FLOAT8;
+    avg_val         FLOAT8;
     coltable_name2  TEXT := coltable_name;
 BEGIN
     -- insert the continuous features
     IF (is_persistent) THEN
+        -- We use average value for a best guess here.
+        -- With average value, we expect to get the least deviations.
```
haradh1, line 1244: I'd say avg is still not right one for the missing value.  Is median better than avg?

renyi533, line 1244: Firstly, some papers use median and some others use average. Median is not explicitly superior to average.

Secondly, I don't know a median implementation in gpdb that runs as efficient as average.

haradh1, line 1244: Is there a good reason to use avg over min?

renyi533, line 1244: Sure. If you use min, how could you do binary split over it?

PR #148, opened 2012-05-10
Fix for MADLIB-497

aojwang:
After discussion, to make the behavior of pruning parameters consistent and make minimal risk(change little code), the solution is that the parameter _node_prune_threshold_ will only apply to the non-root node. Using this approach, we can ensure any one of the trained tree has at least one node (avoid empty tree). The corresponding comments are added to APIs in C45.sql_in and rf.sql_in. ([MADLIB-497](http://jira.madlib.net/browse/MADLIB-497))

1) The comments from rf_train

```
 * @param node_prune_threshold      The minimum number of records required in a 
 *                                  child node. It can't be NULL. The range of it is in [0.0, 1.0].
 *                                  This threshold only applies to the non-root nodes. Therefore, 
 *                                  if the percentage(p) between the sampled training set size of a tree
 *                                  (the number of rows) and the total training set size is less than
 *                                  or equal to the value of this parameter, then the tree only has
 *                                  one node (the root node); 
 *                                  if its value is 1, then the percentage p is less than or equal to 1
 *                                  definitely. Therefore, the tree only has one node (the root node).
 *                                  if its value is 0, then no nodes will be pruned by this parameter.
 * @param node_split_threshold      The minimum number of records required in a 
 *                                  node in order for a further split to be possible.
 *                                  It can't be NULL. The range of it is in [0.0, 1.0].
 *                                  If the percentage(p) between the sampled training set size of a tree
 *                                  (the number of rows) and the total training set size is less than
 *                                  the value of this parameter, then the root node will be a leaf one.
 *                                  Therefore, the trained tree only has one node.
 *                                  If the percentage p is equal to the value of this parameter, then the
 *                                  trained tree only has two levels, since only the root node will grow.  
 *                                  (the root node); 
 *                                  if its value is 0, then trees can grow extensively. 
```

2) The comments from c45_train.

```
 * @param node_prune_threshold      The minimum number of records required in a 
 *                                  child node. It can't be NULL. The range of it is in [0.0, 1.0].
 *                                  This threshold only applys to the non-root nodes. Therefore, 
 *                                  if its value is 1, then the trained tree only has one node (the root node); 
 *                                  if its value is 0, then no nodes will be pruned by this parameter.
 * @param node_split_threshold      The minimum number of records required in a 
 *                                  node in order for a further split to be possible.
 *                                  It can't be NULL. The range of it is in [0.0, 1.0].
 *                                  If it's value is 1, then the trained tree only has two levels, since 
 *                                  only the root node can grow; if its value is 0, then trees can grow 
 *                                  extensively. 
```


methods/cart/src/pg_gp/c45.sql_in
```diff
@@ -336,11 +336,17 @@ CREATE TYPE MADLIB_SCHEMA.c45_classify_result AS
  *                                  is 'explicit' or 'ignore'.
  * @param max_tree_depth            Specifies the maximum number of levels in the result DT 
  *                                  to avoid overgrown DTs. 
- * @param node_prune_threshold      The minimum number of cases required 
- *                                  in a child node.
- * @param node_split_threshold      The minimum number of cases required 
- *                                  in a node in order for a further split to 
- *                                  be possible.
+ * @param node_prune_threshold      The minimum number of records required in a 
+ *                                  child node. It can't be NULL. The range of it is in [0.0, 1.0].
```
haradh1, line 340: So, this is not the exact number of records, but the fraction.  If so, the first line of the comment is confusing.

aojwang, line 340: oh, right. Change the comments in both c45.sql_in and rf.sql_in.

haradh1:
Pulled.

aojwang:
Thanks for your efforts.

PR #149, opened 2012-05-10
Fix for MADLIB-498

aojwang:
Two things were done in this PR.
1) Change the left join to inner join
2) remove unnecessary columns when join.


methods/cart/src/pg_gp/dt.sql_in
```diff
@@ -770,8 +770,9 @@ BEGIN
                             nid,
                             tid 
                      FROM (
-                        SELECT t1.*, nid, tid, weight 
-                        FROM % t1 LEFT JOIN tr_association t2 ON t1.id = t2.id
```
aojwang, line 774: For RF, as we do sampling for each tree, some IDs in the table _t1_ will not be included in tr_association. At this point, left join will introduce garbage data. However, the garbage data will not be used to calculate the split criterion value, since we will ignore them (have no meaning).

haradh1:
Pulled.

PR #150, opened 2012-05-10
Support PGXN. (on top of PR-141)

haradh1:
Please look at only 86568d6e, as this is atop of PR-141.

```
The requirement of PGXN support is, to let users do

 $ pgxn install madlib

to install extension files in postgres install directory.  After that,
users can CREATE EXTENSION in SQL.

The way pgxn client works is download the zip file, unpack it, enter
the directory, configure if any, make install.  To be compatible with
this, we create a Makefile in the top directory when configured, and
let "make install" be "make -C build extension-install".  This becomes
a little surprising behavior to the normal usage out of PGXN, so
the configure creates such Makefile only if the packager indicates,
by touching PGXN_ZIP file in the top directory.  This is automated
with deploy/pgxn.sh, which clones the latest git and put an empty
PGXN_ZIP file, then zip it.

This commit also contains META.json, which is the only requirement
to upload a module to PGXN distribution site.
```

### DIFF
--- a/META.json
+++ b/META.json
@@ -1,0 +1,37 @@
+{
+    "name": "madlib",
+    "abstract": "An open source in-detabase analytics library",
+    "description": "An open source in-database analytics library",
+    "version": "0.3.0-alpha",
+    "maintainer": "MADlib development team",
+    "license": "bsd",
+    "provides": {
+        "madlib": {
+            "file": "madlib--0.3.0.sql"
+            "docfile": "ReadMe.txt",
+            "version": "0.3.0"
+        }
+    },
+    "resources": {
+        "homepage": "http://madlib.net/",
+        "bugtracker": {
+            "web": "http://jira.madlib.net/"
+        },
+        "repository": {
+            "url":  "https://github.com/madlib/madlib.git"
+            "web":  "https://github.com/madlib/madlib",
+            "type": "git"
+        }
+    },
+    "release_status": "testing",
+    "tags": [
+        "machine learning",
+        "analytics"
+    ],
+
+
+    "meta-spec": {
+        "version": "1.0.0",
+        "url": "http://pgxn.org/meta/spec.txt"
+    }
+}
--- a/deploy/pgxn.sh
+++ b/deploy/pgxn.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+# This is the version string for the distribution
+VERSION=0.3.0
+TEMPDIR=`mktemp -d -t madlib`
+ORIG=`pwd`
+cd $TEMPDIR
+git clone git://github.com/madlib/madlib.git madlib-$VERSION
+touch madlib-$VERSION/PGXN_ZIP
+zip -x "madlib-$VERSION/.git/*" -r $ORIG/madlib.zip ./madlib-$VERSION
+rm -rf madlib-$VERSION
+
--- a/src/madpack/sort-module.py
+++ b/src/madpack/sort-module.py
@@ -1,0 +1,34 @@
+#!/usr/bin/python
+
+"""
+sort-module.py
+
+Given number of strings, sort them based on the module name which the
+string contains resolving the dependencies.  Example:
+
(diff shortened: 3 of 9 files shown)


META.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "madlib",
+    "abstract": "An open source in-detabase analytics library",
+    "description": "An open source in-database analytics library",
+    "version": "0.3.0-alpha",
```
cwelton, line 5: Will we need to maintain this version separate from `src/config/Version.yml`, any way we can have this automatic so that we don't have to maintain the version number in multiple places?

cwelton:
We should be sure to include some sort of automated test around pgxn install support for Postgres.  Hopefully our firewalls allow for that...

META.json
haradh1, line 5: First of all, this string is apart from madlib version, strictly speaking.  PGXN distinguishes the version string of "distribution" (the packaged zip file) and the actual module extension, which is described below "provide" section.

And I really want to automate the version string with Version.yml.  It may be possible with configure, but this file should be located in the top directory which is out of build/, and may make packaging process complicated.

haradh1:
For now the http access to pgxn including its API is ok.  The client only uses http.  But not sure if in some days it's banned :(

haradh1:
We need an account in http://manager.pgxn.org/ to upload finally.

src/madpack/sort-module.py
```diff
@@ -0,0 +1,34 @@
...
+so you need change directory first to run it.
+"""
+
+import configyml
+
+
+portspecs = configyml.get_modules("../config")
+
+def find_order(path):
+  for i, moduleinfo in enumerate(portspecs['modules']):
```
fschopp, line 22: Minor issue, but all our Python code uses 4 spaces for indentation.

haradh1, line 22: That's a big issue, indeed.  I'll fix it.

deploy/pgxn.sh
```diff
@@ -0,0 +1,12 @@
```
fschopp, line 0: Why don't you you clone the current repository instead of the remote one? Also, I wonder if we'd rather have a CMake script instead of a shell script, like for the other deployment targets.

META.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "madlib",
+    "abstract": "An open source in-detabase analytics library",
+    "description": "An open source in-database analytics library",
```
fschopp, line 4: s/open source/open-source/

META.json
fschopp, line 5: Couldn't the META.json file be generated if configure is called by pgxn? Or is META.json necessary in the zip file? In that case, it might only be necessary when deployed to pgxn?

Bottom line: I am wondering if we can get rid of this file in the "normal" repository.

fschopp:
Looks fine for now. However, it would be nice if we had a few comments in the repository. Can you add the description of this pull request to a prominent location within one of the pgxn support files?

META.json
haradh1, line 5: Yes, META.json is necessary in the zip file.  I think we could automate PGXN deployment including this file, but the "how" is still not clear.

deploy/pgxn.sh
haradh1, line 0: Hmm, that's a good idea to clone the current repository.

haradh1, line 0: Do you mean RPM package by "the other deployment targets"?  The difference from it is that PGXN package needs almost raw source tree.  I'm not sure about "install" process in cmake, but as I created "extension-install" target for EXTENSION, the cmake install process is not applicable here, I think.

haradh1:
Thanks for your inputs.  Pushed to master.

deploy/pgxn.sh
fschopp, line 0: I thought the product of the PGXN deployment target would be a zip file. Sure, there are some differences in what it includes, but still the expected behavior would be the same as for other deployment targets like RPM or a Mac OS X package installer.

PR #151, opened 2012-05-15
Fix for MADLIB-505

aojwang:
array_upper will return null if the input array is null or empty, which will cause the csvstr_to_array failure. [MADLIB-505](http://jira.madlib.net/browse/MADLIB-505)

The solution is return NULL directly if the input string for the function csv_to_array is null or empty after trimmed with ' '.


methods/cart/src/pg_gp/dt_utility.sql_in
```diff
@@ -456,14 +459,20 @@ DECLARE
...
-    ret = string_to_array(lower(csv_str), ',');
-    
-    IF (ret IS NOT NULL) THEN
-        FOR index IN 1..array_upper(ret, 1) LOOP
-            ret[index] = btrim(ret[index], ' ');
-        END LOOP;
+    -- if the input array is NULL or an empty one,
+    -- then we return NULL directly 
+    -- (array_upper will return NULL otherwise) 
+    IF (csv_str IS NULL OR length(btrim(csv_str, ' ')) = 0) THEN
```
haradh1: How about adding this, instead of your change

``` sql
ret = string_to_array(lower(csv_str, ',');
IF array_upper(ret, 1) IS NULL THEN
    RETURN NULL;
END IF;
```

aojwang: Good suggestions. Thanks.
Add an additional function _btrim_ before lower, and add the corresponding comments. Thanks.

haradh1:
Pulled.

aojwang:
Thanks for your efforts.

PR #152, opened 2012-05-15
k-Means Rewrite

fschopp:
_This is a big pull request. It may be advisable to look at the individual commits that I grouped into logical units. Please do not merge without checking back with me._

k-Means:
- Complete redesign and rewrite. Main new features: 1) The new implementation is nothing more but an iteratively called aggregate function. No conversion between temporary tables and iteration states takes place. 2) The distance metric is fully pluggable. 3) Clear separation between the seeding algorithm and Lloyd's local search heuristic 4) The new code should be much easier to read and maintain
- Uses C++ AL
- First incarnation of general Python abstraction layer (currently still in kmeans.py, though). Main feature: No data movement between backend and PL/Python.

Utilities:
- Implemented a C function exec_sql_using() that can be used to emulate the EXECUTE ... USING command that is not available before PostgreSQL 8.4. This seems to be the only "guaranteed lossless" way to write function arguments into a table.

Linear-Algebra Operations:
- Function closest_column() for determining the column of a matrix that is closest to a vector, according to the given metric
- Function squared_dist_2norm() for determining the square of the Euclidean distance between two vectors

Both functions will be needed for k-means

C++ AL:
- Increased support for Eigen types: (Non-mapped) vectors and matrices can now directly be converted to AnyType (and thus returned or passed to FunctionHandles)
- Added types [Mutable]Mapped{ColumnVector|Matrix} that can now be used as template parameters for AnyType::getAs
- ArrayHandles now understand multi-dimensional PostgreSQL arrays. We store matrices as an array of columns, i.e., in PostgreSQL one would write matrix[col][row] whereas in C++ code the row index comes first (as one would expect)

Linear/logistic regression:
- Minor changes to make use of the [Mutable]MappedColumnVector types.

Sampling module:
- Added a data-parallel UDA for weighted sampling of a single row
- This uses boost's Mersenne Twister, and the random_device class for initialization. Instead of linking to boost's random library, I for now added boost's random_device.cpp to the repository. As we move forward, we probably want to extend our build system to build boost libraries on demand. Because of problems with the boost v1.46.1 headers in the context of using random_device, I bumped the required boost version to v1.47 and the automatically downloaded version to 1.49.0.


agorajek:
Nice Florian, Kmeans finally in-line with the abstraction layer, plus wrapped in a UDA. Looks pretty awesome!

jhellerstein:
Great stuff.  The k-Means design sounds really good -- an exemplar of how to do things in future?  

Some SW eng thoughts from the peanut gallery:

1) Looking forward, I'd advise a culture that avoids cross-module pull reqs.  Having that discipline in the core team will be good for engineering, and help set the community practice as well.
2) I see we're dependent on both Boost and Eigen?  We should assess our existing external dependencies, and have a process for approving the addition of new dependencies.

Joe

src/modules/random/mersenne_twister.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+std::mt19937&
+mersenneTwister() {
+    static bool sInitialized = false;
+    static std::mt19937 sMersenneTwister;
+    
+    if (!sInitialized) {
+        // We assume that seeds generated with random_device are completely
+        // uncorrelated and thus reasonable choice even for multiple
+        // pseudo-random number streams that may be greated with different seeds
```
haradh1: s/greated/created/

haradh1:
I also request to clean up this pull request, now that the part of this is in master branch.  I have some additional questions.
- it is not clear why we introduce MT RNG.  It looks easy to use it, but additional cost including wrapping namespace, library dependency doesn't seem worth paying for the benefit.  Why is the orthodox random() not enough here?
- why did you choose plpgsql over plpython for the driver UDF?
- exec_sql_using is questionable.  We have been building raw SQL text rather than using EXECUTE USING.  I wonder why we need it now.

methods/utils/src/pg_gp/store_args.c
```diff
@@ -0,0 +1,94 @@
...
+    for (int i = 1; i < nargs; i++)
+        if (PG_ARGISNULL(i)) {
+            if (nulls == NULL) {
+                nulls = palloc0(sizeof(char) * (nargs - 1));
+                memset(nulls, ' ', nargs - 1);
+            }
+            nulls[i - 1] = 'n';
+        }
+    
+	SPI_connect();
```
haradh1: Indentation.

methods/utils/src/pg_gp/store_args.c
```diff
@@ -0,0 +1,94 @@
...
+        ereport(ERROR, (
+            errmsg("function \"%s\" called with NULL as first argument",
+                format_procedure(fcinfo->flinfo->fn_oid))
+            ));
+    
+    char* stmt = NULL;
+    if (types[0] == TEXTOID)
+        stmt = DatumGetCString(
+            DirectFunctionCall1(textout, PG_GETARG_DATUM(0)));
+    else if (types[0] == VARCHAROID)
```
haradh1: I don't think we want both of text and varchar types

fschopp:
> I also request to clean up this pull request, now that the part of this is in master branch.

Yes, I'll do a rebase.

> - it is not clear why we introduce MT RNG.  It looks easy to use it, but additional cost including wrapping namespace, library dependency doesn't seem worth paying for the benefit.  Please justify.

We need to generate random numbers for sampling. I could have added a class to the C++AL that makes use of the DBMS's PRNG, but that would have been more code.

> - why did you choose plpgsql over plpython for the driver UDF?

There is only boilerplate code in plpgsql, and the purpose of this code is writing the raw arguments into a table. The benefit of plpgsql here is that it is a native language, and no internal conversion takes place. The plpgsql code should later be generated, of course.

All actual driver logic is in PL/Python. Since the plpgsql wrapper has written all arguments to a temporary table, there is no data movement between PL/Python and the backend, i.e., no conversion between native datums and Python objects takes place.

> - exec_sql_using is questionable.  We have been building raw SQL text rather than using EXECUTE USING.  I wonder why we need it now.

Raw SQL text is clearly inferior to EXECUTE USING because it contains inefficient and potentially lossy to/from-text conversions. For k-means, we can have very large arguments (the array of centroids could easily be several MBs), and we certainly don't want to generate SQL text. Obviously, that would be even worse than conversion to Python objects.

methods/utils/src/pg_gp/store_args.c
fschopp: Fixed.

src/ports/postgres/modules/sample/sample.sql_in
```diff
@@ -0,0 +1,70 @@
...
+ *     description \ref grp_sample.
+ *
+ *//* ----------------------------------------------------------------------- */
+
+/**
+@addtogroup grp_sample
+
+@about
+
+Random-sampling functions.
```
haradh1: We need more description.

src/modules/sample/weighted_sample.cpp
```diff
@@ -0,0 +1,112 @@
...
+AnyType
+weighted_sample_transition::run(AnyType& args) {
+    WeightedSampleTransitionState<MutableArrayHandle<double> > state = args[0];
+    uint64_t identifier = args[1].getAs<int64_t>();
+    double weight = args[2].getAs<double>();
+    
+    // Instead of throwing an error, we will just ignore rows with a negative
+    // weight
+    if (weight > 0.) {
+        state.weight_sum += weight;
```
haradh1: Can you explain why this is a valid weighted sampling?  It seems to me this code accumulates the weight from the beginning of the row and randomly select the row identifier value if it's success.  Doesn't this mean it relies on the order of input, since the weight_sum sees only the rows seen up to now and it gets different if the order changes.  In that sense, I don't think this can be achieved with a single simply aggregate, but it needs something like window functions or multiple queries.  Please point me if I'm wrong.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,301 @@
...
-    @param evaluate Boolean flag indicating whether to compute some model 
-           evaluation coefficients
-    @param out_points Name of the table with data points and assigned centroids
-    @param out_centroids Name of the table with discovered centroids
-    @param p_verbose Boolean flag indicating weather to display INFO messages 
-           during the execution           
+    Driver function for k-Means++ seeding
+    
+    @param schema_madlib Name of the MADlib schema, properly escaped/quoted
+    @rel_args Name of the (temporary) table containing all non-template
```
haradh1: @param is missing.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,301 @@
...
-        # Create new temp table (i)
-        __run_quietly( 'DROP TABLE IF EXISTS TempPoints' + str(i));
-        __run_quietly( 'CREATE TEMP TABLE TempPoints%s (like TempPoints%s)' 
-                        % (str(i), str(i-1)));
+            it.update("""
+                SELECT _args.initial_centroids FROM {rel_args} AS _args
+                """)
+        while it.test("array_upper(_state._state, 1) < _args.k"):
+            it.update("""
+                SELECT _state._state || _src.{col_point}
```
haradh1: Will this have duplicate elements? Or do you have any guard to avoid it?

src/ports/postgres/modules/kmeans/kmeans.sql_in
```diff
@@ -257,571 +259,309 @@ out_centorids | km_c
...
-RETURNS MADLIB_SCHEMA.kmeans_result
-AS $$
-
-    PythonFunctionBodyOnly(`kmeans', `kmeans')
+ * @param rel_source Name of the relation containing input points
+ * @param col_id Name of the column containing the unique identifiers
+ * @param col_point Name of the column containing the point coordinates
+ * @param initial_centroids Matrix containing the initial centroids as columns
+ * @param squared_dist_metric A function with signature
+ *     <tt>DOUBLE PRECISION[] x DOUBLE PRECISION[] -> DOUBLE PRECISION</tt> that
```
haradh1: How do we enforce the function signature?

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,301 @@
...
-                    , temp_target = temp_target
-                    , min_distance = dist_func.replace( '&&&', 'p.coords, c.coords')
-                    , centroids = output_centroids
-                    , prev_cid = i-1
-                )
-            );
+        resultObject = plpy.execute("""
+            SELECT ({condition})::BOOLEAN AS condition
+            FROM {{rel_args}} AS _args
+                LEFT OUTER JOIN (
```
haradh1: I don't get why this is LEFT OUTER JOIN.  Both of arg and state are assumed to be 1 row?

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,301 @@
...
+                    _state {stateType}
+                );
+                """.format(
+                    temp = 'TEMPORARY' if self.temporaryTables else '',
+                    **self.kwargs))
+        self.inWith = True
+        return self
+    
+    def __exit__(self, type, value, tb):
+        self.inWith = False
```
haradh1: Is this With support in the middle?  It seems self.inWith is not used effectively.  I'd like to see drop table here in __exit__

src/ports/postgres/dbconnector/TypeTraits.hpp
```diff
@@ -13,173 +13,429 @@
...
-        const dbal::eigen_integration::ColumnVector>,
-    dbal::ArrayType,
-    dbal::Immutable,
-    PointerGetDatum(value.memoryHandle().array()),
-    ArrayHandle<double>(reinterpret_cast<ArrayType*>(madlib_DatumGetArrayTypeP(value)))
-);
+template <>
+struct TypeTraits<double> {
+    typedef double value_type;
+
```
haradh1: Let's remove trailing tabs and white spaces.

src/ports/postgres/modules/sample/sample.sql_in
fschopp: True. Do you volunteer to write it? :)

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,301 @@
...
+    def __init__(self, rel_args, rel_state, stateType,
+            temporaryTables = True,
+            truncAfterIteration = False,
+            schema_madlib = "MADLIB_SCHEMA_MISSING",
+            **kwargs):
+        self.kwargs = kwargs
+        self.kwargs.update(
+            rel_args = ('pg_temp.' if temporaryTables else '') + rel_args,
+            rel_state = ('pg_temp.' if temporaryTables else '') + rel_state,
+            unualified_rel_state = rel_state,
```
haradh1: "unqualified"?

methods/utils/src/pg_gp/store_args.c
fschopp: True, one is enough, but then it does not really matter I think.

src/modules/sample/weighted_sample.cpp
fschopp: https://github.com/fschopp/madlib/tree/design_integrate/doc/design

src/ports/postgres/modules/sample/sample.sql_in
haradh1: No, thanks ;-)

src/ports/postgres/dbconnector/TypeTraits.hpp
fschopp: Yes, Xcode has been bugging me a bit with leaving trailing white spaces. It seems http://code.google.com/p/google-toolbox-for-mac/wiki/GTMXcodePlugin provides a solution.

What's yours?

haradh1: I'm always on mvim.

```
set listchars=tab:>-,trail:-
```

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: I don't understand your first question.
self.inWIth should be checked in the update method. Good observation.
In the past, not dropping the temporary table has been very useful for debugging and workaround. The state table is most likely not too large, so I want to keep it around. After all it is temporary anyway.

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: At the beginning, the state table might still be empty. But the user might still want to check some argument. -> OUTER JOIN

fschopp: And no, the state table is not necessarily just one row. If self.truncAfterIteration == False, then all states will be kept.

fschopp: Keeping states is a reasonable choice, e.g., for numerical algorithms, which are not guaranteed to converge and you might want to return to a previous solution.

src/ports/postgres/modules/kmeans/kmeans.py_in
haradh1: My question was if you're still working on this "inWith".
Yeah, I see keeping the intermediate tables is very useful for development and trouble shooting.  The problem is, though, how to hide it from the normal use case, as you see.

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: Existing points will have probability (weight) 0 of being chosen.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp: Other signatures will cause a runtime error. This will happen implicitly in the C++ AL.

haradh1: Can you point me to the file?  FunctionHandle?

src/ports/postgres/modules/kmeans/kmeans.py_in
haradh1: Yes, state may have more than 1 row but here you filter the row by "iteration"

fschopp: Iteration may be 0 in the beginning and match no rows.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp: https://github.com/madlib/madlib/blob/master/src/ports/postgres/dbconnector/FunctionHandle_impl.hpp#L78

src/config/Modules.yml
```diff
@@ -7,20 +7,23 @@ modules:
...
     - name: compatibility
+      depends: ['utilities']
     - name: conjugate_gradient
       depends: ['array_ops']
     - name: data_profile
       depends: ['sketch']
     - name: cart
     - name: kmeans
-      depends: ['array_ops','svec']
+      depends: ['array_ops','sample','svec','utilities']
```
haradh1: It seems to me kmeans is not dependent on svec anymore.  Everything is converted to float[]?

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: It's a temporary table with a reserved name (_madlib_*). I don't know what you want to hide more and why.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,301 @@
...
+                CAST((_args.initial_centroids, 1.0, 0) AS
+                    {schema_madlib}.kmeans_result)
+            FROM {rel_args} AS _args
+            """)
+        while it.test("""
+            {iteration} < _args.max_num_iterations AND
+            (_state._state).frac_reassigned > _args.min_frac_reassigned
+            """):
+            it.update("""
+                SELECT {schema_madlib}.kmeans_step(
```
haradh1: So... this means we won't store the relationship between data points and centroids but transition function always calculates all distances and find the closest one even though it was already calculated in the previous iteration.  I don't know what is our main target zone, but this will get a big overhead if the number of K is like over 1,000.

fschopp: Yes, one would expect that there is a turning point where the cost of storing the distances is less than the cost of recomputing. Then again, the factor is at most 2, and we all know that updating many table rows in each iteration is a pain of its own. So I would assume that this implementation is a lot more performant for most typical inputs.

src/ports/postgres/modules/kmeans/kmeans.sql_in
haradh1: Hmm, I see.  It would be much easier with some check function and domain type...

src/ports/postgres/modules/kmeans/kmeans.py_in
haradh1: I agree it's faster in a small k but not sure about up to how many K we can calculate.  Maybe we need to investigate it during the QA test.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp: The C++AL promises complete type safety so it needs to check anyway. An option might be to make FunctionHandle a template type where the template parameter is a function signature. Then the validation could already occur in the with the getAs<...>()  function. But I don't know if it's worth the effort for now.

haradh1: What is the name of this ".template" technique...?

```
            = inMetric(inMatrix.col(i), inVector).template getAs<double>();
```

src/ports/postgres/modules/kmeans/kmeans.py_in
fschopp: I agree it's worth finding out. But then, even if maintaining a temporary table could be done in no time, the speedup is at most 2.

src/ports/postgres/modules/kmeans/kmeans.sql_in
fschopp: Here, template is used as a qualifier. Since inMatrix and inVector are template arguments, getAs is considered a dependent name (even though FunctionHandle::operator() returns a non-template type -- I'm currently not sure if this is according to the C++ standard or whether just gcc has a deficiency here). So the parser would get confused without the template qualifier and treat the first "<" as a comparison.

fschopp: If you happen to know the appropriate section in the C++ Standard re: whether the template qualifier is really necessary (according to the standard), please post it.

haradh1: Wow.  (template getAs<double>()).

haradh1:
I think I've done the first round review.  The code seems ok but we need to verify the speed and scalability is better than before by running actual test cases as soon as possible.

cwelton:
After discussion with Florian I think we would like to defer pulling this until after the v0.4 release.  

We simply don't have sufficient time to adequately test without substantial changes to our existing schedule.

src/modules/linalg/avg_vector.cpp
```diff
@@ -0,0 +1,151 @@
```
fschopp: This is a ridiculously long implementation for such simple functionality. MADLIB-579 will hopefully make this trivial.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,333 @@
...
+                                _src.{col_id},
+                                ({schema_madlib}.closest_column(
+                                    _state._state,
+                                    _src.{col_point},
+                                    _args.squared_dist_metric
+                                )).distance
+                            )
+                        FROM
+                            {rel_source} AS _src, {rel_args} AS _args,
+                            {rel_state} AS _state
```
haying: {rel_state} AS _state is forcing the confusing use of _state._state in the query. I am changing it to be _rel_state in my later pull request to Florian's kmeans branch.

src/ports/postgres/modules/kmeans/kmeans.py_in
```diff
@@ -3,1048 +3,333 @@
...
+                            '{rel_source}',
+                            '{col_id}',
+                            '{col_point}',
+                            CAST(_args.k AS INT2),
+                            _args.squared_dist_metric,
+                            (_state._state).centroids
+                        ), (_state._state).frac_reassigned, 0)
+                    FROM {rel_state} AS _state, {rel_args} AS _args
+                    WHERE _state._iteration = {iteration}
+                    """)
```
haying: Is there a specified reason that we call it.test() with the "it." but use iteration, rel_args, rel_state to refer it.iteration, it.rel_args, it.rel_state?

src/ports/postgres/modules/kmeans/kmeans.sql_in
```diff
@@ -257,571 +259,340 @@ out_centorids | km_c
...
+    initial_centroids DOUBLE PRECISION[][],
+    squared_dist_metric REGPROC /*+ DEFAULT 'squared_dist_norm2' */,
+    agg_mean REGPROC /*+ DEFAULT 'avg(float8)' */,
+    max_num_iterations INTEGER /*+ DEFAULT 20 */,
+    min_frac_reassigned DOUBLE PRECISION /*+ DEFAULT 0.001 */
+) RETURNS MADLIB_SCHEMA.kmeans_result AS $$
+DECLARE
+    theIteration INTEGER;
+    theResult MADLIB_SCHEMA.kmeans_result;
+    oldClientMinMessages VARCHAR;
```
haying: Since we are building the abstract layer, I am mentioning this. Local names like "oldClientMinMessages" does not seem to make sense for SQL, a case-insensitive language.

src/ports/postgres/modules/kmeans/kmeans.py_in
haying: I guess I got it. Because the string is passed to the controller class and gets formatted inside the class?

src/ports/postgres/modules/kmeans/kmeans.sql_in
```diff
@@ -257,571 +259,340 @@ out_centorids | km_c
...
+        (SELECT setting FROM pg_settings WHERE name = 'client_min_messages');
+    EXECUTE 'SET client_min_messages TO warning';
+    PERFORM MADLIB_SCHEMA.create_schema_pg_temp();
+    -- Unfortunately, the EXECUTE USING syntax is only available starting
+    -- PostgreSQL 8.4:
+    -- http://www.postgresql.org/docs/8.4/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-EXECUTING-DYN
+    -- We therefore have to emulate.
+    PERFORM MADLIB_SCHEMA.internal_execute_using_kmeanspp_seeding_args($sql$
+        DROP TABLE IF EXISTS pg_temp._madlib_kmeanspp_args;
+        CREATE TEMPORARY TABLE _madlib_kmeanspp_args AS
```
haying: NOT CONSISTENT with line 355 (the same statements in kmeans()):
     354    +        DROP TABLE IF EXISTS pg_temp._madlib_kmeans_args;
     355    +        CREATE TABLE pg_temp._madlib_kmeans_args AS
BTW. "CREATE TEMPORARY TABLE pg_temp._madlib_kmeans_args AS" is not allowed in GPDB: ERROR: temporary table may not specify a schema name

src/config/Modules.yml
fschopp: Yup. Also, array_ops and utilities are only needed for unit testing.

src/ports/greenplum/4.0/config/Modules.yml
```diff
@@ -7,13 +7,14 @@ modules:
...
     - name: compatibility
+      depends: ['utilities']
     - name: conjugate_gradient
       depends: ['array_ops']
     - name: data_profile
       depends: ['sketch']
 #    - name: cart
     - name: kmeans
-      depends: ['array_ops','svec']
+      depends: ['array_ops','sample','svec','utilities']
```
fschopp: see above

src/ports/greenplum/4.0/config/Modules.yml
```diff
@@ -7,13 +7,14 @@ modules:
       depends: ['svec']
     - name: bayes
     - name: compatibility
+      depends: ['utilities']
```
fschopp: really needed?
_Update_: Ah yes, for create_schema_pg_temp()

haying: Can we add comment to this file to remind us something like this in the future?

fschopp: We could add a YAML comment.

cwelton:
Florian - in your assessment is this ready to pull once we have the tests adapted to the new interface?

fschopp:
Yes, once the tests are adapted, this is ready to merge (though I cannot guarantee that it passes all tests yet).

cwelton:
Based on performance testing the new code is running up to 200x slower than the old code and will fail with errors on some data sets where the prior implementation did not.  

This code is not ready for inclusion at this time.

Example: on the us census data set the code went from under 1 minute to 156 minutes.

fschopp:
New changes with significant performance improvements imminent. Reopening.

PR #153, opened 2012-05-16
Fix for MADLIB-509

aojwang:
If the best feature for the current leaf node is continuous, then the new created child nodes will inherit the dp_ids from this leaf node(the parent node).

dp_ids: the discrete features' IDs used by ancestor nodes.


methods/cart/src/pg_gp/dt.sql_in
```diff
@@ -2483,22 +2484,32 @@ m4_changequote(>>>`<<<, >>>'<<<)
                 
                 EXECUTE curstmt;
                 
+                IF (dp_ids IS NULL) THEN
+                    dp_ids_text = 'NULL'::TEXT;
+                ELSE
+                    dp_ids_text = MADLIB_SCHEMA.__format
+                                    ('ARRAY[%]', array_to_string(dp_ids, ','));
```
haradh1, line 2491: Have you considered the case where dp_ids is an empty array?

haradh1:
Pulled.

methods/cart/src/pg_gp/dt.sql_in
aojwang, line 2491: Sorry for responding later. My register mail is gmail, and i can't receive the mail from it in office.
The dp_ids should not be empty. Its initial value is null (the root node, or the nodes whose ancestor nodes are continuous), after a node's best split is a discrete feature, then an element will be added into this array. Thanks.

PR #154, opened 2012-05-18
Madlib 474

renyi533:
Fix madlib-474. The algorithm will throw error message when NULL element is present in the dictionary.

I verified the case in the jira of madlib-474. Ruilong also think the fix is OK.


methods/plda/src/pg_gp/plda.py_in
```diff
@@ -27,9 +27,17 @@ def plda_train(madlib_schema, num_topics, num_iter, alpha, eta, data_table, dict
...
 	if (dsize == 0):
 	    plpy.error("error: dictionary has not been initialised")
+	
+	dict = plpy.execute("SELECT dict FROM " + dict_table)
+	dict_string = dict[0]['dict'].lower()
+	dict_array = dict_string.strip('{} ').split(',')
+	for word in dict_array:
+		word = word.strip(' ')
+		if (word == 'null'):
+			plpy.error("error: Invalid NULL element in the dictionary")
```
fschopp: This looks like inconsistent indentation (cf. line 32). We use 4 spaces for indentation and no tabs.

renyi533: In the beginning of each line, we use a tab currently. 

So we should only use tab in the beginning of each line. Is that right? Or, should I change all of the tabs in existing files to 4 spaces?

renyi533: I think we should still use tab in the beginning of a line. Otherwise, there are too many changes. It is hard for the reviewer to review.

fschopp: "We use 4 spaces for indentation and no tabs."

fschopp: If you have a strong opinion about either using tabs or not, please file a JIRA and add you arguments.

haradh1:
I don't think this is a good fix.  It seems you extract array value from the dictionary and convert to python text, split it, and examine each element in it if it's "null" or not, and it doesn't sound robust.  I think sampleNewTopics function already captures the error and issue here is only the error message, so the right place to add code should be in sampleNewTopics where we check if the input array contains NULL element.  Here, we should add more precise message, rather than non-robust check in plda_train.

methods/plda/src/pg_gp/plda.py_in
renyi533: In the existing file, we already use tabs in the beginning for each line in plda's code. 

I agree it is nice to convert all these tabs to spaces. But I think we should do it in a separate thread. 

For these newly introduced tabs, I already changed to spaces.

renyi533:
I will investigate further to propose a better solution.

haradh1:
Pulled with my edit.

PR #157, opened 2012-05-24
Fix MADLIB-517

renyi533:
We should error out early in the training process. Otherwise, some exceptions may occur.


cwelton:
Pull request looks reasonable and should be safe to merge.

One improvement that would be desirable:
- Error messages should conform to the Postgres error message standards:
  - Do not begin an error message with Capital letters
  - Do not end an error messages with puctuation, especially an "!"

The 8,000,000 limit seems somewhat arbitrary, there should be some comments or justification regarding the reasons for the specific limit.

haradh1:
Pulled.

PR #158, opened 2012-05-25
Enhance the feature of max split point

renyi533:
Fix madlib-514.

### DIFF
--- a/methods/cart/src/pg_gp/dt.sql_in
+++ b/methods/cart/src/pg_gp/dt.sql_in
@@ -465,8 +465,6 @@ $$ LANGUAGE PLPGSQL;
  *                              for current leaf nodes.
  * @param class_table_name      The name of the class table.
  * @param acc_table_name        The name of the table containing ACC set.
- * @param max_split_point       The upper limit of sampled split points for 
- *                              continuous features. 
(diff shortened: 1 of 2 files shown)


methods/cart/src/pg_gp/dt.sql_in
```diff
@@ -829,48 +764,91 @@ $$ LANGUAGE PLPGSQL;
...
 DECLARE
+    stmt    TEXT;
 BEGIN
+
+    IF (coalesce(max_split_point,0)>0 AND cont) THEN
+        stmt = MADLIB_SCHEMA.__format
+                (
+                    'SELECT tid,nid,
+                        avg(fval) OVER  
+                        (PARTITION BY tid,nid,row_asc/((row_asc+row_desc-1)/%+1))
```
haradh1: Isn't this something like percentile_rank()?  At least row_asc+row_desc means count.  It's faster to reduce the number of distinct window specification (order by fval asc/desc can be one).

renyi533: OK. I used one window function of cume_dist to do the task now. Please review again.

haradh1:
Pulled.

PR #161, opened 2012-05-29
fix madlib 474

renyi533:
Please see log below. The index of words should be in the range of [1, dict_size]. But some indexes are larger than 25335. This should be the root cause. The null value cannot fully resolve this issue. The macros, such as ARR_HASNULL, may generate wrong result.

I also provided a function named "check_array_null_elem" to accurately check the null element in array. 

testdb=# select count(*) from (select unnest(dict) from madlibtestdata.plda_nsf_abs_dict) l;
##  count 

 25335
(1 row)

testdb=# select \* from (select  unnest(contents) as widx from madlibtestdata.plda_nsf_abs_corpus) l where widx > 25335;
##  widx  

 27653
 26404
 27659
 29462
 28241
 29137
 29137
 25640
 25640
 25678
 25800
 25800
 30359

### DIFF
--- a/methods/plda/src/pg_gp/plda.c
+++ b/methods/plda/src/pg_gp/plda.c
@@ -245,52 +245,6 @@ static int32 sampleTopic
 
 
 /**
- * This function detects whether an array has any null elements
- *
- * This gives an accurate answer, whereas testing ARR_HASNULL only tells
- * if the array *might* contain a null.
- * Parameters
- *  @param p_array The array to be checked
- *
- */
-static bool
-check_array_null_elem(ArrayType *array)
-{
-   int        nelems;
-   bits8      *bitmap;
-   int        bitmask;
-
-   /* Easy answer if there's no null bitmap */
-   if (!ARR_HASNULL(array))
-       return false;
-
-   nelems = ArrayGetNItems(ARR_NDIM(array), ARR_DIMS(array));
-
-   bitmap = ARR_NULLBITMAP(array);
-
-   /* check whole bytes of the bitmap byte-at-a-time */
-   while (nelems >= 8)
-   {
-        if (*bitmap != 0xFF)
-            return true;
-        bitmap++;
-        nelems -= 8;
-   }
-
-   /* check last partial byte */
-   bitmask = 1;
-   while (nelems > 0)
-   {
-       if ((*bitmap & bitmask) == 0)
-           return true;
-       bitmask <<= 1;
-       nelems--;
-   }
-
-   return false;
-}
-
-/**
  * This function checks the validity of array parameters for "sampleNewTopics".
  *
  * Parameters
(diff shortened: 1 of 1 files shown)


haradh1:
If there could be such an array such that ARR_HASNULL () is bogus, that value is the problem.  I suggest exploring the reason for that.

Looking at your change, the original code didn't use ARR_HASNULL() but did ARR_NULLBITMAP(), which I think may be bogus.  ARR_HASNULL() check should have been done before ARR_NULLBITMAP()?

renyi533:
Yes. It is not correct to use ARR_NULLBITMAP in the last commit. 

I see from the web that even ARR_HASNULL can generate wrong results.(I do not verify that ARR_HASNULL is wrong in some cases.) So I provided another function. Do you think we can totally rely on ARR_HASNULL? In this way, I can remove the function of check_array_null_elem and use ARR_HASNULL directly.

haradh1:
If ARR_HASNULL return a wrong result, the entire backend code will break, even if you fixed it by your additional function.  So yes, it should be reliable.

renyi533:
OK. I used the existing macros to detect null elements in array now. Please review again.

haradh1:
So did you verify the latest fix with ARR_HASNULL passed the case described above?

renyi533:
Yes. I verified this macro can work.

haradh1:
Hmm, looking closer, I don't see why even ARR_NULLBITMAP doesn't work well.  For example. array_send() relies on ARR_NULLBITMAP returning NULL when the array doesn't contain NULL elements.  Can you describe more exactly what happens with that data?  What is returned from ARR_NULLBITMAP?  Why the array has bigger indices than its size?  Your test of

```
testdb=# select count(*) from (select unnest(dict) from madlibtestdata.plda_nsf_abs_dict) l;
```

should return the number of elements in plda_nsf_abs_dict, not the number of non-NULL elements, because the argument of count is asterisk.  Honestly I don't understand the connection between your fix and your explanation of what you observed.

renyi533:
Hitoshi, can you turn online on the communicator? Or, can you tell me your phone number? I can call you immediately. I think this will be much more effective way.

renyi533:
ARR_NULLBITMAP does not return NULL when the array doesn't contain NULL elements. It is incorrect for us to use it here.  Please see its definition. 

ARR_HASNULL works well during my test. But I read that even this macro can generate wrong results. So I originally add another function to check. But from your point, we should rely on existing macro. So I removed the function again.

Except for the NULL element problem, another issue is invalid range of work indexes. The sql you listed is for this separate issue. We should not mix them together.

cwelton:
I'm fine with the new wording.  It sounds like we understand that the ARR_HASNULL vs ARR_NULLBITMAP change is not expected to change any behavior; in this case ARR_HASNULL is more semantically meaningful.

Looks good to pull.

haradh1:
Pulled as it stands.

PR #162, opened 2012-05-31
DT/RF: add check for feature_per_node and feature names

aojwang:
Jira: [MADLIB-520](http://jira.madlib.net/browse/MADLIB-520)
If the value of feature_per_node is greater than the total number of feature, we will raise exception. If the listed features' name are not the columns of training table, we will raise exception

### DIFF
--- a/methods/cart/src/pg_gp/dt_utility.sql_in
+++ b/methods/cart/src/pg_gp/dt_utility.sql_in
@@ -439,6 +439,86 @@ $$ LANGUAGE sql STABLE;
 
 
 /*
+ * @brief Test if each element in the given array is a column of the table.
+ *
+ * @param column_names      The array containing the columns to be tested.
+ * @param table_name        The full table name.
+ * 
+ * @return True if each element of column_names is a column of the table.
+ *
+ */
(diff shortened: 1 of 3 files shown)


methods/cart/src/pg_gp/dt_utility.sql_in
```diff
@@ -439,6 +439,86 @@ $$ LANGUAGE sql STABLE;
...
+ * 
+ * @return True each element of the first array is in the sencond array. 
+ *
+ * @note There is no order between elements of subarr. Here, we just test if an
+ *       array contains all the elements of the other array. For example, assuming 
+ *       subarr=ARRAY[1,2] and arr=ARRAY[1,3,2], then the function will return 
+ *       true.
+ *
+ */
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__array_contains
```
haradh1: Is this "<@" operator?

```
haradh1=# select array['a','b'] <@ array['a','b','c'];
 ?column? 
----------
 t
(1 row)
```

aojwang: Changed. Cool, that's really what i want. Next time, I will use/search the existing operators first. Thanks.

PR #164, opened 2012-06-04
Hypothesis tests: Fixes for MADLIB-526, MADLIB-527, MADLIB-528

fschopp:
- Welch–Satterthwaite formula gives non-integrate degress of freedom, but documentation used to say that an integral value is returned.
- Since the MW test is a rank-based test, there is no reason to disallow infinities. NaNs on the other hand now propagate. No error is raised.
- §4.15.4 ("Aggregate functions") of ISO/IEC 9075-2:2003, "SQL/Foundation" demands that aggregate functions ignore rows with NULL values. It is therefore confusing to use NULL arguments. The chi-squared GOF now has numeric default values for the expected number of observations (1) and degree of freedom (0), and the documentation explains how these values are interpreted.

### DIFF
--- a/src/ports/postgres/modules/stats/hypothesis_tests.sql_in
+++ b/src/ports/postgres/modules/stats/hypothesis_tests.sql_in
@@ -1,4 +1,4 @@
-/* ----------------------------------------------------------------------- *//** 
+/* ----------------------------------------------------------------------- *//**
  *
  * @file hypothesis_tests.sql_in
  *
@@ -116,7 +116,7 @@ LANGUAGE C IMMUTABLE STRICT;
  */
 CREATE AGGREGATE MADLIB_SCHEMA.t_test_one(
     /*+ value */ DOUBLE PRECISION) (
-    
+
     SFUNC=MADLIB_SCHEMA.t_test_one_transition,
     STYPE=DOUBLE PRECISION[],
     FINALFUNC=MADLIB_SCHEMA.t_test_one_final,
@@ -164,7 +164,7 @@ LANGUAGE C IMMUTABLE STRICT;
  *    \f]
  *    where
  *    \f[
- *        s_p^2 = \frac{\sum_{i=1}^n (x_i - \bar x)^2 
+ *        s_p^2 = \frac{\sum_{i=1}^n (x_i - \bar x)^2
  *                         + \sum_{i=1}^m (y_i - \bar y)^2}
  *                     {n + m - 2}
  *    \f]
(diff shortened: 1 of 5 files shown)


src/ports/postgres/modules/stats/hypothesis_tests.sql_in
```diff
@@ -321,7 +321,12 @@ CREATE AGGREGATE MADLIB_SCHEMA.f_test(
 );
 
 
-
+/*
+ * @brief Pearson's chi-squared goodness-of-fit test: Transition function
+ *
+ * This function is declared as CALLED ON NULL INPUT because the \c expected and
+ * \c df arguments may be NULL.
+ */
```
fschopp: Argh. I added this comment, and then made it obsolete. Of course, this needs to be removed.

cwelton:
Note that there is an exception to §4.15.4 in the SQL standard:

> NOTE 267 — Null values are not eliminated when computing <array aggregate function>. This, plus the optional <sort specification list>, sets <array aggregate function> apart from <general set function>s.

Since the SQL standard itself does not support UDAs this exception makes general assumptions about how NULLs should be handled within a UDA questionable.

fschopp:
Which version of the SQL standard did you look at? I do not find that note in SQL 2003. I suppose that is mainly referring to `array_agg`? 

On the other hand: Yes, saying the standard "demands that aggregate functions ignore rows with NULL values" is only a rough summary. The standard only demands this for the aggregate functions it explicitly lists. Still, I think we should find an API that could be a "natural" extension to the standard. And NULL arguments in the way I used them before are not a canonic extension. That's why I changed it.

fschopp:
Hmm, I looked at http://www.wiscorp.com/sql20nn.zip (from http://www.wiscorp.com/SQLStandards.html), which is the latest draft version I have seen, and it actually does not make an exception for `array_agg`. Do you have more info?

_Update_: Great, the standard (or at least the draft) contradicts itself. In the document I linked, the note is in §10.9.

> NOTE 319 — Null values are not eliminated when computing <array aggregate function>. This, plus the optional <sort specification list>, sets <array aggregate function> apart from <general set function>s.

whereas in §4.16.4 it says:

> [...] result of the aggregate function is: [...]
> - If ARRAY_AGG is specified, then an array value with one element formed from the <value expression> evaluated for each row that qualifies.

and previously:

> Of the rows in the aggregation, the following do not qualify: [...]
> - Every row in which the <value expression> evaluates to the null value.

haradh1:
Looks good to me!

fschopp:
OK, I'll merge this now.

fschopp:
Closed since commit 842f1da.

PR #165, opened 2012-06-05
Random Forest: Revert the code related with madlib 514 (Jira: MADLIB-529)

renyi533:
The code change for madlib 514 turns to be extremely inefficient for some data sets. Though the change seems fine for certain data sets, we should revert the change back based on the test results.

In this pull request, we just revert to a previous version. There is no other changes. So it is very safe to merge.


haradh1:
Do you have any sense what was wrong?

haradh1:
Ah, ok, I found it in the original jira. http://jira.madlib.net/browse/MADLIB-514?focusedCommentId=11500&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-11500

renyi533:
Thanks a lot. Hitoshi.

PR #166, opened 2012-06-06
SVEC: Add test case for "locale aware sorting in the module of svec" JIRA madlib-512

renyi533:
Resolve madlib-512.


cwelton:
Looks reasonable to me.

haradh1:
Done.

PR #167, opened 2012-06-07
SVEC: Another fix for text order in gp_sfv.

haradh1:
Jira: MADLIB-457
The previous fix was not correct in that replacing strcmp with strcol was
still not enough to match behaviors with the text sort operation in the
database backend.  There is a little hack to deal with undefined orders in
non-C collation, in which case the backend uses strcmp as the last resort.

I rewrote the existing implementation to be efficient with use of bttextcmp.
This contains a little more of cosmetic changes than addressing issue, as
well as the fix for test failure in use of the ordered aggregate in PG8.4
and GPDB4.0.

### DIFF
--- a/methods/svec/src/pg_gp/gp_sfv.c
+++ b/methods/svec/src/pg_gp/gp_sfv.c
@@ -6,22 +6,32 @@
 #include <stdlib.h>
 #include <math.h>
 
+#include "catalog/pg_type.h"
+#if PG_VERSION_NUM >= 90100
+#include "catalog/pg_collation.h"
+#endif
 #include "utils/array.h"
 #include "utils/builtins.h"
 #include "utils/lsyscache.h"
-#include "catalog/pg_type.h"
 #include "access/tupmacs.h"
 
 #include "sparse_vector.h"
 
-static char **get_text_array_contents(ArrayType *array, int *numitems);
-
-SvecType * classify_document(char **features, int num_features, 
-			     char **document, int num_words);
-
 Datum gp_extract_feature_histogram(PG_FUNCTION_ARGS);
 
-void gp_extract_feature_histogram_errout(char *msg);
+static void gp_extract_feature_histogram_errout(char *msg);
+
+static SvecType * classify_document(Datum *features, int num_features,
+				  Datum *document, int num_words, bool *null_words);
+
+#if PG_VERSION_NUM >= 90100
+#include "catalog/pg_collation.h"
+#define TextDatumCmp(x, y) (DatumGetInt32( \
+			DirectFunctionCall2Coll(bttextcmp, DEFAULT_COLLATION_OID, (x), (y))))
+#else
+#define TextDatumCmp(x, y) (DatumGetInt32( \
+			DirectFunctionCall2(bttextcmp, (x), (y))))
+#endif
 
 /**
  * 	gp_extract_feature_histogram
@@ -115,17 +125,23 @@ void gp_extract_feature_histogram_errout(char *msg);
 PG_FUNCTION_INFO_V1( gp_extract_feature_histogram );
 Datum gp_extract_feature_histogram(PG_FUNCTION_ARGS)
 {
-	SvecType *returnval;
-	char **features, **document;
-	int num_features, num_words, result;
-	ArrayType * arr0, * arr1;
+	SvecType   *returnval;
+	ArrayType  *arr0, *arr1;
+	Datum	   *features, *document;
+	int			num_features, num_words;
+	bool	   *null_words;
+	int16		elmlen;
+	bool		elmbyval;
+	char		elmalign;
+	int			i;
 
-        if (PG_ARGISNULL(0)) PG_RETURN_NULL();
+	if (PG_ARGISNULL(0))
+		PG_RETURN_NULL();
 
-        /* Error checking */
-        if (PG_NARGS() != 2) 
+	/* Error checking */
+	if (PG_NARGS() != 2) 
 		gp_extract_feature_histogram_errout(
-	          "gp_extract_feature_histogram called with wrong number of arguments");
+			"gp_extract_feature_histogram called with wrong number of arguments");
 
 	arr0 = PG_GETARG_ARRAYTYPE_P(0);
 	arr1 = PG_GETARG_ARRAYTYPE_P(1);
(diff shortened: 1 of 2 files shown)


methods/svec/src/pg_gp/gp_sfv.c
```diff
@@ -203,80 +225,26 @@ int str_bsearch(char * word, char ** features, int num_features)
...
-	for (i=0; i!=num_words; i++) {
-		if (document[i] == NULL) continue;
-		idx = str_bsearch(document[i], features, num_features);
+	for (i = 0; i < num_words; i++)
+	{
+		/* Skip if this word is NULL */
+		if (null_words[i])
+			continue;
+		idx = textdatum_bsearch(document[i], features, num_features);
+elog(INFO, "idx = %d", idx);
```
haradh1: Oops

methods/svec/src/pg_gp/gp_sfv.c
```diff
@@ -74,33 +84,19 @@ SvecType * classify_document(char **features, int num_features,
  *
  * Returns:
  * 	SFV of the document with counts of each feature, stored in a Sparse Vector (svec) datatype
- *
- * TODO:
```
haradh1, line 78: What is this TODO?

haradh1:
w=1 is suggested.

methods/svec/src/pg_gp/gp_sfv.c
```diff
@@ -139,42 +141,50 @@ Datum gp_extract_feature_histogram(PG_FUNCTION_ARGS)
...
 	ereport(ERROR,
 		(errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
 		 errmsg(
 		"%s\ngp_extract_feature_histogram internal error.",msg)));
 }
 
-int str_bsearch(char * word, char ** features, int num_features) 
+/*
+ * The reason we use this instead of standard bsearch is only historical.
+ */
```
fschopp, line 185: This comment is unclear to me. If you change the name of a function that is only there for historical reasons, why not just get rid of it?

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
```diff
@@ -47,7 +47,7 @@ CREATE TEMP TABLE dictionary_array( dict text[] )
     m4_ifdef(`GREENPLUM',`distributed randomly');
 
 insert into dictionary_array
-    select array_agg( term order by term asc ) from dictionary
+    select array(select term from dictionary order by term asc)
```
fschopp: Would you need to add a "LIMIT ALL" on Greenplum in order to have a deterministic order?

methods/svec/src/pg_gp/gp_sfv.c
```diff
@@ -6,43 +6,53 @@
...
 
-Datum gp_extract_feature_histogram(PG_FUNCTION_ARGS);
+static SvecType * classify_document(Datum *features, int num_features,
+				  Datum *document, int num_words, bool *null_words);
 
-void gp_extract_feature_histogram_errout(char *msg);
+#if PG_VERSION_NUM >= 90100
+#include "catalog/pg_collation.h"
+#define TextDatumCmp(x, y) (DatumGetInt32( \
+			DirectFunctionCall2Coll(bttextcmp, DEFAULT_COLLATION_OID, (x), (y))))
```
fschopp, line 30: This doesn't look right to me. TEXT is a collatable type, so we should use the collation associated with the argument. I'm currently trying to find out whether collations are associated with function calls or individual arguments...

methods/svec/src/pg_gp/gp_sfv.c
haradh1, line 185: Right.  I was just too lazy to do it.

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
haradh1: Really?  I think I did like this in viterbi modules, too.

haradh1: Ah, that module doesn't support 4.0.

methods/svec/src/pg_gp/gp_sfv.c
fschopp, line 30: It should be PG_GET_COLLATION(), but then you need to carry the fcinfo argument around.

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
fschopp: According to the standard LIMIT ALL should not be necessary. The <array value constructor by query> is expected to preserve order.

methods/svec/src/pg_gp/gp_sfv.c
fschopp, line 185: Prepend a "FIXME:"?

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
fschopp: Olek once reported this: "In a SELECT statement which does not pass the results directly back to the user ORDER BY clause is not enforced to run on the master."

fschopp: This is a GP backend-specific question. Hitoshi/Caleb, could you please clarify?

fschopp: This is on GP4.0:

```
schopf=# SELECT ARRAY(SELECT id FROM dummy ORDER BY id) AS not_really_sorted;
   not_really_sorted    
------------------------
 {2,4,6,8,10,1,3,5,7,9}
(1 row)
```

fschopp: whereas:

```
schopf=# SELECT ARRAY(SELECT id FROM dummy ORDER BY id LIMIT ALL) AS this_is_sorted;
     this_is_sorted     
------------------------
 {1,2,3,4,5,6,7,8,9,10}
(1 row)
```

fschopp:
Two main issues:
- Add support for PG>=9.1 collations, `DEFAULT_COLLATION_OID` is not enough
- On Greenplum, `SELECT ARRAY(SELECT x FROM source ORDER BY x)` is **not** equivalent to `SELECT array_agg(x ORDER BY x) FROM source`. GP deviates from the standard here, and in order to enforce standard behavior, adding `LIMIT ALL` is necessary.

methods/svec/src/pg_gp/gp_sfv.c
haradh1, line 30: Yeah, I think it's our option to support COLLATE in this function and I don't like it.  As it's been not aware of it, I think it's not urgent.

methods/svec/src/pg_gp/gp_sfv.c
haradh1, line 185: Working on this.

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
haradh1: I mean, I actually fixed this issue in 4.2.  But 4.0 and 4.1 should need LIMIT ALL.

haradh1:
Thanks Florian,

I don't think we should support COLLATE at this moment as it's too complicated.
Fixed LIMIT ALL and bsearch issue.

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
fschopp: You mean GP4.2 is now SQL standard compliant here?

methods/svec/src/pg_gp/gp_sfv.c
fschopp, line 30: It's fairly easy to add proper collation support, so I think we should do it. Otherwise, the same issue remains to exist, even if its surface may have reduced.

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
```diff
@@ -47,7 +47,7 @@ CREATE TEMP TABLE dictionary_array( dict text[] )
     m4_ifdef(`GREENPLUM',`distributed randomly');
 
 insert into dictionary_array
-    select array_agg( term order by term asc ) from dictionary
+    select array(select term from dictionary order by term asc limit all)
```
cwelton, line 50: Is there a reason you prefer array constructors over array_agg?  I don't think it makes much difference, just curious.

haradh1, line 50: ordered aggregate failed in 8.4 and 4.0.

methods/svec/src/pg_gp/gp_sfv.c
cwelton, line 30: I tend to agree.  The old code was using C sort order, which is what we are fixing.  If we are adding proper collation support then we should presumably handle this case, otherwise we aren't fully addressing the reported bug due to inconsistencies between db sort order and lda sort order.

methods/svec/src/pg_gp/gp_sfv.c
cwelton, line 78: I believe this is a left over todo based on when this code was using a hash table instead of a sorted string.  On the surface this is a good thing, but the problem was that it was a global hash table and the assumptions about the hash table's current state were invalid, which is why keesiong rewrote it to the current form.

methods/svec/src/pg_gp/sql/gp_sfv_sort_order.sql_in
cwelton, line 50: Ah, yes of course.  Makes sense.

methods/svec/src/pg_gp/gp_sfv.c
haradh1, line 30: What would be the "collation mismatch" situation??  I guess it's like

```
insert into dictionary_array
select array(select term from dictionary order by term collate "fr_FR");

select
  madlib.svec_sfv( dict.dict, ARRAY['0']) sfv
from
  ( select dict from dictionary_array limit 1 ) dict                          ;
```

This case svec_sfv should agree the collation of fr_FR?  Is the information carried between INSERT and second SELECT to svec_sfv?  I'm not sure right now.  I need to know more about it.  That's why I'm thinking it's not the best time to do it.  I'm not disagreeing to support this ever.  I just say it's not the right time to do it.

fschopp, line 30: I think http://www.postgresql.org/docs/9.1/static/collation.html explains this. You explicitly assigned collation "fr_FR" to dict.dict, so this is what will dominate and be passed to svec_sfv. If you also explicitly assigned a collation to ARRAY['0'] you would see an error.

So, all function arguments share the same collation, because the parser will error out in case of inconsistencies.

cwelton, line 30: After discussion I think I'm okay here.  My understanding was that DEFAULT_COLLATION_OID was going to be a _specific_ collation, not the collation defined for the database.  So long as the collation matches the default collation for the database I don't see an issue here so long as we document that this is the behavior.  

Which I guess means that we need to add a doxygen comment to this affect.

fschopp, line 30: I agree that it's not worth spending much time on this right now, but as I said, proper collation support is easy. The only "difficulty" is passing the result of PG_GET_COLLATION() into the comparison function.

renyi533:
I am a little confused. When strcoll returns 0, it may mean true equal or undefined sort order. Is this right?

haradh1:
Honestly, I don't know.  The requirement is to make the behavior identical to the one in the backend.

PR #168, opened 2012-06-08
Probability functions: Improved handling of invalid arguments

fschopp:
Jira: MADLIB-538, MADLIB-541, MADLIB-542, MADLIB-543, MADLIB-548

Fixes:
- Chi-Squared GOF test: return NULL for p-values if d.f. <= 0 instead of raising an error (MADLIB-538)
- "parameter x is not accepting 0 in pareto cdf and pdf" (MADLIB-541)
- "parameter x is not accepting positive and negative infinities in triangular/uniform cdf/pdf" (MADLIB-542)
- "parameter x is not accepting 0 in pareto cdf and pdf" (MADLIB-541)
- "fisher f quantile with NaN as df2 should return NaN, but it raises internal error" (MADLIB-543)
- "rayleigh cdf/pdf are accepting positive infinity as their parameter scale while they should not" (MADLIB-548)

### DIFF
--- a/src/modules/prob/boost.hpp
+++ b/src/modules/prob/boost.hpp
@@ -519,6 +519,12 @@ struct DomainCheck<boost::math::fisher_f_distribution<RealType, Policy> >
         static const char* function = "madlib::modules::prob::<unnamed>::"
             "DomainCheck<fisher_f_distribution<%1%> >::quantile(...)";
 
+        if (!boost::math::detail::check_df(function,
+            inDist.degrees_of_freedom1(), &outResult, Policy())
+            || !boost::math::detail::check_df(function,
+                inDist.degrees_of_freedom2(), &outResult, Policy()))
+            return kResultIsReady;
+
         if (std::isnan(inP)) {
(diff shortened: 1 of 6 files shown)


src/modules/prob/boost.hpp
```diff
@@ -865,6 +871,37 @@ struct DomainCheck<boost::math::pareto_distribution<RealType, Policy> >
...
+            return kResultIsReady;
+        }
+        return Base::template cdf<Complement>(inDist, inX, outResult);
+    }
+
+    static ProbFnOverride pdf(const Distribution& inDist,
+        const RealType& inX, RealType& outResult) {
+
+        static const char* function = "madlib::modules::prob::<unnamed>::"
+            "DomainCheck<pareto_distribution<%1%> >::cdf(...)";
```
fschopp: Ah, this should be pdf.

haradh1:
Looks OK to me.

fschopp:
OK, will merge soon.

PR #169, opened 2012-06-08
Fixes in probability module

fschopp:
**Contains the following changes:**

Wilcoxon-Signed-Rank test: Changes semantics for precision argument, Documentation update

Jira: MADLIB-405, MADLIB-534, MADLIB-553

Probability functions:
- Hollander/Wolfe ("Nonparametric statistical methods", 2nd ed., p. 38, 1999) confirm the formula that we have been using. The warning about conflicting source has been deleted. (MADLIB-405)
- No longer treat NULL as special value for precision argument (MADLIB-534)
- Negative precision will now result in using the default precision, non-finite precision will be rejected (MADLIB-553)

SQL2C++ converter:
- Support for unary expressions (needed for negative default arguments)

Probability functions (Extreme-value/Exponential): Fixed handling of non-finite parameters

Jira: MADLIB-555, MADLIB-556

These are boost "bugs". Implemented fixes for the cases where boost does not catch non-finite parameters.

Probability functions: Documentation update

Jira: MADLIB-546

Fixes:
- Noncentrality parameter for Chi-squared may also be 0

Kolmogorov-Smirnov test: Fixed wrong calculation of test statistic

Jira: MADLIB-554

Fixed the following issue:
Whenever, we saw a new row, we computed the distance of the empirical distribution functions at the current position as

```
double diff = std::fabs(state.num(0) / state.expectedNum(0)
                - state.num(1) / state.expectedNum(1));
```

This could give wrong results in case of ties. After all, we increased state.num(0) and state.num(1) one after the other, so diff could temporarily be larger than it would be at the end of a group of ties.

One-way ANOVA: return NULL for p-values if d.f. <= 0 instead of raising error

Jira: MADLIB-552

### DIFF
--- a/src/modules/stats/kolmogorov_smirnov_test.cpp
+++ b/src/modules/stats/kolmogorov_smirnov_test.cpp
@@ -31,12 +31,10 @@ public:
     KSTestTransitionState(const AnyType &inArray)
       : mStorage(inArray.getAs<Handle>()),
         num(&mStorage[0], 2),
-        last(&mStorage[2]),
-        maxDiff(&mStorage[3]),
-        expectedNum(&mStorage[4], 2) {
-        madlib_assert(mStorage.size() >= 6, std::runtime_error(
-            "Out-of-bounds array access detected."));
-    }
+        expectedNum(&mStorage[2], 2),
+        last(&mStorage[4]),
+        maxDiff(&mStorage[5]),
+        lastDiff(&mStorage[6]) { }
 
     inline operator AnyType() const {
         return mStorage;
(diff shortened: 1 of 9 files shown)


src/modules/stats/kolmogorov_smirnov_test.cpp
```diff
@@ -31,12 +31,10 @@ class KSTestTransitionState {
     KSTestTransitionState(const AnyType &inArray)
       : mStorage(inArray.getAs<Handle>()),
         num(&mStorage[0], 2),
-        last(&mStorage[2]),
-        maxDiff(&mStorage[3]),
-        expectedNum(&mStorage[4], 2) {
-        madlib_assert(mStorage.size() >= 6, std::runtime_error(
```
haradh1, line 37: Why is it not necessary now?

fschopp, line 37: ArrayHandle::operator[]() does bounds checking. Previously, we were also using mStorage[5], but only calling operator[](4).

haradh1:
OK, I think it's done.

PR #170, opened 2012-06-12
Bugfix: MADLIB-556, MADLIB-545, MADLIB-544

KevinYL:
Bugfix: MADLIB-556, MADLIB-545, MADLIB-544.
Overriding existing domain check.


fschopp:
Included as of commit 0b445cc.

PR #171, opened 2012-06-12
Probability functions: Improved corner-case handling for [negative_]binomial functions

fschopp:
Jira: MADLIB-537, MADLIB-544

Fixes:
- binomial and negative_binomial: Corner cases with success probability 0 or 1 are overridden, instead of being wrongly handled by boost
- Disallow success probability 0 for negative binomial
- Documentation: inverse_gamma_quantile() now accepts probability 0 (MADLIB-544)
- New unit tests for the above


haradh1:
I think it's ok.

PR #173, opened 2012-06-14
Documentation

fschopp:
Documentation: Overview of Probability Module

Jira: MADLIB-570

Documentation: Overview of Hypothesis Tests Module

Jira: MADLIB-571

### DIFF
--- a/src/ports/postgres/modules/prob/prob.sql_in
+++ b/src/ports/postgres/modules/prob/prob.sql_in
@@ -9,12 +9,97 @@
  *
  *//* ----------------------------------------------------------------------- */
 
+m4_include(`SQLCommon.m4')
+
 /**
 @addtogroup grp_prob
 
 @about
 
-Probability-distribution functions.
+The Probability Functions module provides cumulative distribution, density/mass,
+and quantile functions for a wide range of probability distributions.
+
+Unless otherwise documented, all of these functions are wrappers around
+functionality provided by the boost C++ library [1, “<a href=
+"http://www.boost.org/doc/libs/1_49_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist.html"
+>Statistical Distributions and Functions</a>”].
+
+For convenience, all cumulative distribution and density/mass functions (CDFs
+and PDF/PMFs in short) are defined over the range of all floating-point numbers
+including infinity. Inputs that are \c NULL or \c NaN (not a number) will always
+produce a \c NULL or \c NaN result, respectively. Inputs that are plus or minus
+infinity will return the respective limits.
+
+A quantile function for a probability distrution with CDF \f$ F \f$ takes a
+probability argument \f$ p \in [0,1] \f$ and returns the value \f$ x \f$ so that
+\f$ F(x) = p \f$, provided such an \f$ x \f$ exists and it is unique. If it does
+not, the result will be
+\f$
+    \sup \{ x \in D \mid F(x) \leq p \}
+\f$
+(interpreted as 0 if the supremum is over an empty set) if \f$ p < 0.5 \f$, and
+\f$
+    \inf \{ x \in D \mid F(x) \geq p \}
+\f$
+if \f$ p \geq 0.5 \f$. Here \f$ D \f$ denotes the domain of the distribution,
+which is the set of reals \f$ \mathbb R \f$ for continuous and the set of
+nonnegative integers \f$ \mathbb N_0 \f$ for discrete distributions.
+
+Intuitively, the formulas in the previous paragraph deal with the following
+special cases. The 0-quantile will always be the “left end” of the support,
+and the 1-quantile will be the “right end” of the support of the distribution.
+For discrete distributions, most values of \f$ p \in [0,1] \f$ do not admit an
+\f$ x \f$ with \f$ F(x) = p \f$. Instead, there is an \f$ x \in \mathbb N_0 \f$
+so that \f$ F(x) < p < F(x + 1) \f$. The above formulas mean that the
+value returned as \f$ p \f$-quantile is \f$ x \f$ if \f$ p < 0.5 \f$, and it
+is \f$ x + 1 \f$ if \f$ p \geq 0.5 \f$. (As a special case, in order to ensure
+that quantiles are always within the support, the \f$ p \f$-quantile will be 0
+if \f$ p < F(0) \f$).
+
+The rationale for choosing this behavior is that \f$p\f$-quantiles for
+\f$ p < 0.5 \f$ are typically requested when interested in the value
+\f$ x \f$ such that with confidence level <strong>at least</strong>
+\f$ 1 - p \f$ a random variable will be \f$ > x \f$ (or equivalently, with
+probability <strong>at most</strong> \f$ p \f$, it will be \f$ \leq x \f$).
+Likewise, \f$p\f$-quantiles for \f$ p \geq 0.5 \f$ are typically requested when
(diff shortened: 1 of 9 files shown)


src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -14,7 +14,84 @@
 
 @about
 
-Probability-distribution functions.
+The Probability-functions module provides cumulative distribution, density/mass,
```
haradh1: Is "Probability-functions" a common name?  Other modules are named without hyphen.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -14,7 +14,84 @@
...
+and quantile functions for a wide range of probability distributions.
+
+Unless otherwise documented, all of these functions are wrappers around
+functionality provided by the boost C++ library [1, “<a href=
+"http://www.boost.org/doc/libs/1_49_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist.html"
+>Statistical Distributions and Functions</a>”].
+
+For convenience, all cumulative distribution and density/mass functions (CDFs
```
haradh1, line 27: What about quantile functions?  Is there general rule for invalid inputs for them?

haradh1:
Done.

src/ports/postgres/modules/prob/prob.sql_in
fschopp: http://en.wikipedia.org/wiki/English_compound#Hyphenated_compound_modifiers

But I should probably capitalize Functions.

src/ports/postgres/modules/prob/prob.sql_in
fschopp, line 27: Yes, probabilities outside of [0,1] will raise an error. Is it worth mentioning?

cwelton:
Looks reasonable to me.

src/ports/postgres/modules/prob/prob.sql_in
haradh1, line 27: As a novice, I wondered.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -14,7 +14,84 @@
...
+\f$
+if \f$ p \leq 0.5 \f$ and
+\f$
+    \inf \{x \in D \mid F(x) \geq p \}
+\f$
+if \f$ p > 0.5 \f$. Here we denote by \f$ D \f$ the domain of the distribution,
+which is the set of reals \f$ \mathbb R \f$ for continuous and the set of
+integers \f$ \mathbb Z \f$ for discrete distributions.
+
+Intuitively, the 0-quantile will always be the “left end” of the support
```
haying: Is the only functionality of this paragraph to further explain the previous paragraph? Perhaps "Intuitively" -> "For special/trivial cases"

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -14,7 +14,84 @@
...
+and the 1-quantile will be the “right end” of the support of the distribution.
+For discrete distributions, most values of \f$ p \in [0,1] \f$ do not admit an
+\f$ x \f$ with \f$ F(x) = p \f$. Instead, there is an \f$ x \in \mathbb Z \f$ so
+that \f$ F(x) < p \f$ and \f$ F(x + 1) > p \f$. The above formula means that the
```
haying: Not sure whether this is just a personal preference: I like \f$ F(x) < p < F(x+1) \f$ better

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -14,7 +14,84 @@
...
+\f$ p > 0.5 \f$ are typically requested when interested in the value \f$ x \f$
+such that with confidence level <strong>at least</strong> \f$ p \f$ a random
+variable will be \f$ \leq x \f$. See also [1, “<a href=
+"http://www.boost.org/doc/libs/1_46_1/libs/math/doc/sf_and_dist/html/math_toolkit/policy/pol_tutorial/understand_dis_quant.html"
+>Understanding Quantiles of Discrete Distributions</a>”].
+
+@usage
+
+- Cumulative distribution functions:
+  <pre>SELECT <em>distribution</em>_cdf(<em>random variate</em>[, <em>parameter1</em> [, <em>parameter2</em> [, <em>parameter3</em>] ] ])</pre>
```
haying, line 74: Is <em> the convention through all MADlib docs for variable portion in function signatures? Otherwise, it does not seem to stand out immediately to allow new users to understand they can replace "distribution" by "student_t" or "uniform".

Also, is it the case that parameter1, parameter2 and parameter3 have (1) different meanings according to the "distribution" that the user calls, or (2) same for any distributions? If (1), perhaps a link to more actually function signatures should be put in this @usage section (the one in @sa?); if (2), it might be better to change the parameter names to be more informative.

src/ports/postgres/modules/prob/prob.sql_in
fschopp: I totally agree...

src/ports/postgres/modules/prob/prob.sql_in
fschopp: Yes, this paragraph gives an intuitive explanation of the formal definition. This statement ("the 0-quantile will be ...") actually applies to all quantile functions in MADlib, not just for special cases.

src/ports/postgres/modules/prob/prob.sql_in
fschopp, line 74: We have as convention that italics indicate placeholders. Still, it might be helpful to add that <em>distribution</em> stands for beta, binomial, ...

As fpr parameter1,...,3: That's a somewhat informal usage description. The distribution functions have different numbers of parameters, ranging from 0 to 3. But then, I don't want to list all distributions here, because that kind of prevents seeing the forest for the trees.

haying, line 74: How about just a sentence with a link? like:
For detailed function signatures, see [here].

This can at least tell readers that there are placeholders in these signature-like descriptions.

src/ports/postgres/modules/prob/prob.sql_in
```diff
@@ -14,7 +14,84 @@
...
+infinity will return the respective limits.
+
+A quantile function for a probability distrution with CDF \f$ F \f$ takes a
+probability argument \f$ p \in [0,1] \f$ and returns the value \f$ x \f$ so that
+\f$ F(x) = p \f$, provided such an \f$ x \f$ exists and it is unique. If it does
+not, the result will be
+\f$
+    \sup \{x \in D \mid F(x) \leq p \}
+\f$
+if \f$ p \leq 0.5 \f$ and
```
fschopp: Actually, I find:

```
select poisson_cdf(0, 1), poisson_quantile(0.25, 1);
    poisson_cdf    | poisson_quantile 
-------------------+------------------
 0.367879441171442 |                0
(1 row)
```

So, boost just returns the lower bound of the support. However, this is actually rounding up, like in the case where p >= 0.5. A bit annoying, as it makes the description more complicated.

fschopp: Thinking about it again, I find this is a deficiency in boost. It simply violates the boost documentation: ["This gives no more than the requested probability in the tails"](http://www.boost.org/doc/libs/1_46_1/libs/math/doc/sf_and_dist/html/math_toolkit/policy/pol_tutorial/understand_dis_quant.html)

PR #175, opened 2012-06-29
Generic Convex Module: matrix factorization; a marker "HAYING:" is inserted to places that I feel unsatisfying

haying:
JIRA: MADLIB-578

First model - low-rank matrix factorization, and
First algorithm - incremental gradient


haying:
100K lines are a data set for test, LOC < 1000.

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
+/* ----------------------------------------------------------------------- *//** 
+ *
+ * @file igd.hpp
+ *
+ *//* ----------------------------------------------------------------------- */
```
cwelton: I would prefer to see more comments in the code overall.  A file like this is small enough that it can be understood pretty quickly, but good to have an initial description as well to provide some clarity.  Especially when this file is so similar to loss.hpp, it would be good to more quickly understand the purpose of the file and where it fits in overall.

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
...
+#ifndef MADLIB_CONVEX_ALGO_IGD_HPP_
+#define MADLIB_CONVEX_ALGO_IGD_HPP_
+
+namespace madlib {
+
+namespace modules {
+
+namespace convex {
+
+// The reason for using ConstState instead of const State: in loss.hpp
```
cwelton: Personally I prefer to have the explanation for the reason here as well rather than just a pointer to more source code.   Especially if its only a three line explanation.

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
...
+    static void merge(state_type &state, const_state_type &otherState);
+    static void final(state_type &state);
+};
+
+template <class State, class ConstState, class Task>
+void
+IGD<State, ConstState, Task>::transition(state_type &state,
+        const tuple_type &tuple) {
+    // The reason for update model inside a Task:: function instead of
+    // returning the gradient and do it here: the gradient is a sparsere
```
cwelton: s/sparsere/sparser/

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
...
+IGD<State, ConstState, Task>::merge(state_type &state,
+        const_state_type &otherState) {
+    // The reason of this weird algorithm instead of an intuitive one
+    // -- (w1 * m1 + w2 * m2) / (w1 + w2): we have only one mutable state,
+    // therefore, (m1 / w2  + m2)  * w2 / (w1 + w2).
+    // Order:     11111111  22222  3333333333333333
+
+    // model averaging, weighted by rows seen
+    double totalNumRows = state.algo.numRows + otherState.algo.numRows;
+    state.algo.incrModel *= (1. * state.algo.numRows) / otherState.algo.numRows;
```
cwelton: What if other state saw no rows?  => div by zero

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
...
+    state.algo.incrModel += otherState.algo.incrModel;
```
cwelton, line 76: Having gone as far as creating "+=(LMFModel)", why doesn't that operation handle the weighting based on numRows?

src/modules/convex/algo/loss.hpp
```diff
@@ -0,0 +1,65 @@
...
+    static void transition(state_type &state, const tuple_type &tuple);
+    static void merge(state_type &state, const_state_type &otherState);
+    static void final(state_type &state);
+};
+
+template <class State, class ConstState, class Task>
+void
+Loss<State, ConstState, Task>::transition(state_type &state,
+        const tuple_type &tuple) {
+    // "task.model" NOT "algo.incrModel"! Also see: igd.hpp:final()
```
cwelton, line 40: Having read through igd.hpp and here I still don't quite understand the distinction here.  Hopefully will become clear after reading further.

src/modules/convex/task/lmf.hpp
```diff
@@ -0,0 +1,105 @@
+/* ----------------------------------------------------------------------- *//**
+ *
+ * @file lmf.hpp
```
cwelton, line 3: Can we distinguish convex/lmf.hpp and task/lmf.hpp a little more in filename?  Whenever possible I prefer keeping unique filenames in a project to reduce confusion.

src/modules/convex/algo/igd.hpp
haying: For fixing a typo like this, how should it be handled? Should I make one more commit and push it here or?

src/modules/convex/algo/igd.hpp
haying: Currently this is handled from the caller, which is lmf_igd_merge::run() here.

src/modules/convex/algo/igd.hpp
haying, line 76: This will require numRows to be a member field in class LMFModel, which is not currently the case. But +=(State::Model) or LMFModel::some_interface_for_model_averaging is necessary, because we want to have this to be generic across LMF, Logit, SVM, etc.

src/modules/convex/algo/loss.hpp
haying, line 40: First, think about why the value of task.model is different from algo.incrModel: the former one is the model resulted from last iteration, the latter is the model in the middle of this iteration, which is subject to change in this transition function within the same iteration but different tuples. The resulting loss for the two cases: loss = f(m_k_0, t1) + f(m_k_0, t2) + ... + f(m_k_0, t_N) = \sum_{i=1}^{N} f(m_k_0, t_i); loss = f(m_k_1, t_1) + f(m_k_2, t_2) + ... + f(m_k_N, t_N). m_k_i is the resulting model at the i-th tuple of the k-th iteration, t_i is the i-th tuple, f(m, t) is the objective function we are minimizing. N is the total number of tuples. I decided the former one is more desirable...

src/modules/convex/task/lmf.hpp
haying, line 3: I agree. But I did not want to make this decision on my own. This also relates to how we want to structure the source code in files once we have many tasks and many algo -- are we still keeping the same convention as in the regress module, where we kept all algos for the same task in a single cpp & hpp file.

src/modules/convex/algo/igd.hpp
haying: Sure.

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
+/* ----------------------------------------------------------------------- *//** 
+ *
+ * @file igd.hpp
+ *
+ *//* ----------------------------------------------------------------------- */
+
+#ifndef MADLIB_CONVEX_ALGO_IGD_HPP_
```
haradh1: Reading existing hpp files, the convention seems like MADLIB_MODULES_CONVEX_ALGO_IGD_HPP

src/modules/convex/lmf.cpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // tuple
+    LMFTuple tuple;
+    uint32_t i = args[1].getAs<int32_t>();
+    uint32_t j = args[2].getAs<int32_t>();
+    tuple.depVar = args[3].getAs<double>();
+    if (i > std::numeric_limits<uint16_t>::max() ||
+            j > std::numeric_limits<uint16_t>::max()) {
+        throw std::domain_error("Indices cannot be larger than 65535.");
```
haradh1: Negative value check, too?  What about NaN?

src/modules/convex/type/model.hpp
```diff
@@ -0,0 +1,116 @@
...
+    //void rebind(const Handle &inHandle, uint16_t inRowDim, uint16_t inColDim,
+    //        uint16_t inMaxRank) {
+    //    matrixU.rebind(&inHandle[0], inRowDim, inMaxRank);
+    //    matrixV.rebind(&inHandle[inRowDim * inMaxRank], inColDim, inMaxRank);
+    //}
+
+    /**
+     * @brief Initialize the model randomly with a user-provided scale factor
+     */
+    void initialize(const double &inInitValue) {
```
haradh1: The variable name doesn't make sense to me.  Isn't it simply "scaleFactor" or something?

src/modules/convex/type/state.hpp
```diff
@@ -0,0 +1,185 @@
...
+     * something for the convex programming; therefore, it looks a bit weird to
+     * me. But I am not sure where I can move this to...
+     */
+    inline void computeRMSE() {
+        task.RMSE = sqrt(algo.loss / algo.numRows);
+    }
+
+    static inline uint32_t arraySize(const uint16_t inRowDim, 
+            const uint16_t inColDim, const uint16_t inMaxRank) {
+        return 8 + 2 * LMFModel<Handle>::arraySize(inRowDim, inColDim, inMaxRank);
```
haradh1, line 109: Where does this 8 come from?

src/modules/convex/type/tuple.hpp
```diff
@@ -0,0 +1,53 @@
+/* ----------------------------------------------------------------------- *//**
+ *
+ * @file tuple.hpp
```
haradh1, line 3: What is it for?  It seems no one is using it?

src/ports/postgres/modules/convex/lmf.sql_in
```diff
@@ -0,0 +1,300 @@
...
+--    ELSE
+--        EXECUTE '
+--        CREATE TABLE ' || output_table || ' (
+--            id                  serial,
+--            flatten_matrix_u    DOUBLE PRECISION[],
+--            flatten_matrix_v    DOUBLE PRECISION[],
+--            rmse                DOUBLE PRECISION)';
+--    END IF;
+    BEGIN
+        EXECUTE 'SELECT 1 FROM ' || output_table || ' LIMIT 0';
```
haradh1: Can we perform DROP TABLE IF EXISTS & CREATE TABLE here?

src/modules/convex/type/state.hpp
haying, line 109: This follows the practice of using C++ abstraction layer, see the comment for the LMFModel::rebind() function, see also the same use from the regress module: https://github.com/madlib/madlib/blob/master/src/modules/regress/logistic.cpp#L127

src/modules/convex/type/model.hpp
haying: Yes, we may change it. It is a scale factor for random initial values...

src/modules/convex/lmf.cpp
haying: I think this should probably be checked within the function getAs(), although I am not super clear whether it is or not now...

src/ports/postgres/modules/convex/lmf.sql_in
```diff
@@ -0,0 +1,300 @@
...
+        -- update by calling the aggregate function
+        EXECUTE 'INSERT INTO _madlib_lmf_igd 
+            SELECT ' || iteration || ' AS iter,
+                MADLIB_SCHEMA.lmf_igd_step('
+                    || row_name || ', '
+                    || col_name || ', '
+                    || value_name || ', 
+                    (SELECT _madlib_state
+                    FROM _madlib_lmf_igd
+                    WHERE _madlib_iteration = ' || iteration - 1 || '), '
```
haradh1: It looks to me this subquery returns NULL for the first step.  Is it expected?

haying: Yes, exactly. We are checking NULL in the transition function, if null, we perform some initialization. https://github.com/madlib/madlib/pull/175/files#L4R49

src/ports/postgres/modules/convex/lmf.sql_in
haying: I think we may not want to drop a user table here. If the user is willing to drop a table, he should have this done explicitly. We should move this discussion to the mailing list. http://madlib.net/pipermail/devel/2012-July/thread.html

src/modules/convex/lmf.cpp
```diff
@@ -0,0 +1,178 @@
...
+ */
+AnyType
+lmf_igd_transition::run(AnyType &args) {
+    // The real state. 
+    // For the first tuple: args[0] is nothing more than a marker that 
+    // indicates that we should do some initial operations.
+    // For other tuples: args[0] holds the computation state until last tuple
+    LMFIGDState<MutableArrayHandle<double> > state = args[0];
+
+    // initilize the state if first tuple
```
haradh1: typo: initialize

src/modules/convex/type/tuple.hpp
haying, line 3: LMFTuple is used here: https://github.com/madlib/madlib/pull/175/files#L4R28

src/modules/convex/algo/igd.hpp
haying: Agree.

src/modules/convex/lmf.cpp
haying: @fschopp How do you think we should use this getAs<uint16_t>() for now?

haradh1: getAs<int32_t>() sounds like something that checks (signed) integer, but I'm not sure, either.

src/modules/convex/type/tuple.hpp
haradh1, line 3: Ah, I missed typedef below...

haradh1:
Assign Florian to review and pull this.

src/modules/convex/algo/igd.hpp
cwelton: For a simple typo we could theoretically fix in merge.  I'm inclined to say here that there will be at least one followup commit to account for some other review feedback and the change can be incorporated into that.

src/modules/convex/algo/igd.hpp
cwelton: In which case I'd like to see some comment here explaining the expectations placed upon a caller of this function.

src/modules/convex/algo/igd.hpp
cwelton, line 76: Ah, I see.  Makes sense.

src/modules/convex/lmf.cpp
```diff
@@ -0,0 +1,178 @@
...
+            // configuration parameters
+            // HAYING: all the checking of numeric_limis should be moved to
+            // the implementation of getAs()
+            uint32_t m = args[5].getAs<int32_t>();
+            uint32_t n = args[6].getAs<int32_t>();
+            uint32_t r = args[7].getAs<int32_t>();
+            if (m > std::numeric_limits<uint16_t>::max() ||
+                    n > std::numeric_limits<uint16_t>::max() ||
+                    r > std::numeric_limits<uint16_t>::max()) {
+                throw std::domain_error("m, n, r cannot be larger than 65535");
```
cwelton: lmf_igd_transition is defined in the sql_in file as:

```
CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.lmf_igd_transition(
        state           DOUBLE PRECISION[], 
        row_num         INTEGER,    
        col_num         INTEGER,
        val             DOUBLE PRECISION,
        previous_state  DOUBLE PRECISION[],
        row_dim         INTEGER,
        col_dim         INTEGER,
        max_rank        INTEGER,
        stepsize        DOUBLE PRECISION,
        init_value      DOUBLE PRECISION)
...
```

Where `m,n,r` apparently map to row_dim, col_dim, max_rank.  But then the transition function is intended to be internal and only called via the aggregate, which is also internal.

Can we have this error message actually given feedback to the user in terms that they can recognize and act on?

src/modules/convex/lmf.cpp
```diff
@@ -0,0 +1,178 @@
...
+    // We request a mutable object. Depending on the backend, this might perform
+    // a deep copy.
+    LMFIGDState<MutableArrayHandle<double> > state = args[0];
+
+    // Aggregates that haven't seen any data just return Null.
+    if (state.algo.numRows == 0) { return Null(); }
+
+    // finalizing
+    LMFIGDAlgorithm::final(state);
+    // LMFLossAlgorithm::final(state); // empty function call causes a warning
```
cwelton: My c++ is rusty, but it seems to me having empty functions should be acceptable and from an interface point of view it is better to call them than leave them commented out so that the implementation may do something in the future.

Florian, do you have a recommendation here?

src/modules/convex/task/lmf.hpp
cwelton, line 3: Open for discussion.  Maybe for now we have `convex/lmf.hpp` and `task/lmf_task.hpp`.  I'm fine with alternatives, but my experience working on other projects is that having multiple files name `xxx.yyy` only leads to confusion.  Any differentiation we can add here is good.

src/modules/convex/type/model.hpp
```diff
@@ -0,0 +1,116 @@
...
+     * Before modifying HandleTraits.hpp, we have to rebind the matrices
+     * in a higher level.
+     *
+     * @brief Rebind to a provided double array.
+     */
+    //void rebind(const Handle &inHandle, uint16_t inRowDim, uint16_t inColDim,
+    //        uint16_t inMaxRank) {
+    //    matrixU.rebind(&inHandle[0], inRowDim, inMaxRank);
+    //    matrixV.rebind(&inHandle[inRowDim * inMaxRank], inColDim, inMaxRank);
+    //}
```
cwelton: Fix or remove, commented out code is undesirable.

src/modules/convex/type/model.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * @brief Initialize the model randomly with a user-provided scale factor
+     */
+    void initialize(const double &inInitValue) {
+        srand(static_cast<unsigned>(time(NULL)));
```
cwelton: I'd prefer us not to go initializing random seeds, the database may care about it for some reason.  An improvement here would be to use the Databases RNG, perhaps expose this through the DB abstraction layer?

src/modules/convex/lmf.cpp
haying: I would make some ad hoc adjustments here to resolve this here, although I am wondering whether we'd like to place some naming restrictions/conventions to avoid these mistakes in the future.

src/config/Modules.yml
```diff
@@ -9,6 +9,7 @@ modules:
     - name: compatibility
     - name: conjugate_gradient
       depends: ['array_ops']
+    - name: convex
```
fschopp: Probably fine for now, but eventually I would expect convex optimization to be completely subsumed by the abstraction layer. In that case, it will no longer be a 'module' but just an operation provided by the AL.

src/modules/convex/algo/igd.hpp
```diff
@@ -0,0 +1,80 @@
...
+            state.task.stepsize);
+}
+
+template <class State, class ConstState, class Task>
+void
+IGD<State, ConstState, Task>::merge(state_type &state,
+        const_state_type &otherState) {
+    // The reason of this weird algorithm instead of an intuitive one
+    // -- (w1 * m1 + w2 * m2) / (w1 + w2): we have only one mutable state,
+    // therefore, (m1 / w2  + m2)  * w2 / (w1 + w2).
```
fschopp: Looks like a typo: You want to do:
`(m1 * w1 / w2  + m2)  * w2 / (w1 + w2)`

src/modules/convex/lmf.cpp
fschopp: I advise against `uint32_t i = args[1].getAs<int32_t>();`. This indeed gives the possibility that the user may specify negative values. (Actually, since I have increased the warning level, this should also produce a warning with gcc!) If a value is NULL, `getAs()` will raise an exception.

fschopp: For limiting the range to [0,2^15-1], use `getAs<uint16_t>()`. At the moment that requires the SQL type to be INT2 (i.e., may be less convenient to use), but I may change this in the abstraction layer that also INT4 or INT8 are accepted, but that these type just raise an exception if the value is too large to fit into the template argument given to `getAs`.

src/modules/convex/lmf.cpp
```diff
@@ -0,0 +1,178 @@
...
+AnyType
+internal_lmf_igd_result::run(AnyType &args) {
+    LMFIGDState<ArrayHandle<double> > state = args[0];
+
+    AnyType tuple;
+    tuple << state.task.model.matrixU
+        << state.task.model.matrixV
+        << static_cast<double>(state.task.RMSE);
+    // HAYING: I don't know why this casting is necessary, again, type
+    // checking thing...
```
fschopp: It's due to the way AnyType objects can currently be constructed. Currently (for the PostgreSQL port) there is one template constructor, and it will eventually be missing an implementation for `TypeTraits<Reference<...> >`. One could improve by using SFINAE (i.e., prevent type `Reference<T>` from being instantiated and force a conversion to `T` instead) or changes in the design, but so far I have not given it priority.

src/modules/convex/type/model.hpp
fschopp: Yes, a boost-style frontend to the database RNG is desirable.

src/modules/convex/type/state.hpp
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * Intra-iteration components (updated in transition step):
+     *   modelLength = (rowDim + colDim) * maxRank
+     * - 6 + modelLength: numRows (number of rows processed in this iteration)
+     * - 7 + modelLength: loss (sum of squared errors)
+     * - 8 + modelLength: incrModel (volatile model for incrementally update)
+     *
+     * HAYING: This function will currently causes a invalid memory access if
+     * an 8-length zero array is assigned to mStorage. The rebind of the last 
+     * two matrices seems to be touching the slot next to the last element
```
fschopp: Sure, mStorage[8] would be out of bounds if the array is on 8 elements long.

src/ports/postgres/modules/convex/lmf.sql_in
```diff
@@ -0,0 +1,300 @@
...
+AS 'MODULE_PATHNAME'
+LANGUAGE C IMMUTABLE STRICT;
+
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.lmf_igd_final(
+        state DOUBLE PRECISION[])
+RETURNS DOUBLE PRECISION[]
+AS 'MODULE_PATHNAME'
+LANGUAGE C IMMUTABLE STRICT;
+
+DROP AGGREGATE IF EXISTS MADLIB_SCHEMA.lmf_igd_step(
```
fschopp: Don't add drop statements in the SQL. They are not needed.

src/ports/postgres/modules/convex/lmf.sql_in
```diff
@@ -0,0 +1,300 @@
...
+CREATE TYPE MADLIB_SCHEMA.lmf_result AS (
+        flatten_matrix_u    DOUBLE PRECISION[],
+        flatten_matrix_v    DOUBLE PRECISION[],
+        rmse                DOUBLE PRECISION
+);
+
+--------------------------------------------------------------------------
+-- create SQL functions for IGD optimizer
+--------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.lmf_igd_transition(
```
fschopp: `OR REPLACE` is not needed.

src/ports/postgres/modules/convex/lmf.sql_in
```diff
@@ -0,0 +1,300 @@
...
+--    EXECUTE '
+--    INSERT INTO ' || output_table || '(flatten_matrix_u, flatten_matrix_v, rmse)
+--  SELECT (res).*
+--    FROM (
+--        SELECT MADLIB_SCHEMA.internal_lmf_igd_result(_madlib_state) AS res
+--        FROM _madlib_lmf_igd
+--        WHERE _madlib_iteration = ' || iteration_run || '
+--      ) subq
+--    RETURNING id, rmse'
+--    INTO model_id, rmse;
```
fschopp: A workaround that is commented out is confusing. :)

src/modules/convex/type/model.hpp
```diff
@@ -0,0 +1,116 @@
...
+        int i, j, r;
+        for (i = 0; i < matrixU.rows(); i ++) {
+            for (r = 0; r < matrixU.cols(); r ++) {
+                matrixU(i, r) = (inInitValue * rand()) / RAND_MAX;
+            }
+        }
```
fschopp: We use Eigen's default storage order, which is column-major. It is therefore a good habit to also write loops in that order.

Better yet, use Eigen's specialized operations, e.g., [`unaryExpr ()`](http://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#a23fc4bf97168dee2516f85edcfd4cfe7) in this case.

src/modules/convex/type/model.hpp
```diff
@@ -0,0 +1,116 @@
...
+#define MADLIB_CONVEX_TYPE_MODEL_HPP_
+
+namespace madlib {
+
+namespace modules {
+
+namespace convex {
+
+// The necessity of this wrapper is to allow classes in algo/ and task/ to
+// have a type that they can template over
```
fschopp, line 20: That comment is not immediately clear to me.

fschopp:
I'm currently drawing a class diagram to understand if so many different classes are indeed needed.

src/config/Modules.yml
haying: Have you already got the code structure with this properly placed in mind?

src/modules/convex/algo/igd.hpp
haying: yes, that is what I am doing :)

src/modules/convex/type/state.hpp
haying: But this is not the case if there is a scalar variable at the very end of the components. So, do you think suggesting one more zero (in this case 9) in the aggregate initial string is what we want to do in all modules? You could ignore this if this is no longer important with the new modular transition states.

src/modules/convex/task/lmf.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+template <class Model, class Tuple>
+void
+LMF<Model, Tuple>::gradientInPlace(
+        model_type                          &model,
+        const independent_variables_type    &x, 
+        const dependent_variable_type       &y, 
+        const double                        &stepsize) {
+    // Please refer to the design document for an explanation of the following
+    double e = model.matrixU.row(x.i) * trans(model.matrixV.row(x.j)) - y;
```
haying, line 71: @fschopp Is it better if I use the dbal::eigen_integration::dot() instead if operator*() here to perform dot product?

src/modules/convex/lmf.cpp
haying: @fschopp What's your opinion?

src/modules/convex/lmf.cpp
haying: @fschopp There is actually another problem that there is not "unsigned" types in postgres... Is that why you suggest 2^15-1 instead of 2^16-1? So I just go with "CREATE DOMAIN usmallint AS smallint CHECK (VALUE >= 0);"

src/modules/convex/lmf.cpp
haying: This is still to be made sense to me :( If you are firing a JIRA for this I would place it in the code.

cwelton:
Is this PR ready for merge?

PR #177, opened 2012-07-16
Modular Transition States

fschopp:
_Please do not merge using the merge button. Let's keep the history linear._

Jira: MADLIB-499

Initial list of changes:
- Initial revision of transition states that allow recursion and reusability: A
  transition state can consist of other transition states. No serialization
  takes place, instead we operate directly on the byte stream.
- Adapted linear regression to the new programming model.
- Added a new type 'blob' within the database that is 8-byte aligned. (The
  PostgreSQL default for bytea is only 4-byte alignment, which is not enough.)

More changes:
- Made "dynamic structs" an integral platform-independent part of the
  abstraction layer.
- Dynamic structs now can be initialized even with ByteStreams that are not long
  enough. E.g., this could happen in the merge function, when one segment did
  not have any rows.
- Started minor modifications to the existing platform-specific part of the AL:
  E.g., MutableTransparentHandle<typename T> is gone in favor of
  TranspararentHandle<typename T, bool IsMutable>. This needs to be continue
  for other platform-specific classes.


src/modules/regress/LinearRegression_impl.hpp
```diff
@@ -0,0 +1,252 @@
...
+template <class Parent>
+template <class OtherParent>
+inline
+LinearRegressionAccumulator<Parent>&
+LinearRegressionAccumulator<Parent>::operator=(
+    const LinearRegressionAccumulator<OtherParent>& inOther) {
+
+    this->copy(inOther);
+    return *this;
+}
```
fschopp: The function seems to be used only once. But it might still be good to have when linear regression is used as a sub-state?

src/modules/regress/LinearRegression_impl.hpp
```diff
@@ -0,0 +1,252 @@
...
+ * @sa For the mathematical description, see \ref grp_linreg.
+ */
+template <class Parent>
+inline
+LinearRegression&
+LinearRegression::compute(const LinearRegressionAccumulator<Parent>& inState) {
+    Allocator& allocator = defaultAllocator();
+
+    // See MADLIB-138. At least on certain platforms and with certain versions,
+    // LAPACK will run into an infinite loop if pinv() is called for non-finite
```
cwelton: Check seems sensible.  But now that we are no longer using LAPACK isn't the comment a little confusing in terms of current justification?

fschopp: Yes, the comment should be changed.

src/ports/postgres/dbconnector/State_impl.hpp
```diff
@@ -0,0 +1,441 @@
...
+    BinaryStream<StreamBuf>& inStream,
+    Ref<T>& inReference) {
+
+    T* data = inStream.template read<T>();
+    if (!inStream.isInDryMode())
+        inReference.rebind(data);
+    return inStream;
+}
+
+/*template <class T, bool Immutable>
```
cwelton: dead code?

src/ports/postgres/dbconnector/State_impl.hpp
```diff
@@ -0,0 +1,441 @@
...
+operator>>(
+    BinaryStream<StreamBuf>& inStream,
+    dbal::eigen_integration::HandleMap<
+        EigenType,
+        TransparentHandle<typename EigenType::Scalar>
+    >& inReference) {
+
+    typedef BinaryStream<StreamBuf> BinaryStream_type;
+
+//    typename add_const_if_immutable<
```
cwelton: dead code?

fschopp: Yup.

src/ports/postgres/modules/regress/linear.sql_in
```diff
@@ -241,26 +241,56 @@ CREATE TYPE MADLIB_SCHEMA.linregr_result AS (
     condition_no DOUBLE PRECISION
 );
 
+/*
+ * Create type blob with 8-byte alignment.
+ */
```
cwelton: Reasons to avoid the name "blob"
- could be confusing this with Postgres "Large Objects"
- "blob" is a reserved keyword in the SQL Standard

I suggest something more like binarystream, bytestream, bytea8, or something similar.

src/ports/postgres/modules/regress/linear.sql_in
```diff
@@ -241,26 +241,56 @@ CREATE TYPE MADLIB_SCHEMA.linregr_result AS (
...
+CREATE TYPE MADLIB_SCHEMA.blob;
+
+CREATE FUNCTION MADLIB_SCHEMA.blobin(cstring)
+RETURNS MADLIB_SCHEMA.blob AS 'byteain'
+LANGUAGE internal IMMUTABLE STRICT;
```
cwelton: If you agree we should have naming standards around internal functions these should be included.  E.g. "__bytestreamin"

src/ports/postgres/modules/regress/linear.sql_in
```diff
@@ -241,26 +241,56 @@ CREATE TYPE MADLIB_SCHEMA.linregr_result AS (
...
+
+CREATE TYPE MADLIB_SCHEMA.blob(
+    INPUT = MADLIB_SCHEMA.blobin,
+    OUTPUT = MADLIB_SCHEMA.blobout,
+    RECEIVE = MADLIB_SCHEMA.blobrecv,
+    SEND = MADLIB_SCHEMA.blobsend,
+    LIKE = pg_catalog.bytea,
+    ALIGNMENT = double
+);
```
cwelton: This should be defined in a _sql file specific to the abstraction layer, not part of linear.sql_in

cwelton: Would it make sense to have cast operators to bytea?  

Or is the intent here that this is solely internal, in which case perhaps a name like "madlib.__internal_binary_state" would be preferable?

fschopp: I think it should be internal.

src/ports/postgres/modules/regress/linear.sql_in
fschopp: I agree in general, even though underscores for function names are not ideal, because C function names should not begin with underscores (such names are reserved).

cwelton:
A few minor things to cleanup, but overall looks pretty reasonable to me.

src/ports/postgres/modules/regress/linear.sql_in
fschopp: Fine with me.

src/modules/regress/LinearRegression_proto.hpp
```diff
@@ -0,0 +1,62 @@
...
+class LinearRegressionAccumulator
+  : public State<LinearRegressionAccumulator<Parent>, Parent> {
+public:
+    MADLIB_STATE_TYPEDEFS(LinearRegressionAccumulator, Parent)
+
+    LinearRegressionAccumulator(Init_type& inInitialization);
+    void bind(BinaryStream_type& inStream);
+    LinearRegressionAccumulator& feed(MappedColumnVector x, double y);
+    template <class OtherParent> LinearRegressionAccumulator& feed(
+        const LinearRegressionAccumulator<OtherParent>& inOther);
```
haying: I am not sure that having feed(x, y) and feed(other) the same function name "feed" is good. Although you can think that both are feeding something into the accumulator, the former one is essentially a transition function and the latter one a merge function. I can't quite get the necessity of hiding the differences of the two by assigning the same "name", though different signature.

src/modules/regress/LinearRegression_proto.hpp
```diff
@@ -0,0 +1,62 @@
...
+    template <class OtherParent> LinearRegressionAccumulator& operator=(
+        const LinearRegressionAccumulator<OtherParent>& inOther);
+
+    Ref<uint64_type> numRows;
+    Ref<uint16_type> widthOfX;
+    Ref<double_type> y_sum;
+    Ref<double_type> y_square_sum;
+    MappedColumnVector_type X_transp_Y;
+    MappedMatrix_type X_transp_X;
+};
```
haying: Code with class type names like "State", "Ref" can be made easier to read if, within the same file, a explicit "using" statement or even comment saying which namespaces they are under. It is actually better to have these also for types like uint16_type until they are really everywhere in the code base. But since they have a special name, it is okay to find them by "grep". The declaration of "State" or "Ref" is hard to find... They are too common.

src/modules/regress/LinearRegression_impl.hpp
```diff
@@ -0,0 +1,252 @@
...
+        throw std::domain_error("Design matrix is not finite.");
+    else if (x.size() > std::numeric_limits<uint16_t>::max())
+        throw std::domain_error("Number of independent variables cannot be "
+            "larger than 65535.");
+
+    // Initialize in first iteration
+    if (numRows == 0) {
+        widthOfX = static_cast<uint16_t>(x.size());
+        this->resize();
+    }
```
haying, line 81: This block of code (from checking numRows == 0) may have to be moved in a upper level, to be consistent with other modules that require args other than (x, y) to initialize the state, e.g., logistic regression (actually any iterative algorithms) requires a previous state.

src/ports/postgres/dbconnector/State_proto.hpp
```diff
@@ -0,0 +1,405 @@
...
+
+protected:
+    T *mPtr;
+};
+
+template <class Storage, bool IsMutable = Storage::isMutable>
+struct StorageTraits;
+
+template <class StreamBuf>
+class BinaryStream {
```
haying: The name "BinaryStream" (ambiguous "binary"):
  Looking at the interfaces of this class, it looks like something similar to std::iostream, in which context, "binary" probably means "ios_base::binary" (no memory alignment) and it is not why you name this class with the word "Binary". Something like "ByteStream", "WordStream", or "CharStream" may or may not work, but it does at least avoid this ambiguous...

src/modules/regress/LinearRegression_proto.hpp
haying: The same issue applies when it refers to function names: the calling of "Null" in Linregr_final::run(...).

One more place is the "trangularView<Lower>()". It needs some effort to find out where "Lower" comes from and finally confirm that "Upper" instead of "Higher" is what one needs when he wants to play with another side of the matrix :)

src/ports/postgres/dbconnector/State_proto.hpp
```diff
@@ -0,0 +1,405 @@
...
+
+    using Base::mParent;
+    using Base::mBinaryStream;
+};
+
+
+template <class Derived, class Parent, bool IsMutable = Parent::isMutable>
+class State : public StateBase<Derived, Parent, Parent::isMutable> {
+public:
+    typedef StateBase<Derived, Parent> Base;
```
haying: "Parent", "Base", "Derived"... This seems to have everything necessary for confusing people in C++ class inheritance... I need some more comment or a conversation with Florian before I can try to understand this...

src/ports/postgres/dbconnector/State_proto.hpp
```diff
@@ -0,0 +1,405 @@
...
+     * @brief The associated storage to the stream (similar to a streambuf for
+     *     IOStreams)
+     */
+    StreamBuf_type* mStreamBuf;
+
+    /**
+     * @brief In dry mode read/write operations will only move the cursor, but
+     *     other objects are not touched
+     */
+    int32_t mDryMode;
```
haying: "Dry Mode C++" from Google still shows up some air conditioner related pages... What if in wet mode :)

src/ports/postgres/dbconnector/State_proto.hpp
fschopp: Yep, this is light on documentation so far. The main point is that we have two hierarchies: The inheritance hierarchy and the containment hierarchy. A state A can **contain** other states B and C. In that case A is a "parent" of B and C. But all transition states also **inherit** from class `State`, i.e., have `State` as "base" class. Now the C++ idiom used here is the "curiously recurring template pattern", i.e., transition states derive from `State` using itself as template argument. This achieves static polymorphism and avoids run-time virtual-table lookups. (Benefits: Slight performance benefit plus more compile-time checking.)

fschopp: Maybe I should rename "Parent" to "Container"!

src/ports/postgres/dbconnector/State_proto.hpp
fschopp: Dry mode = lots of hot air. :)

I guess I wanted to write "dry run"...

src/ports/postgres/dbconnector/State_proto.hpp
```diff
@@ -0,0 +1,405 @@
...
+        inState.bind(inStream);
+        inStream.template seek<MAXIMUM_ALIGNOF>(0, std::ios_base::cur);
+        if (!inStream.isInDryMode())
+            inState.mEnd = inStream.tell();
+        return inStream;
+    }
+
+protected:
+    size_t mBegin;
+    size_t mEnd;
```
haying: What do you think about adding operator= for value copying? or this has been done somehow thru copy constructor?

src/ports/postgres/modules/regress/linear.sql_in
```diff
@@ -241,26 +241,56 @@ CREATE TYPE MADLIB_SCHEMA.linregr_result AS (
...
+CREATE FUNCTION MADLIB_SCHEMA.blobsend(MADLIB_SCHEMA.blob)
+RETURNS bytea AS 'byteasend'
+LANGUAGE internal IMMUTABLE STRICT;
+
+CREATE TYPE MADLIB_SCHEMA.blob(
+    INPUT = MADLIB_SCHEMA.blobin,
+    OUTPUT = MADLIB_SCHEMA.blobout,
+    RECEIVE = MADLIB_SCHEMA.blobrecv,
+    SEND = MADLIB_SCHEMA.blobsend,
+    LIKE = pg_catalog.bytea,
```
haradh1: GPDB doesn't apparently have LIKE option.

src/modules/regress/LinearRegression_proto.hpp
fschopp: My first idea was to define a tuple type and `operator<<` instead of `feed`. This operator I would have used for both transition and merge. However, I found working with tuple types less convenient than just having each field as an individual argument of a transition function. Despite the change, I stuck with using the same name for both transition and merge though.

Now for the convex optimization framework, it seems that a tuple type is unavoidable at least before C++11. I wonder what you think about `operator<<` versus `transition` and `merge`.

In any case, a merge function is not really a different concept. It is in fact a transition function during the second-phase aggregation.

src/modules/regress/LinearRegression_proto.hpp
fschopp: While I agree in principle, one should never have using directives or declarations at file scope in header files. However, using declarations in a definition of a class can only be used to import names from a base class.

src/modules/regress/LinearRegression_impl.hpp
fschopp, line 81: I have no definite answer yet. But I want to mention that for some modules initialization (at least in the first iteration) will depend on the first input tuple and cannot be done before seeing this.

src/ports/postgres/dbconnector/State_proto.hpp
fschopp: I see your point and tend to agree. Names are but sound and smoke. I'll change it!

src/modules/regress/LinearRegression_proto.hpp
cwelton: Philosophically I disagree that merge and transition are the same concept.
- Transition is a single row accumulator, Merge is a multi row combiner
- Transition is necessary for correct results, Merge is an optional optimization
- Merge is not just about second-phase aggregation, but about multi-phase aggregation, though in practice our use currently only makes use of the two phase techniques (e.g. we don't use cube and rollup aggregation).

I've had to explain the difference between Merge and Reduce often enough that I think the distinctions are important.

fschopp:
https://github.com/fschopp/madlib/tree/modular_states_temp has seen some code reorganization. That branch does not compile yet though. Once it does it will replace what is in this branch.

cwelton:
Florian - please comment on the merge-readiness of this PR.   

Is the intent of your previous comment that we should do nothing with this PR until your private branch is ready?  If so, what timeframe are we looking at?  Before dev-complete?

fschopp:
Yes, once my reorganization is done, I want that reorganized code to go into MADlib v0.5. Yes, this will happen soon and before dev-complete. I do not want to merge this pull request before the reorganization is done.

src/ports/postgres/dbconnector/State_proto.hpp
fschopp: See, e.g., LinearRegressionAccumulator::operator=()

src/modules/regress/LinearRegression_impl.hpp
haying, line 81: What I did is that I avoided the dependencies on the first tuple by explicitly passing necessary arguments (e.g., numOfCoefficients) as a separate function parameter of the UDA. I found it cleaner, since following the "Anatomy of an iterative MADlib module" (Start(), Initialize(), ...), we are not supposed to read information from the first tuple for Initialize(), but through arguments instead.

src/ports/postgres/dbconnector/State_proto.hpp
haying: A function like LinearRegressionAccumulator::operator=() is exactly what I like to abstract out; otherwise, we are having the exactly same code for most (prob. all) derived classes of DynamicStruct

src/modules/regress/LinearRegression_impl.hpp
fschopp, line 81: That is, of course, an option, and arguably a good discussion to have. In general, I am not dogmatic about the anatomy of a MADlib module and am fine with moving the intra-iteration (per aggregate) initialization to after reading the first input tuple (or maybe even _k_ input tuples). Conceptually, we would just be giving the initialize method a look-ahead of one row (or _k_ rows, respectively).

On the other hand, there are cases where we need to pass explicit arguments anyways, when some information cannot be deduced with a look-ahead of just constantly many rows (e.g., the some hypothesis tests need count information).

src/ports/postgres/dbconnector/State_proto.hpp
fschopp: Sure, but there are some C++ peculiarities to keep in mind.

I would have to look up how C++ treats the case when only a base class of `LinearRegressionAccumulator<Container>` contains `operator=(LinearRegressionAccumulator<Container>&)`, but not `LinearRegressionAccumulator<Container>` itself. Does it do an element-wise copy then, or would it (only) call the base class's `operator=`? If the first, an `operator=` in a derived class will always be necessary. If the latter, we could get rid of the derived class's `operator=` implementations.

In any case, the base class would also need `LinearRegressionAccumulator::operator=(LinearRegressionAccumulator<OtherContainer>&)`. In order to allow that, one could give `operator=` in `DynamicStruct` another template parameters, and require derived classes to implement something like the [`rebind` in STL allocators](http://en.cppreference.com/w/cpp/concept/Allocator).

So far, reducing the 2-line implementation of `operator=` in derived classes has not been enough incentive for me to look into this a bit deeper. But I will (after my vacation). But even in the current implementation, note that there is no logic in `LinearRegressionAccumulator::operator=` but it only calls the Base class's `copy` function!

src/modules/regress/LinearRegression_impl.hpp
haying, line 81: I see. I am kind of convinced. Sticking to a specified anatomy for no special reasons is not necessary. So I just leave one more point here in case in the future we'd like to make a more serious decision on this. Some DBMS's (e.g., Oracle) allows users to provide an Initialize() function for a UDA. In such case, a peek at data tuple may not be as natural as in GPDB/Postgres. This is a point for portability.

fschopp:
Since I am on vacation, I do not plan to add more before our targeted dev-complete date, Aug 20. This is now ready to be merged from my point of view. Dynamic Structs are now an integral part of the C++ AL.

PR #178, opened 2012-07-17
Added a sampling module, currently containing only weighted_sample().

fschopp:
_Please let me know if you have comments. I'll merge myself then._

Jira: MADLIB-584

Sampling module:
- Added a data-parallel UDA for weighted sampling of a single row

C++ AL:
- Added a boost/TR1-style front-end to PostgreSQL's random number generator.

### DIFF
--- a/src/modules/sample/weighted_sample.cpp
+++ b/src/modules/sample/weighted_sample.cpp
@@ -1,0 +1,115 @@
+/* ----------------------------------------------------------------------- *//**
+ *
+ * @file weighted_sample.cpp
+ *
+ * @brief Generate a single weighted random sample
+ *
+ *//* ----------------------------------------------------------------------- */
+
+#include <dbconnector/dbconnector.hpp>
+#include <modules/shared/HandleTraits.hpp>
+
+#include <boost/tr1/random.hpp>
+
+#include "weighted_sample.hpp"
+
+// Import TR1 names (currently used from boost). This can go away once we make
+// the switch to C++11.
+namespace std {
+    using tr1::bernoulli_distribution;
+}
+
+namespace madlib {
+
+namespace modules {
+
+namespace sample {
+
+/**
+ * @brief Transition state for weighted sample
+ *
+ * Note: We assume that the DOUBLE PRECISION array is initialized by the
+ * database with length 2, and all elemenets are 0.
+ */
+template <class Handle>
+class WeightedSampleTransitionState {
+public:
+    WeightedSampleTransitionState(const AnyType &inArray)
+      : mStorage(inArray.getAs<Handle>()),
+        sample_id(&mStorage[1]),
+        weight_sum(&mStorage[0]) { }
+
+    inline operator AnyType() const {
+        return mStorage;
+    }
+
+private:
+    Handle mStorage;
+
+public:
+    typename HandleTraits<Handle>::ReferenceToInt64 sample_id;
+    typename HandleTraits<Handle>::ReferenceToDouble weight_sum;
+};
+
+/**
+ * @brief Perform the weighted-sample transition step
+ */
+AnyType
+weighted_sample_transition::run(AnyType& args) {
+    WeightedSampleTransitionState<MutableArrayHandle<double> > state = args[0];
+    uint64_t identifier = args[1].getAs<int64_t>();
+    double weight = args[2].getAs<double>();
+
+    // Instead of throwing an error, we will just ignore rows with a negative
+    // weight
+    if (weight > 0.) {
+        state.weight_sum += weight;
+        std::bernoulli_distribution success(weight / state.weight_sum);
+        // Note that a NativeRandomNumberGenerator object is stateless, so it
+        // is not a problem to instantiate an object for each RN generation...
+        NativeRandomNumberGenerator generator;
+        if (success(generator))
+            state.sample_id = identifier;
+    }
+
+    return state;
+}
+
+/**
+ * @brief Perform the merging of two transition states
+ */
+AnyType
+weighted_sample_merge::run(AnyType &args) {
+    WeightedSampleTransitionState<MutableArrayHandle<double> > stateLeft
+        = args[0];
+    WeightedSampleTransitionState<ArrayHandle<double> > stateRight = args[1];
+
+    // FIXME: Once we have more modular states (independent of transition/merge
+    // function), implement using the logic from the transition function
+    stateLeft.weight_sum += stateRight.weight_sum;
+    std::bernoulli_distribution success(
+        stateRight.weight_sum / stateLeft.weight_sum);
+    // Note that a NativeRandomNumberGenerator object is stateless, so it
+    // is not a problem to instantiate an object for each RN generation...
+    NativeRandomNumberGenerator generator;
+    if (success(generator))
+        stateLeft.sample_id = stateRight.sample_id;
+
+    return stateLeft;
+}
+
+/**
+ * @brief Perform the weighted-sample final step
+ */
+AnyType
+weighted_sample_final::run(AnyType &args) {
+    WeightedSampleTransitionState<ArrayHandle<double> > state = args[0];
+
+    return static_cast<int64_t>(state.sample_id);
+}
+
+} // namespace stats
+
+} // namespace modules
+
+} // namespace madlib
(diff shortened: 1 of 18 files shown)


haying:
The word "weighted" does not seems to appear in the design document. Is this an implementation of something inside the design document or not?

src/ports/postgres/modules/sample/test/sample.sql_in
```diff
@@ -0,0 +1,27 @@
...
+-- once per 10,000 runs.
+SELECT
+    assert(
+        kolmogorov_cdf(
+            (ks_test(first, value, 10000, 55 ORDER BY value)).k_statistic
+        ) < 1e-5,
+        'Results of weighted_sample() do not match the expected distribution.'
+    )
+FROM (
+    SELECT TRUE AS first, madlib.weighted_sample(i, i) AS value
```
haying: "madlib." does not seem to be needed.

fschopp: Indeed. In fact, it would cause an error if the schema name is not madlib!

src/modules/sample/weighted_sample.cpp
```diff
@@ -0,0 +1,115 @@
...
+    // Instead of throwing an error, we will just ignore rows with a negative
+    // weight
+    if (weight > 0.) {
+        state.weight_sum += weight;
+        std::bernoulli_distribution success(weight/state.weight_sum);
+        // Note that an NativeRandomNumberGenerator object is stateless, so it
+        // is not a problem to instantiate an object for each RN generation...
+        NativeRandomNumberGenerator generator;
+        if (success(generator))
+            state.sample_id = identifier;
```
haying: Even after some search, I am still not exactly sure how bernoulli_distribution.operator(...) works. I assume it simply returns something like 
"generator.operator() < generator.max() \* this->p ? true : false".
(http://en.cppreference.com/w/cpp/numeric/random/bernoulli_distribution/operator()
http://www.boost.org/doc/libs/1_35_0/libs/math/doc/sf_and_dist/html/math_toolkit/dist/dist_ref/dists/bernoulli_dist.html
http://www.cplusplus.com/reference/std/random/bernoulli_distribution/operator()/ )

src/ports/postgres/dbconnector/NativeRandomNumberGenerator_impl.hpp
```diff
@@ -0,0 +1,70 @@
...
+NativeRandomNumberGenerator::NativeRandomNumberGenerator() { }
+
+/**
+ * @brief Sets the current state of the engine
+ */
+inline
+void
+NativeRandomNumberGenerator::seed(result_type inSeed) {
+    DirectFunctionCall1(setseed, Float8GetDatum(inSeed));
+}
```
haying, line 28: In terms of how the user of this class should perform seeding, I think it should at least be explained in comment somewhere. Since I can still abuse this seed function when I use this class with the interfaces provided now. :)

haying:
No more review comment at this point.

src/modules/sample/weighted_sample.cpp
fschopp: Yep, see boost/random/bernoulli_distribution.hpp.

src/ports/postgres/dbconnector/NativeRandomNumberGenerator_impl.hpp
fschopp, line 28: Added a sentence that this does nothing more than calling the backend's seed function. The best way to use it is use it only for debugging!

src/ports/postgres/modules/sample/test/sample.sql_in
```diff
@@ -0,0 +1,27 @@
...
+-- We compare two samples that we know stem from the same distribution. We
+-- require that the probability of seeing a Kolmogorov statistic at least as
+-- extreme as the one observed is at most 1e-5.
+-- Hence, the following test fails with probability at most 1e-5, i.e., at most
+-- once per 10,000 runs.
+SELECT
+    assert(
+        kolmogorov_cdf(
+            (ks_test(first, value, 10000, 55 ORDER BY value)).k_statistic
+        ) < 1e-5,
```
fschopp: This is pretty big nonsense. The KS test is not for discrete distributions. Will change to chi2_gof_test.

fschopp:
Changed the design document, so that it says `weighted_sample` instead of `discrete_sample` now.

src/modules/sample/weighted_sample.cpp
```diff
@@ -0,0 +1,115 @@
...
+    WeightedSampleTransitionState<MutableArrayHandle<double> > state = args[0];
+    uint64_t identifier = args[1].getAs<int64_t>();
+    double weight = args[2].getAs<double>();
+
+    // Instead of throwing an error, we will just ignore rows with a negative
+    // weight
+    if (weight > 0.) {
+        state.weight_sum += weight;
+        std::bernoulli_distribution success(weight/state.weight_sum);
+        // Note that an NativeRandomNumberGenerator object is stateless, so it
```
haradh1: an -> a

src/modules/sample/weighted_sample.cpp
```diff
@@ -0,0 +1,115 @@
...
+weighted_sample_transition::run(AnyType& args) {
+    WeightedSampleTransitionState<MutableArrayHandle<double> > state = args[0];
+    uint64_t identifier = args[1].getAs<int64_t>();
+    double weight = args[2].getAs<double>();
+
+    // Instead of throwing an error, we will just ignore rows with a negative
+    // weight
+    if (weight > 0.) {
+        state.weight_sum += weight;
+        std::bernoulli_distribution success(weight/state.weight_sum);
```
haradh1: style:  can you put spaces around operator '/'?

src/ports/postgres/dbconnector/NativeRandomNumberGenerator_impl.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @brief Sets the current state of the engine
+ *
+ * All this function does is calling the backend's seed functions.
+ */
+inline
+void
+NativeRandomNumberGenerator::seed(result_type inSeed) {
+    DirectFunctionCall1(setseed, Float8GetDatum(inSeed));
```
haradh1, line 27: Why don't you call srandom() instead of setseed()?  I don't see the reason to use SQL function here.  Same for drandom()

src/modules/sample/weighted_sample.cpp
```diff
@@ -0,0 +1,115 @@
...
+    // FIXME: Once we have more modular states (independent of transition/merge
+    // function), implement using the logic from the transition function
+    stateLeft.weight_sum += stateRight.weight_sum;
+    std::bernoulli_distribution success(
+        stateRight.weight_sum/stateLeft.weight_sum);
+    // Note that an NativeRandomNumberGenerator object is stateless, so it
+    // is not a problem to instantiate an object for each RN generation...
+    NativeRandomNumberGenerator generator;
+    if (success(generator))
+        stateLeft.sample_id = stateRight.sample_id;
```
haradh1: Isn't there possibility that either of left/right has no sample_id?  In that case, do we still need to check with success()?

src/ports/postgres/dbconnector/NativeRandomNumberGenerator_impl.hpp
fschopp, line 27: Because this should map to the backend's RNG (which could change in future versions).

src/ports/postgres/modules/sample/sample.sql_in
```diff
@@ -0,0 +1,70 @@
...
+ *     identifier occurs multiple times, the probability of sampling this
+ *     identifier is proportional to the sum of its weights.
+ * @param weight Weight for row. A negative value here is treated has zero
+ *     weight.
+ * @return \c identifier of the selected row. The probability of sampling any
+ *     particular row <tt>(identifier, weight)</tt> is
+ *     <tt>weight/SUM(weight)</tt>.
+ */
+CREATE AGGREGATE MADLIB_SCHEMA.weighted_sample(
+    /*+ "identifier" */ BIGINT,
```
haradh1, line 62: Do we support "any type" in the future once we get more generic transition state?

src/modules/sample/weighted_sample.cpp
haradh1: Say, all rows that have been seen on the right were negative weights.

src/ports/postgres/modules/sample/sample.sql_in
fschopp, line 62: I don't think there is a way to make this work with PostgreSQL. The transition state is of type bytea, so the final function cannot return anytype...

src/modules/sample/weighted_sample.cpp
fschopp: Then, the weight for stateRight would be 0. So no need to make this a special case.

PR #179, opened 2012-07-17
Improved Design Doc: Low-rank Matrix Factorization

haying:
JIRA: MADLIB-578
1. added more descriptions after the pseudo-code
2. made the symbols consistent with the code
3. added some more citations

### DIFF
--- a/doc/design/modules/low-rank-matrix-decomposition.tex
+++ b/doc/design/modules/low-rank-matrix-decomposition.tex
@@ -1,48 +1,88 @@
-% When using TeXShop on the Mac, let it know the root document. The following must be one of the first 20 lines.
+% When using TeXShop on the Mac, let it know the root document.
+% The following must be one of the first 20 lines.
 % !TEX root = ../design.tex
 
 \chapter{Low-rank Matrix Factorization}
 
 % Abstract. What is the problem we want to solve?
-This module implements "factor model" for representing a sparse matrix using a low-rank approximation \cite{DBLP:conf/icml/SrebroJ03}.
-Mathematically, this model seeks to find matrices U and V that, for any given sparse matrix A, minimizes:
-\[ ||\boldsymbol A - \boldsymbol UV' ||_2 \]
-subject to $rank(\boldsymbol UV') \leq k$, where $||\cdot||_2$ denotes the Frobenius norm and $k \leq rank(\boldsymbol A)$.
-If $A$ is $m \times n$, then $U$ will be $m \times k$ and $V$ will be $n \times k$.
-This model is not intended to do the full decomposition, or to be used as part of inverse procedure.
-This model has been widely used in recommendation systems, such as Netflix \cite{:TheNetflixPrize07}.
+This module implements "factor model" for representing a sparse matrix
+using a low-rank approximation \cite{DBLP:conf/icml/SrebroJ03}.
+Mathematically, this model seeks to find matrices U and V (also referred 
+as factors) that, for any given sparse matrix A, minimizes:
+\[ ||\boldsymbol A - \boldsymbol UV^{T} ||_2 \]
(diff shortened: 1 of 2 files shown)


doc/design/modules/low-rank-matrix-decomposition.tex
```diff
@@ -1,48 +1,88 @@
```
fschopp, line 0: Lots of changes here seem to be due to rewrapping text. It would be good to keep such cosmetic changes separate from content changes.

BTW, I personally do not like manually wrapping LaTeX code. I instead prefer keeping paragraphs on a single line, even if GitHub does not wrap long lines and is therefore rather unusable as a diff tool. Rationale: With manual wrapping there is no usable diff tool (unless you accept that editing produces very ragged lines in the LaTeX source code). Without manual wrapping, there are at least external diff tools that do a good job.

doc/design/modules/low-rank-matrix-decomposition.tex
```diff
@@ -1,48 +1,88 @@
...
-\[ ||\boldsymbol A - \boldsymbol UV' ||_2 \]
-subject to $rank(\boldsymbol UV') \leq k$, where $||\cdot||_2$ denotes the Frobenius norm and $k \leq rank(\boldsymbol A)$.
-If $A$ is $m \times n$, then $U$ will be $m \times k$ and $V$ will be $n \times k$.
-This model is not intended to do the full decomposition, or to be used as part of inverse procedure.
-This model has been widely used in recommendation systems, such as Netflix \cite{:TheNetflixPrize07}.
+This module implements "factor model" for representing a sparse matrix
+using a low-rank approximation \cite{DBLP:conf/icml/SrebroJ03}.
+Mathematically, this model seeks to find matrices U and V (also referred 
+as factors) that, for any given sparse matrix A, minimizes:
+\[ ||\boldsymbol A - \boldsymbol UV^{T} ||_2 \]
```
fschopp, line 12: You can get the two parallel lines for the norm operator via `\|` (as opposed to `||`).

doc/design/modules/low-rank-matrix-decomposition.tex
```diff
@@ -1,48 +1,88 @@
...
+\algoutput{Factors $U$ ($m \times r$) and $V$ ($n \times r$)}
+\algprecond{$iteration = 0$}
 \begin{algorithmic}[1]
-	\State $U \set \{ \text{each element is chosen uniformly at random from } (0, 1) \}$
-	\State $V \set \{ \text{each element is chosen uniformly at random from } (0, 1) \}$
+	\State $U \set GenerateFactor(m, r)$
+	\State $V \set GenerateFactor(n, r)$
 	\Repeat
-		\State $i \set i + 1$
+		\State $iteration \set iteration + 1$
```
fschopp, line 50: I suggest `\mathit{iteration}`, otherwise LaTeX will typeset as if it was a product of all the letters (note the different spacing between letters).

fschopp:
@haying My comments are obviously very minor. Should I just merge, or do you want to address anything?

doc/design/modules/low-rank-matrix-decomposition.tex
haying, line 0: I see. I am cool with that. But what do you think about keeping sentences instead of paragraphs on a single line?

haying:
@fschopp Only one wording that I'd like to change: "widely used in recommendation systems, such as Netflix, and image processing" -> "widely used in recommendation systems (e.g., Netflix [?]) and feature selection (e.g., image processing [?])"

Apparently, more people care about feature selection then just image processing. :)

Please go ahead and merge it at your convention. Thanks.

doc/design/modules/low-rank-matrix-decomposition.tex
haying, line 50: same for "temp", in the same algorithm block

fschopp:
Merged with commit d74d950.

PR #180, opened 2012-07-20
Enhance the encoding method

aojwang:
Jira: MADLIB-592

In MADlib 0.4 version, as we encode the training table column by column, the scalability is not good and the encoding time is very long for some datasets with a large number of columns, such as internet_ads (it has about 1600 columns). Now, we encode all the columns of training table in parallel.

The steps of the encoding method for training.

``` javascript
1. Breakup the training table(n columns and m rows) to a new table with fixed columns (id, fid, fval, is_cont, class) and n * m rows. We call this new table as breakup table.

2. Based on the breakup table, generating the Key-Value (KV) (We only have two KV tables: One is for class and the other is for features)

3. Generate the encoded table by joining the breakup table with the KV table for features.
```

The steps of the encoding method for classification

``` javascript
1. Transform the classification table to the breakup table.

2. Generate the encoded table by joining the breakup table with the KV table for features, which is produced during training.
```

### DIFF
--- a/methods/cart/src/pg_gp/dt_preproc.sql_in
+++ b/methods/cart/src/pg_gp/dt_preproc.sql_in
@@ -610,11 +610,12 @@ DECLARE
 BEGIN
     PERFORM MADLIB_SCHEMA.__assert_table(meta_tbl_name, 't');
     
-    -- get the Key-Value tables
+    -- if one of those KV tables doesn't exist, 
+    -- we raise exception.
     curstmt = MADLIB_SCHEMA.__format
         (
-            'SELECT MADLIB_SCHEMA.__regclass_to_text
-                (table_oid) as table_name 
+            'SELECT MADLIB_SCHEMA.__assert_table
+                (MADLIB_SCHEMA.__regclass_to_text(table_oid), ''t'') 
              FROM 
                 (
                 SELECT table_oid
@@ -626,11 +627,7 @@ BEGIN
                  meta_tbl_name
              ]
         );
-    
-    -- drop all the Key-Value tables
-    FOR name IN EXECUTE curstmt LOOP
-        PERFORM MADLIB_SCHEMA.__assert_table(name, 't');
-    END LOOP;
+    EXECUTE curstmt;
 END
 $$ LANGUAGE PLPGSQL;
 
@@ -1317,7 +1314,7 @@ BEGIN
             'CREATE TEMP TABLE %(id, fvals, class) AS
              SELECT
                  id,
-                 madlib.__array_indexed_agg(fval, %, fid) as fvals,
+                 MADLIB_SCHEMA.__array_indexed_agg(fval, %, fid) as fvals,
                  min(class)::INT as class
              FROM %
              GROUP BY id
--- a/methods/cart/src/pg_gp/dt_preproc.sql_in
+++ b/methods/cart/src/pg_gp/dt_preproc.sql_in
@@ -482,7 +482,7 @@ BEGIN
     
     -- drop the metatable
     EXECUTE 'DROP TABLE ' || meta_tbl_name || ' CASCADE;';
-end
+END
 $$ LANGUAGE PLPGSQL;
 
 
@@ -591,6 +591,48 @@ BEGIN
 END
 $$ LANGUAGE PLPGSQL;
 
+
+/*
+ * @brief Validate if the metatable exists or not.
+ *        Validate if the tables in "table_oid" column exists or not.
+ *
+ * @param meta_tbl_name     The full name of the metatable.
+ *
+ */
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__validate_metatable 
+    (
+    meta_tbl_name TEXT
+    )
+RETURNS VOID AS $$
+DECLARE
+    curstmt TEXT;
+    name    TEXT;
+BEGIN
+    PERFORM MADLIB_SCHEMA.__assert_table(meta_tbl_name, 't');
+    
+    -- get the Key-Value tables
+    curstmt = MADLIB_SCHEMA.__format
+        (
+            'SELECT MADLIB_SCHEMA.__regclass_to_text
+                (table_oid) as table_name 
+             FROM 
+                (
+                SELECT table_oid
+                FROM %
+                WHERE table_oid IS NOT NULL
+                GROUP BY table_oid
+                ) t',
+             ARRAY[
+                 meta_tbl_name
+             ]
+        );
+    
+    -- drop all the Key-Value tables
+    FOR name IN EXECUTE curstmt LOOP
+        PERFORM MADLIB_SCHEMA.__assert_table(name, 't');
+    END LOOP;
+END
+$$ LANGUAGE PLPGSQL;
 
 /*
  * @brief Get the number of distinct values for the feature with given ID.
--- a/methods/cart/src/pg_gp/dt_preproc.sql_in
+++ b/methods/cart/src/pg_gp/dt_preproc.sql_in
@@ -255,23 +255,6 @@ CREATE TYPE MADLIB_SCHEMA.__enc_tbl_result AS
     post_proc_time           INTERVAL
 );
 
-/*
- * @brief Test if the input value is a missing value or not.
- *
- * @param val   The value to be tested.
- *
- * @note The supported missing values are: '', '?', and NULL.
- *       The input value will be trimed.
- *
- */
-CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__is_miss_value
-    (
-    val TEXT 
-    ) 
-RETURNS BOOL
-AS 'MODULE_PATHNAME', 'dt_is_miss_value'
-LANGUAGE C IMMUTABLE;
-
 
 /*
  * @brief Check if the input table has unsupported data type or not.
@@ -1087,20 +1070,16 @@ m4_changequote(>>>`<<<, >>>'<<<)
 
     EXECUTE curstmt;
 
-    IF (h2hmv_routine_id = 1) THEN
-        where_txt = ' WHERE NOT MADLIB_SCHEMA.__is_miss_value(fval)';
-    ELSE
-        fval_txt  = 'CASE WHEN (MADLIB_SCHEMA.__is_miss_value(fval)) THEN
-                        NULL::TEXT
-                     ELSE 
-                        fval
-                     END ';    
-    END IF;
-    
     -- the supported missing value representation (' ', '?' and NULL) will
     -- be replace with NULL for easy processing later.
     -- the function __to_char is needed because on some databases an explicit
     -- cast to text is unavailable.
+    IF (h2hmv_routine_id = 1) THEN
+        where_txt = ' WHERE NULLIF(NULLIF(btrim(fval, '' ''), ''?''), '''') IS NOT NULL';
+    ELSE
+        fval_txt  = ' NULLIF(NULLIF(btrim(fval, '' ''), ''?''), '''')';
+    END IF;
+    
     IF (cls_col_name IS NULL) THEN
         -- if the kv_cls_name is null, then the class column will be null        
         curstmt = MADLIB_SCHEMA.__format
--- a/methods/cart/src/pg_gp/dt_preproc.sql_in
+++ b/methods/cart/src/pg_gp/dt_preproc.sql_in
@@ -3,7 +3,9 @@
  * @file dt_preproc.sql_in
  *
  * @brief Functions used in C4.5 and random forest for data preprocessing.
- * @date April 5, 2012
+ *
+ * @create    April 5, 2012
+ * @modified  July 19, 2012
  *
  *//* ----------------------------------------------------------------------- */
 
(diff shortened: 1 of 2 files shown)


methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -285,48 +561,83 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.__insert_into_metatable
...
+                WHERE table_oid IS NOT NULL
+                GROUP BY table_oid
+                ) t',
+             ARRAY[
+                 meta_tbl_name
+             ]
+        );
+    
+    -- drop all the Key-Value tables
+    FOR name IN EXECUTE curstmt LOOP
```
haradh1: Why does it need FOR loop?  It's simply

``` sql
SELECT madlib.__assert_table(table_name, 't')
FROM(
  SELECT madlib.__regclass_to_text(table_oid) AS table_name
  FROM ...
)s;
```

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -1453,254 +1447,423 @@ $$ LANGUAGE PLPGSQL;
...
+m4_ifdef(`__HAS_ORDERED_AGGREGATES__', >>>
+        curstmt = MADLIB_SCHEMA.__format
             (
-                'CREATE TABLE % (%, key)  
-                 AS SELECT DISTINCT %, 1 FROM %   
-                 m4_ifdef(`__GREENPLUM__', `DISTRIBUTED BY(%)');', 
+                'SELECT array_agg(attname ORDER BY attname) as attnames 
+                 FROM   pg_attribute 
+                 WHERE  attrelid = ''%''::regclass and attnum > 0 AND 
+                        attname <> ''%'' AND
```
haradh1, line 1489: Redundant.

haradh1, line 1489: Oh, I see.  attname not in (id_col_name, cls_col_name)

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -1453,254 +1447,423 @@ $$ LANGUAGE PLPGSQL;
...
+    exec_begin = clock_timestamp();
+    
+    -- create the KV table for the class column.
+    EXECUTE 'DROP TABLE IF EXISTS ' || kv_cls_name;
+    curstmt = MADLIB_SCHEMA.__format
+            (
+                'CREATE TABLE % AS
+                    SELECT 0 as fid, 
+                        MADLIB_SCHEMA.__to_char(%) AS fval, 
+                        rank() OVER (ORDER BY %) AS code
```
haradh1, line 1541: Are you aware this gets into master and is not scalable?

methods/cart/src/pg_gp/dt_utility.sql_in
```diff
@@ -528,3 +507,52 @@ BEGIN
...
+	src_arr ANYARRAY, 
+	tst_arr ANYARRAY
+	)
+RETURNS BOOLEAN[] AS $$
+    SELECT array_agg(elem in (SELECT unnest($1)))
+    FROM
+       (
+       SELECT unnest($2) as elem
+       ) t
+$$ LANGUAGE sql STABLE;
```
haradh1: Why stable?  It looks like IMMUTABLE.

methods/cart/src/pg_gp/dt_utility.sql_in
```diff
@@ -528,3 +507,52 @@ BEGIN
     RETURN ret;
 END
 $$ LANGUAGE PLPGSQL;
+
+
+/*
+ * @brief Retrieve an BOOL array. The ith element in the array
```
haradh1: s/an BOOL/a BOOL/

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -825,627 +992,454 @@ $$ LANGUAGE PLPGSQL;
...
+    -- Generating the encoded table through join the breakup table with
+    -- the KV table for all the features
+    curstmt = MADLIB_SCHEMA.__format
         (
-            id_column_name,
-            metatable_name,
-            'f'
+            'INSERT INTO %(id, fid, fval, is_cont, class)
+             SELECT p.id AS id, p.fid AS fid,
+                 CASE WHEN (p.is_cont AND p.fval IS NOT NULL) THEN
```
haradh1, line 1240: Is this true??   If it's not continuous value but its value is NULL, what should be m.code?

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -825,627 +992,454 @@ $$ LANGUAGE PLPGSQL;
...
+            'CREATE TEMP TABLE %
+             (
+             id      BIGINT ENCODING (compresstype=RLE_TYPE),
+             fid     INT    ENCODING (compresstype=RLE_TYPE),
+             fval    TEXT   ENCODING (compresstype=RLE_TYPE),
+             is_cont BOOL   ENCODING (compresstype=RLE_TYPE),
+             class   INT    ENCODING (compresstype=RLE_TYPE)
+             )
+             WITH(appendonly=true, orientation=column)
+             DISTRIBUTED BY(id)',
```
haradh1, line 1046: DISTRIBUTED BY (fid) may be better, as this table will be joined on it.

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -825,627 +992,454 @@ $$ LANGUAGE PLPGSQL;
...
-    PERFORM MADLIB_SCHEMA.__assert
+    EXECUTE 'DROP TABLE IF EXISTS ' || hor_tbl_name;
+    curstmt = MADLIB_SCHEMA.__format
         (
-            length(MADLIB_SCHEMA.__strip_schema_name(encoded_table_name)) <= 63, 
-            'The maximum length of ' || encoded_table_name || ' is 63'
+            'CREATE TEMP TABLE %(id, fvals, class) AS
+             SELECT
+                 id,
+                 madlib.__array_indexed_agg(fval, %, fid) as fvals,
```
haradh1: s/madlib/MADLIB_SCHEMA/

haradh1: Was there __array_indexed_agg?

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -825,627 +992,454 @@ $$ LANGUAGE PLPGSQL;
...
+	curstmt = MADLIB_SCHEMA.__format
+		(
+			'CREATE TABLE %(fid, fval, code) AS
+		     SELECT
+				fid,
+				fval,
+				(rank() OVER (PARTITION BY fid ORDER BY fval))::FLOAT8 AS code
+			 FROM tmp_dist_table
+			 WHERE (NOT is_cont)
+			 m4_ifdef(`__GREENPLUM__', `DISTRIBUTED BY (fid, fval)')',
```
haradh1, line 1385: Do you think fval is necessary in DISTRIBUTED BY?  If breakup table is distributed by fid, and as the join condition I saw below is COALESCE(...), only fid may make more sense.

methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -1453,254 +1447,423 @@ $$ LANGUAGE PLPGSQL;
...
+            id_col_name,
+            cls_col_name,
+            attr_col_names,
+            is_conts,
+            2,
+            verbosity
+        );  
+    
+    -- generate the vertical encoded table.
+    PERFORM MADLIB_SCHEMA.__gen_vertical_encoded_table
```
haradh1, line 1820: Do we still need both of vertical and horizontal encoded tables?

haradh1:
Looks good overall for the objective.  It occurred to me that it might be better to split encoding method and DT/RF code completely, so that the intermediate tables can be seen by the users.  I guess it's convenient to have the DT/RF function do the encoding inside it, but sometimes it may not be necessary.  From the design perspective, encoding operation is orthogonal to the ML algorithm and may be applicable to other modules in the future.

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang: Thanks. I will use the following SQL statement. 

``` javascript
    curstmt = MADLIB_SCHEMA.__format
        (
            'SELECT MADLIB_SCHEMA.__assert_table
                (MADLIB_SCHEMA.__regclass_to_text(table_oid), ''t'') 
             FROM 
                (
                SELECT table_oid
                FROM %
                WHERE table_oid IS NOT NULL
                GROUP BY table_oid
                ) t',
             ARRAY[
                 meta_tbl_name
             ]
        );
    EXECUTE curstmt;

```

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang, line 1541: Yes, I did. We need to use such function (rank or row_number) to generate the UNIQUE CONTINUOUS NUMBERs. And the time for this operation (generating KV tables) is less than 20% of the total encoding time. Please see the testing results from connect and internet_ads. Therefore, i think it's acceptable here. If you have better solution, please let me know. Thanks.

``` sql
dataset           breakup table     generating KV tables     generating vertical table   
connect          00:00:05.532092     00:00:01.163905          00:00:04.470939   
internet_ads     00:00:08.114627     00:00:02.594773          00:00:05.439457    
```

haradh1, line 1541: Great.

methods/cart/src/pg_gp/dt_utility.sql_in
aojwang: Done.

methods/cart/src/pg_gp/dt_utility.sql_in
aojwang: Right. Thanks.  Add Immutable/stable to all the functions in this file. Please help check if they are right. THanks.

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang, line 1240: Yes.
In explicit mode, for a continuous feature, null will be encoded as the average value of this feature; for a discrete attribute, null will be encoded as another kind of distinct value. For example,  the 1st feature (fid=1) is a continuous feature, its average value is 25.66; the 2nd feature (fid=2) is a discrete one, it has three distinct null value, then null value will be encoded to 4. Therefore, they will occupies two rows in the KV table for attributes. Please see the following table. At this point,  **If it's not continuous value but its value is NULL**, then it should be m.code.

The KV table for attributes.

``` sql
fid fval  code
1  null  25.66
2  null   4
```

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang, line 1046: Two reasons:
1) Using "fid" as distribution key will cause data redistribution, since the input table is distributed by id. Especially for the big dataset (such as kddcup, it has about 5M rows, and 40 columns), the redistribution is very expensive according to our test. However, using "id" as distribution key, there is no data redistribution.

2) If the number of columns of the input table is less than the number of segments of GPDB, then the data can't be distributed among all segments.

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang: Thanks. Done.
This aggregate will be in dt.c, which will be in another PR. I will let you know the exact PR number if we raised it.
It's very similar with the array_agg. However, we will use the fid as the subscript of the result array, and use the fval as the array element. The definition of this aggregate is as follows.

``` sql
/*
 * @brief The aggregate is the same with array_agg, which will accumulate
 *        The elements in each group to an array, except that we allow users 
 *        provide the subscript for each element. This aggregate will be 
 *        invoked as HashAggregate, while array_agg will be called as 
 *        GroupAggregate. Therefore, our implementation have better performance
 *        than the array_agg.
 * 
 * @param elem     The element to be fed into the returned array of this aggregate.
 * @param elem_cnt The number of elements.
 * @param elem_idx The subscript of the element.
 *
 * @return The aggregated array.
 *
 */
CREATE AGGREGATE MADLIB_SCHEMA.__array_indexed_agg(float8, int8, int8) (
    SFUNC = MADLIB_SCHEMA.__array_indexed_agg_sfun,
    m4_ifdef( `GREENPLUM',`PREFUNC   = MADLIB_SCHEMA.__array_indexed_agg_pfun,')
    FINALFUNC = MADLIB_SCHEMA.__array_indexed_agg_ffun, 
    STYPE = float8[]
);
```

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang, line 1820: No need this table for classification. So you can see that we will drop the vertical table after we generated the horizontal table. Thanks.

``` sql
    -- generate the horizontal encoded table.
    EXECUTE 'DROP TABLE IF EXISTS ' || enc_tbl_name;    
    PERFORM MADLIB_SCHEMA.__gen_horizontal_encoded_table
        (
            enc_tbl_name,
            'dt_tmp_ver_table',
            array_upper(is_conts, 1),
            verbosity
        ); 

    EXECUTE 'DROP TABLE IF EXISTS dt_tmp_ver_table';
```

methods/cart/src/pg_gp/dt_preproc.sql_in
aojwang, line 1385: As above said, breakup table should not be distributed by fid. As you can see from the query plan of the join statement you mentioned. Compared to the breakup table, the KV table (kv_attr_name) here is very small. Therefore, this table will be broadcasted according to the plan. If the number of attribute is less than the number of segments, then the data will be clustered to some of those segments and only those segments will participate in the broadcast motion. The following code shows the query plans. Thanks.

The table test_connect_dt_ed_col is distributed by (fid, fval)

``` sql
testdb=# explain analyze INSERT INTO test.test_connect_dt_ed2(id, fid, fval, is_cont, class)
testdb-#              SELECT p.id AS id, p.fid AS fid,
testdb-#                  CASE WHEN (p.is_cont AND p.fval IS NOT NULL) THEN
testdb-#                     p.fval::FLOAT8
testdb-#                  ELSE 
testdb-#                     m.code::FLOAT8 
testdb-#                  END AS fval,
testdb-#                  p.is_cont AS is_cont,
testdb-#                  p.class::INT AS class
testdb-#             FROM
testdb-#                 tmp_breakup_table p LEFT JOIN test.test_connect_dt_ed_col m
testdb-#             ON
testdb-#                 m.fid = p.fid AND
testdb-#                 (coalesce(m.fval, '') = (coalesce(p.fval, '')));
                                                                             QUERY PLAN                                                                             
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Insert (slice0; segments: 4)  (rows=10971 width=57)
   ->  Hash Left Join  (cost=54.04..1367.24 rows=10971 width=57)
         Hash Cond: p.fid = m.fid AND COALESCE(p.fval, ''::text) = COALESCE(m.fval, ''::text)
         Rows out:  Avg 709348.5 rows x 4 workers.  Max 709380 rows (seg2) with 7.007 ms to first row, 1464 ms to end, start offset by 25 ms.
         Executor memory:  4K bytes avg, 4K bytes max (seg0).
         Work_mem used:  4K bytes avg, 4K bytes max (seg0).
         (seg2)   Hash chain length 1.0 avg, 1 max, using 126 of 262151 buckets.
         ->  Append-only Columnar Scan on tmp_breakup_table p  (cost=0.00..544.84 rows=10971 width=49)
               Rows out:  Avg 709348.5 rows x 4 workers.  Max 709380 rows (seg2) with 0.464 ms to first row, 455 ms to end, start offset by 31 ms.
         ->  Hash  (cost=27.52..27.52 rows=442 width=44)
               Rows in:  Avg 126.0 rows x 4 workers.  Max 126 rows (seg0) with 7.227 ms to end, start offset by 24 ms.
               ->  Broadcast Motion 4:4  (slice1; segments: 4)  (cost=0.00..27.52 rows=442 width=44)
                     Rows out:  Avg 126.0 rows x 4 workers at destination.  Max 126 rows (seg0) with 4.988 ms to first row, 7.112 ms to end, start offset by 24 ms.
                     ->  Seq Scan on test_connect_dt_ed_col m  (cost=0.00..5.42 rows=111 width=44)
                           Rows out:  Avg 31.5 rows x 4 workers.  Max 32 rows (seg1) with 0.216 ms to first row, 0.245 ms to end, start offset by 30 ms.
 Slice statistics:
   (slice0)    Executor memory: 4640K bytes avg x 4 workers, 4640K bytes max (seg0).  Work_mem: 4K bytes max.
   (slice1)    Executor memory: 256K bytes avg x 4 workers, 256K bytes max (seg0).
 Statement statistics:
   Memory used: 128000K bytes
 Total runtime: 3448.242 ms
(21 rows)

```

PR #181, opened 2012-07-24
Top-level Makefile fix and Eigen issue

haradh1:
Once we get it in master, I'll backport them to 0.4.1.

Jira:MADLIB-589

It seems the wild card in Makefile rule doesn't work with fixed target
in some environment.  Also inlucde some pgxn package improvement (it will
be rewritten in cmake in the future.)

Jira:MADLIB-604

Eigen has been downloaded from Bitbucket, but now it seems it's better
to have it included in our source tree, as Bitbucket's subversion access
is not officially supported and there were a couple of reports that
it's not working these days.


fschopp:
Not sure if adding binary files to the git repository is a good idea.

haradh1:
Yeah, maybe we can mirror dependencies in our madlib.net or somewhere.  I initially thought about mirroring it in github, but git protocol is not graceful for us and it still needs HTTPS.

src/CMakeLists.txt
```diff
@@ -36,27 +36,22 @@ if(NOT BOOST_TAR_SOURCE)
```
fschopp, line 35: I'll ignore changes in this file and instead use pull request #183, OK?

haradh1, line 35: Sure.

PR #182, opened 2012-07-24
madlib-605: We write some C aggregations to improve the performance of C4.5/RF

renyi533:
We write some C aggregations to improve the performance of C4.5/RF. This task is related with these aggregations.

### DIFF
--- a/methods/cart/src/pg_gp/dt.c
+++ b/methods/cart/src/pg_gp/dt.c
@@ -14,9 +14,11 @@
 
 #include "postgres.h"
 #include "fmgr.h"
+#include "access/tupmacs.h"
 #include "utils/array.h"
 #include "utils/lsyscache.h"
 #include "utils/builtins.h"
+#include "utils/typcache.h"
 #include "catalog/pg_type.h"
 #include "catalog/namespace.h"
 #include "nodes/execnodes.h"
@@ -43,24 +45,26 @@ PG_MODULE_MAGIC;
 
 
 /*
- * This function is used to get the mask of the given value
- * val - ((val >> power) << power) equals to val % (2^power)
+ * This macro is used to get the mask bit of the given feature
+ * id. 
+ * fid - ((fid >> power) << power) equals to fid % (2^power)
  */
-#define dt_fid_mask(val, power) \
-			(1 << (val - ((val >> power) << power)))
+#define dt_fid_mask(fid, power) \
+			(1 << (fid - ((fid >> power) << power)))
 
 
 /*
- * We will do a lot of float number calculations on log, square, 
- * division, and multiplication, etc. Our tests show that the 
- * DBL_EPSILON is too precise for us. Therefore we define our 
- * own precision here.
+ * We use a lot of floating number operations during the training. 
+ * For these operations, DBL_EPSILON defined in float.h, leads to error
+ * add-up and wrong results. For our calculations, we need to redefine 
+ * that to a bigger number. Any floating number whose absolute value is 
+ * smaller than the one defined here will be treated as zero. 
  */
 #define DT_EPSILON 0.000000001
 
 /*
- * This function is used to test if a float value is 0.
- * Due to the precision of floating numbers, we can not
+ * This macro is used to test if a float value is 0.
+ * Due to the precision loss of floating numbers, we can not
  * compare them directly with 0.
  */
 #define dt_is_float_zero(value)  \
@@ -68,11 +72,26 @@ PG_MODULE_MAGIC;
 
 
 /*
+ * calculate the value of (val)log(val)
+ *
+ * @param   val     the value to be calculated
+ *
+ * NOTE: when x approximates 0, x*log(x) also approximates 0.
+ * Therefore, we directly return 0 when v is 0.
+ */
+#define dt_cal_log(v)  (dt_is_float_zero(v) ? 0.0 : (v) * log(v))
+
+#define dt_cal_sqr(v)  ((v) * (v))
+
+#define dt_cal_sqr_div(v1, v2)  (dt_is_float_zero(v2) ? \
+                                0.0 : ((v1) * (v1))/(v2))
+
+/*
  * For Error Based Pruning (EBP), we need to compute the additional errors
  * if the error rate increases to the upper limit of the confidence level.
  * The coefficient is the square of the number of standard deviations
  * corresponding to the selected confidence level.
- * (Taken from Documenta Geigy Scientific Tables (Sixth Edition),
+ * (Excerpt from Documenta Geigy Scientific Tables (Sixth Edition),
  * p185 (with modifications).)
  */
 static float8 DT_CONFIDENCE_LEVEL[] =
@@ -81,8 +100,8 @@ static float8 DT_CONFIDENCE_DEV[]   =
 				{4.0, 3.09, 2.58, 2.33, 1.65, 1.28, 0.84, 0.25, 0.00};
 
 
-#define MIN_DT_CONFIDENCE_LEVEL 0.001L
-#define MAX_DT_CONFIDENCE_LEVEL 100.0L
+#define MIN_DT_CONFIDENCE_LEVEL 0.001
+#define MAX_DT_CONFIDENCE_LEVEL 100.0
 
 
 #define dt_check_error_value(condition, message, value) \
@@ -107,11 +126,14 @@ static float8 DT_CONFIDENCE_DEV[]   =
 			} while (0)
 
 
+/* 
+ * a forward declaration. 
+ */ 
 static
 float8
-dt_ebp_calc_errors_internal
+dt_ebp_calc_additional_errors
 	(
-	float8 total_cases,
+	float8 total_samples,
 	float8 num_errors,
 	float8 conf_level,
 	float8 coeff
@@ -120,11 +142,10 @@ dt_ebp_calc_errors_internal
 
 /*
  * @brief Calculates the total errors used by Error Based Pruning (EBP).
- *        This will be wrapped as a plc function.
  *
- * @param total         The number of total cases represented by the node
+ * @param total         The number of total samples represented by the node
  *                      being processed.
- * @param probability   The probability to mis-classify cases represented
+ * @param probability   The probability to mis-classify samples represented
  *                      by the child nodes if they are pruned with EBP.
  * @param conf_level    A certainty factor to calculate the confidence limits
  *                      for the probability of error using the binomial theorem.
@@ -138,12 +159,12 @@ dt_ebp_calc_errors
 	PG_FUNCTION_ARGS
 	)
 {
-    float8 total_cases 	= PG_GETARG_FLOAT8(0);
-    float8 probability 	= PG_GETARG_FLOAT8(1);
-    float8 conf_level 	= PG_GETARG_FLOAT8(2);
-    float8 result 		= 1.0L;
-    float8 coeff 		= 0.0L;
-    unsigned int i 		= 0;
+    float8 total_samples 	= PG_GETARG_FLOAT8(0);
+    float8 probability 	    = PG_GETARG_FLOAT8(1);
+    float8 conf_level 	    = PG_GETARG_FLOAT8(2);
+    float8 result 		    = 1.0L;
+    float8 coeff 		    = 0.0L;
+    unsigned int i 		    = 0;
 
     if (!dt_is_float_zero(100 - conf_level))
     {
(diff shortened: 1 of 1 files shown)


methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+    if (rebuild_array)
+    {
+        /* construct a new array to keep the aggr states. */
+        pg_count_array =
+        	construct_array(
+        		(Datum *)count_array,
+                array_length,
+                INT8OID,
+                sizeof(int64),
+                true,
```
haradh1, line 2461: We shouldn't hard-code it, should we?

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+ *       The input value will be trimed before it is tested. 
+ *       We define this test simply because a lot of datasets
+ *       use these three representations for missing values before 
+ *       they are loaded into databases. After loading, if these
+ *       strings are not properly translated, the tables will contain 
+ *       '', '?' instead of NULL. 
+ *
+ */
+Datum
+dt_is_miss_value
```
haradh1: Hmm, wondering what it is for.  If the special character is needed, why isn't it only NULL?

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+			build_state.dvalues[iterator_idx]   = Float8GetDatum(1);
+		}
+
+		/* put the elem into the target slot */
+		build_state.dvalues[elem_idx << 1]       = elem;
+		build_state.dvalues[(elem_idx << 1) + 1] = 
+            Float8GetDatum(PG_ARGISNULL(1) ? 1 : 0);
+		lbs[0] = 1;
+
+		state = construct_md_array
```
haradh1: Is this actually construct_array?

KevinYL: Actually, construct_array calls construct_md_array internally.

methods/cart/src/pg_gp/dt.c
renyi533, line 2461: What do you mean here? Actually, the size of the array is not hard-coded. We calculate and store the value to array_length.

methods/cart/src/pg_gp/dt.c
renyi533: After investigation, we think "construct_array" cannot handle arrays with null values. Therefore, we directly use construct_md_array here.

methods/cart/src/pg_gp/dt.c
renyi533: Many original data sets form UCI or somewhere else use '?' or empty string to denote NULL values. During encoding, we need convert these values to NULL. 

This function is used to identify these fields with null values.

methods/cart/src/pg_gp/dt.c
aojwang, line 2461: I think Hitoshi means that we hard code "pass by value" to true.

Actually, we plan to fix that after the code review done. The reason why we don't fix it here is to separate the performance improvement and bug fixes, so that you can review them easily. If you prefer to fix them in this review, we will do that. Thanks.

renyi533, line 2461: Yes. Some other components, such as sparse vector, also have this issue. We have a separate dedicated issue number to resolve this.

We think we can submit another pull request to address this issue.

methods/cart/src/pg_gp/dt.c
haradh1: I don't see you put NULL elements in the array, as you designed it like array[2n] = value[n] and array[2n+1] = isnull[n].

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+    dt_check_error_value
+        (
+            elem_cnt > 0,
+            "array_size:%d should be bigger than zero",
+            elem_cnt
+        );
+
+    dt_check_error_value
+        (
+            elem_idx >= 0 && elem_idx < elem_cnt,
```
haradh1, line 2551: You have to check it with the actual array size, not the user input size.

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+            "the size of the two state array must be the same"
+        );
+
+	elem_cnt = (ARR_DIMS(arg0)[0]) >> 1;
+
+	for (iterator_idx = 0; iterator_idx < elem_cnt; iterator_idx++)
+	{
+        elem_idx = iterator_idx << 1;
+
+		if (0 == (int)((float8*)ARR_DATA_PTR(arg1))[elem_idx + 1])
```
haradh1, line 2681: Are we sure if arg0[elem_idx] is NULL?  I suggest adding "else" for this condition just in case.

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+        );
+
+    dt_check_error_value
+        (
+            ARR_DIMS(state)[0] == (elem_cnt << 1),
+            "The dimension of state array should be %d",
+            (elem_cnt << 1)
+        );
+
+	((float8*)ARR_DATA_PTR(state))[(elem_idx << 1)]     = DatumGetFloat8(elem);
```
haradh1, line 2610: Don't we care if the input has duplicate?  I guess you are overwriting the element specified by the subscript argument, but it is possible that the subscript has been assigned before.

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,482 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+        result = ('\0' == p[0]) || ('?' == p[0] && 1 ==  strlen(p));
+    }
+
+    PG_RETURN_BOOL(result);
+}
+PG_FUNCTION_INFO_V1(dt_is_miss_value);
+
+
+/*
+ * @brief The step function of the aggregate array_indexed_agg.
```
haradh1, line 2504: Can you explain up to how many elements did you try?  I believe it's not scalable with many of elements in an array, and that's why array_agg doesn't use HashAgg.

methods/cart/src/pg_gp/dt.c
KevinYL, line 2551: At this point we don't know whether the array has been allocated. Checked bellow in this function.

methods/cart/src/pg_gp/dt.c
KevinYL, line 2681: No else need to do. Just taken the non-null one.  And in this environment, the pre-steps make sure there is no duplicates in a given index. 

At the same time we can't detect conditions that a given index has two duplicate nulls.

methods/cart/src/pg_gp/dt.c
KevinYL: modified to use consturct_array

methods/cart/src/pg_gp/dt.c
KevinYL, line 2504: It's up to the number of attributes. The most number of elements we tested is about 1500. I want to know why this is not scalable. Using this method, the aggregate process is the same as using ArrayBuildState.

methods/cart/src/pg_gp/dt.c
KevinYL, line 2610: The pre-steps make sure that there is no duplicates.

methods/cart/src/pg_gp/dt.c
haradh1, line 2551: I'm suggesting adding the boundary check later than here.

methods/cart/src/pg_gp/dt.c
haradh1, line 2610: What is pre-steps?  It would look better if you added the check here to make sure the input is as you expected.

methods/cart/src/pg_gp/dt.c
haradh1: What about NULLIF()?

methods/cart/src/pg_gp/dt.c
KevinYL, line 2551: Please see line 2637

methods/cart/src/pg_gp/dt.c
renyi534, line 2610: In previous steps, we can guarantee the input data do not have duplicate index.

methods/cart/src/pg_gp/dt.c
aojwang: Done. use the following code.

``` sql
NULLIF(NULLIF(btrim(fval, ' '), '?'), '') as fval
```

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,438 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+	(
+	PG_FUNCTION_ARGS
+	)
+{
+    ArrayType *pg_count_array    = NULL;
+    int array_dim                = 0;
+    int *p_array_dim             = NULL;
+    int array_length             = 0;
+    int64 *count_array           = NULL;
+    int max_num_of_classes  	 = PG_GETARG_INT32(1);
```
cwelton, line 2373: And what if the first arg is null?

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,438 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+    int64  count                 = PG_ARGISNULL(2)?0:PG_GETARG_INT64(2);
+    int    class                 = PG_ARGISNULL(3)?0:PG_GETARG_INT32(3);
+    bool rebuild_array     		 = false;
+
+    dtelog(WARNING, "max_num_of_classes:%d, count:%d, class:%d",
```
cwelton: WARNING?

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,438 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+			"It must be in range from 0 to the number of classes",
+			class
+		);
+
+    /* test if the first argument (class count array) is null */
+    if (PG_ARGISNULL(0))
+    {
+    	/*
+    	 * We assume the maximum number of classes is limited (up to millions),
+    	 * so that the allocated array won't break our memory limitation.
```
cwelton, line 2399: You assume that, but don't check it?

methods/cart/src/pg_gp/dt.c
```diff
@@ -2126,3 +2339,438 @@ Datum table_exists(PG_FUNCTION_ARGS)
...
+    		(
+    			array_length == max_num_of_classes,
+    			"dt_acc_count_sfunc invalid array length: %d. "
+    			"The length of class count array must be "
+    			"equal to the total number classes",
+    			array_length
+    		);
+    }
+
+    count_array[class - 1] += count;
```
cwelton, line 2450: We have guaranteed that class >= 0.  However if class == 0 this will result in memory corruption.

methods/cart/src/pg_gp/dt.c
renyi534, line 2373: I will add more check here.

methods/cart/src/pg_gp/dt.c
renyi534: dtelog is a macro which is only present in debug build. In release build, we never enable this macro.

We can also replace WARNING with NOTICE.

methods/cart/src/pg_gp/dt.c
renyi534, line 2399: OK. I will check it.

methods/cart/src/pg_gp/dt.c
renyi533, line 2450: Good catch. We should ensure it is greater than 0 rather than greater than or equal to zero. I will change.

methods/cart/src/pg_gp/dt.c
renyi533: I removed the dtelog, which is only useful for debugging purpose.

PR #184, opened 2012-07-27
Performance Improvement on Tree Growing Engine

aojwang:
Jira: MADLIB-607
This file implements the tree growing engine for C4.5 and RF, which includes training, classification, scoring and displaying.
There are few changes on classification/scoring/displaying/clean. Most of changes are in training process.
The steps of training:
0. Encoding the training table.
1. Generating ACC (a group by statement).
2. Generating ACS and calculating SCVs and finding the best SCVs (in a single SQL statement)
3. creating the child nodes(a sigle SQL statement)
4. recursive the step 1 to step 3 until we met the stop condition.

Details are available at this [page](http://confluence.greenplum.com/x/WITaAg).

### DIFF
--- a/methods/cart/src/pg_gp/dt.sql_in
+++ b/methods/cart/src/pg_gp/dt.sql_in
@@ -25,7 +25,13 @@ m4_ifelse(
     `m4_define(`__POSTGRESQL_PRE_9_0__')'
 )
 
-
+m4_ifelse(
+    m4_eval(
+        m4_ifdef(`__GREENPLUM__', 1, 0) &&
+        __DBMS_VERSION_MAJOR__ * 100 + __DBMS_VERSION_MINOR__ >= 402
+    ), 1,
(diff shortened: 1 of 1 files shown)


aojwang:
It seems the GitHub can't display the difference. I will use fisheye later (the fisheye can't show my latest changes now).

aojwang:
Please review at this [page](http://fisheye.greenplum.com/cru/CR-5446) on fisheye.

haradh1:
It contains a lot of additional plpgsql functions based on existing dt/rf code.  I believe we should address the code integration issue now instead of investing a lot of effort to just extend existing code.  Why don't you look at other pull requests around C++ AL and pl/python abstraction.  If the discussion is slow for you, you can propose your idea and work on it.  Now we agree with our direction towards the abstraction layer, we should invest on that rather than duplicating code.

PR #185, opened 2012-07-27
Peformance Improvement: APIs of C4.5 and RF

aojwang:
Jira: MADLIB-608
For C4.5, there is no changes for the training/classification/scoring/genrules/displaying/clean.
For RF, we remove the max_split_points parameter in training API, since we doesn't need it any more.
Another change is ,that for both C4.5 and RF, we now can use an encoded table as training table.

### DIFF
--- a/methods/cart/src/pg_gp/c45.sql_in
+++ b/methods/cart/src/pg_gp/c45.sql_in
@@ -97,7 +97,7 @@ that it does not need a class column.
   This will create the decision tree output table storing an abstract object
   (representing the model) used for further classification. Column names:
   <pre>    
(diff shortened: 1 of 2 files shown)


haradh1:
Can you remind me why we don't max_split_point anymore?

aojwang:
The reason why we add the max_split_point is that the performance is not good in MADlib v0.4. Since the max_split_point will decrease the classification's correctness and now the performance looks good, we remove the parameter. Thanks.

haradh1:
During my merge work, I found the install-check fails due to the rf_train's argument changes.  Can you update it?

PR #191, opened 2012-08-06
Design doc: added Newton's to convex

haying:
1. A section for Newton's method is added
2. And minor fixes for notation consistency

### DIFF
--- a/doc/design/modules/convex-programming.tex
+++ b/doc/design/modules/convex-programming.tex
@@ -16,25 +16,25 @@ Driven by this, convex programming framework in MADlib also implements other alg
 \section{Introduction}
 % Problem definition. What are the problems that we can solve, formally and example applications?
 % linearly separable, unconstrained, continuous, deterministic, convex, minimization problems.
-This section is to first explain, formally, the type of problems that we consider in MADlib convex programming framework, and then give a few example modules.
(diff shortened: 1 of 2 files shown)


doc/design/modules/convex-programming.tex
```diff
@@ -21,20 +21,20 @@ \section{Introduction}
 \subsection{Formulation}
 We support numerical optimization problems with a objective function that is a sum of many component functions \cite{springerlink:10.1007/s10107-011-0472-0}, such as
 \[\min_{w \in \mathbb{R}^N} \sum_{m=1}^M f_{z_m}(w),\]
-where $z_m \in \mathcal{O}, m = 1,...,M,$ are observations, and $f_{z_m} : \mathbb{R}^N \mapsto \mathbb{R}$ are convex functions.
+where $z_m \in \mathcal{O}, m = 1,...,M$, are observations, and $f_{z_m} : \mathbb{R}^N \mapsto \mathbb{R}$ are convex functions.
```
fschopp: Convention is to use \to in the signature, and \mapsto for the definition.

haying: Cool. BTW, any reference that I can look up for these conventions?

fschopp: Good question (don't know without looking myself). But that's how all the math literature does it that I am aware of.

PR #193, opened 2012-08-08
Fix Performance issues with Naive Bayes

renyi533:
Jiras: http://jira.madlib.net/browse/MADLIB-611 and http://jira.madlib.net/browse/MADLIB-619
Also http://jira.madlib.net/browse/MADLIB-626

The changes are: 
1. replacing cross joins, which is very slow for training sets with a lot of features,  whenever possible. 
2. replacing union with outer join to improve performance. 
3. verify the supplied attrNum is consistent with the training set. 
4. replacing distinct with group by for better performance. 
5. remove costly window functions, which generates constant result for each record.

### DIFF
--- a/src/ports/postgres/modules/bayes/bayes.py_in
+++ b/src/ports/postgres/modules/bayes/bayes.py_in
@@ -385,11 +385,19 @@ def create_prepared_data(**kwargs):
     @param featureProbsDestName Name of feature-probabilities relation to create 
         
     """
-    if 'trainingSource' in kwargs:
-        __verify_attr_num( 
-            kwargs["trainingSource"], 
-            kwargs["trainingAttrColumn"],
-            kwargs["numAttrs"])
+    if not kwargs.has_key('numAttrs'):
+        plpy.error("'numAttrs' must be provided")
+    
+    if not kwargs.has_key('trainingSource'):
+        plpy.error("'trainingSource' must be provided")
+
+    if not kwargs.has_key('trainingAttrColumn'):
+        plpy.error("'trainingAttrColumn' must be provided")
+
+    __verify_attr_num( 
+        kwargs["trainingSource"], 
+        kwargs["trainingAttrColumn"],
+        kwargs["numAttrs"])
 
     if kwargs['whatToCreate'] == 'TABLE':
         # FIXME: ANALYZE is not portable.
@@ -506,11 +514,23 @@ def create_classification(**kwargs):
     
     __init_prepared_data(kwargs)
 
+    if kwargs.has_key('trainingSource') <> kwargs.has_key('trainingAttrColumn'):
+        plpy.error("'trainingSource' and 'trainingAttrColumn' must be provided together")
+
+    if not kwargs.has_key('numAttrs'):
+        plpy.error("'numAttrs' must be provided")
+    
     if 'trainingSource' in kwargs:
         __verify_attr_num( 
             kwargs["trainingSource"], 
             kwargs["trainingAttrColumn"],
             kwargs["numAttrs"])
+
+    if not kwargs.has_key('classifySource'):
+        plpy.error("'classifySource' must be provided")
+
+    if not kwargs.has_key('classifyAttrColumn'):
+        plpy.error("'classifyAttrColumn' must be provided")
 
     __verify_attr_num( 
         kwargs["classifySource"], 
@@ -605,11 +625,23 @@ where P(A = a | C = c) = ||  P(A_i = a_i | C = c).
 
     __init_prepared_data(kwargs)
 
+    if kwargs.has_key('trainingSource') <> kwargs.has_key('trainingAttrColumn'):
+        plpy.error("'trainingSource' and 'trainingAttrColumn' must be provided together")
+
+    if not kwargs.has_key('numAttrs'):
+        plpy.error("'numAttrs' must be provided")
+    
     if 'trainingSource' in kwargs:
         __verify_attr_num( 
             kwargs["trainingSource"], 
             kwargs["trainingAttrColumn"],
             kwargs["numAttrs"])
+
+    if not kwargs.has_key('classifySource'):
+        plpy.error("'classifySource' must be provided")
+
+    if not kwargs.has_key('classifyAttrColumn'):
+        plpy.error("'classifyAttrColumn' must be provided")
 
     __verify_attr_num( 
         kwargs["classifySource"], 
@@ -704,7 +736,7 @@ def __init_prepared_data(kwargs):
     a relation.
 
     """
-
+    
     if not 'classPriorsSource' in kwargs:
         kwargs.update(dict(
                 classPriorsSource = "(" + __get_class_priors_sql(**kwargs) + ")" 
@@ -722,12 +754,13 @@ def __verify_attr_num(sourceTable, attrColumn, numAttr):
     """
     Verify the number of attributes for each record is as expected.
     """
-    result = plpy.execute("SELECT count(*) size FROM "+sourceTable+
-        " WHERE array_upper("+attrColumn+", 1) <>"+str(numAttr))
-
-    if (result.nrows() <> 1):
-        plpy.error("error: source table is not of the expected form")
+    result = plpy.execute("""
+        SELECT count(*) size FROM {0}
+        WHERE array_upper({1}, 1) <> {2}
+        """.format(sourceTable, attrColumn, numAttr)
+        )
 
     dsize = result[0]['size']
     if (dsize <> 0):
-        plpy.error("error: Some records have wrong number of attributes")    
+        plpy.error("error: Some records in '{0}' have wrong number of attributes\
+            ".format(sourceTable))    
--- a/src/ports/postgres/modules/bayes/bayes.py_in
+++ b/src/ports/postgres/modules/bayes/bayes.py_in
@@ -762,5 +762,5 @@ def __verify_attr_num(sourceTable, attrColumn, numAttr):
 
     dsize = result[0]['size']
     if (dsize <> 0):
-        plpy.error("error: Some records in '{0}' have wrong number of attributes\
-            ".format(sourceTable))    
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -90,16 +90,15 @@ def __get_feature_probs_sql(**kwargs):
...
-                generate_series(1, {numAttrs}) AS attr,
                 {trainingSource} AS trainingSource
             GROUP BY
-                trainingSource.{trainingClassColumn},
-                attr.attr,
-                trainingSource.{trainingAttrColumn}[attr.attr]
+                class,
+                attr,
+                value
         ) AS triple_counts
```
cwelton, line 102: I'm surprised that this change would make a large difference in performance.

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -679,3 +717,17 @@ def __init_prepared_data(kwargs):
         kwargs.update(dict(
                 smoothingFactor = 1
             ))
+
+def __verify_attr_num(sourceTable, attrColumn, numAttr):
+    """
+    Verify the number of attributes for each record is as expected.
+    """
+    result = plpy.execute("SELECT count(*) size FROM "+sourceTable+
+        " WHERE array_upper("+attrColumn+", 1) <>"+str(numAttr))
```
cwelton: It would be cleaner to use "".format()

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -375,7 +385,12 @@ def create_prepared_data(**kwargs):
     @param featureProbsDestName Name of feature-probabilities relation to create 
         
     """
-    
+    if 'trainingSource' in kwargs:
+        __verify_attr_num( 
+            kwargs["trainingSource"], 
+            kwargs["trainingAttrColumn"],
+            kwargs["numAttrs"])
```
cwelton: You validate that 'trainingSource' is in the kwargs, but not 'trainingAttrColumn' or 'numAttrs' which can lead to:  `KeyError: 'trainingAttrColumn'` when the parameter is not set.

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -490,6 +505,18 @@ def create_classification(**kwargs):
...
+
+    __verify_attr_num( 
+        kwargs["classifySource"], 
+        kwargs["classifyAttrColumn"],
+        kwargs["numAttrs"])
```
cwelton, line 544: Same issue here

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -577,6 +604,18 @@ where P(A = a | C = c) = ||  P(A_i = a_i | C = c).
...
```
cwelton, line 655: And here.

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -679,3 +717,17 @@ def __init_prepared_data(kwargs):
         kwargs.update(dict(
                 smoothingFactor = 1
             ))
+
+def __verify_attr_num(sourceTable, attrColumn, numAttr):
```
cwelton, line 759: Do you see how in plpython we can create internal functions like this and _not_ expose them at the SQL layer?  This is a significant benefit from a programming perspective and is one of the advantages of using plpython as our implementation language instead of plpgsql.  It is also _much_ faster to call a function directly instead of having to go through an SPI function call.

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -679,3 +717,17 @@ def __init_prepared_data(kwargs):
...
+    """
+    Verify the number of attributes for each record is as expected.
+    """
+    result = plpy.execute("SELECT count(*) size FROM "+sourceTable+
+        " WHERE array_upper("+attrColumn+", 1) <>"+str(numAttr))
+
+    if (result.nrows() <> 1):
+        plpy.error("error: source table is not of the expected form")
```
cwelton: This error will be impossible to hit.  `SELECT count(*) FROM table` will ALWAYS return 1 row.

cwelton: Unless we hit an error from a malformed query, in which case checking nrows() is the wrong way of detecting the issue.

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -679,3 +717,17 @@ def __init_prepared_data(kwargs):
...
+
+    dsize = result[0]['size']
+    if (dsize <> 0):
+        plpy.error("error: Some records have wrong number of attributes")
```
cwelton: This is a terrible error message that provides no actionable information for the user.

cwelton:
Review completed.  Minor issues requiring update remain.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533, line 102: I added detailed query plans for both versions. You can study at http://jira.madlib.net/browse/MADLIB-619.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533: Accept.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533: If one of the parameters is present, the other two parameters should always be present. Otherwise, we need throw a error.

I will add more check and throw appropriate error messages here.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533, line 544: "classifySource" and "classifyAttrColumn" should be always present. But I agree it is fine to add more checking here.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533, line 759: Yes. With plpython, we can create real internal functions and hide them from users. This is really better than plpgsql. 

And the speed is faster. I agree.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533: If "sourceTable" or "attrColumn" does not exist, this line will be hit. 

Currently, we do not have the utilities of table_exist and column_exist. So I used this not so elegant way.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533: At this line, we don't know which record is of wrong attribute number. We just know how many records are wrong.

I added this general error message because it is costly for us to find which record is wrong. If most records are wrong for a large dataset, the data movement between python and gpdb will be large.

renyi533: For example, if we use the sql of:

select id from source_table where array_upper(attr_column, 1) <> num_attr.

This query can be potentially costly.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533: You are right. If the sql is wrong, the execution above will pop an error. As a result, I removed this check.

src/ports/postgres/modules/bayes/bayes.py_in
cwelton, line 102: Thanks, that is helpful.

src/ports/postgres/modules/bayes/bayes.py_in
cwelton: How about:

```
if dsize <> 0:
  plpy.error('found %d records where "%s" was not of expected length (%d)' % (dsize, attrColumn, numAttr))
```

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -375,7 +385,20 @@ def create_prepared_data(**kwargs):
     @param featureProbsDestName Name of feature-probabilities relation to create 
         
     """
+    if not kwargs.has_key('numAttrs'):
```
cwelton, line 394: I usually prefer this idiom:

```
if 'numAttrs' not in kwargs:
  ...
```

But your method works, so no real complaint from me here.

src/ports/postgres/modules/bayes/bayes.py_in
cwelton: Additional checks accepted.

cwelton:
Improve the error message and this looks fine to me.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533: Thanks. It is very reasonable. I have changed.

src/ports/postgres/modules/bayes/bayes.py_in
renyi533, line 394: Thanks for your understanding. As a C programmer, I always wonder what 'in' really do under the table. 'has_key' seems very safe and clear to me.

haradh1, line 394: "in" is more semantical.  This is not the case here, but if you have set or dict as the argument, "in" works for both.  Therefore, "in" is more preferred in general wherever available.

src/ports/postgres/modules/bayes/bayes.py_in
haradh1, line 102: I wonder if materializing the training source with unnest would be more efficient if you use it multiple times.

renyi533, line 102: From CROSS JOIN to UNNEST, we gained a speed up of up to 100 times. But, overall, we can only see speed up of less than 10 times. 

After this change, the bottleneck is not in the queries of 'UNNEST'. So materializing its result is not very valuable.

PR #194, opened 2012-08-08
Fixed documentation for hypothesis tests

fschopp:
Jira: MADLIB-624

Commit 884ff80 introduced a regression in that the three ordered-aggregate tests
{ks|mw|wsr}_test did no longer show up in the documentation. This has been
fixed. Also, all SQL macros that are otherwise defined in "SQLCommon.m4" are now
also define when running doxygen. This should increase the compatibility in the
future.

Biggest change: I moved the CMake macro 'define_m4_macros' from
PostgreSQLUtils.cmake to Utils.cmake.

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -1,3 +1,53 @@
+# Define GNU M4 preprocessor macros
+#
+# Generate M4 code and M4 command-line arguments to define a set of macros that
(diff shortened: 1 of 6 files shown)


cwelton:
This is fine, but I don't want to have to trigger a new RC because of this.

PR #195, opened 2012-08-09
Cox prop pull request

srikris:
http://jira.madlib.net/browse/MADLIB-576
Added the cox-proportional hazards module.


src/CMakeLists.txt
```diff
@@ -42,7 +42,10 @@ endif (NOT BOOST_TAR_SOURCE)
 # -DEIGEN_TAR_SOURCE=/path/to/eigen-x.x.x.tar.gz
 
 set(EIGEN_VERSION "3.0.3")
-set(EIGEN_TAR_MD5 695f24be85c4fe957ce6a4dd11161f48)
+set(EIGEN_TAR_MD5 9a69b27398e2b7159190313e6b81926f)
```
haying, line 45: The download of eigen has been changed in madlib/madlib/master, could you rebase on it?

src/ports/postgres/modules/stats/cox_prop_hazards.sql_in
```diff
@@ -0,0 +1,302 @@
...
+ * @param depColumn Name of the dependant column measuring time of death
+ * @param maxNumIterations The maximum number of iterations
+ * @param optimizer The optimizer to use (either
+ *        <tt>'newton'</tt>/<tt>'newton'</tt> for the newton method
+ * @param precision The difference between log-likelihood values in successive
+ *        iterations that should indicate convergence. Note that a non-positive
+ *        value here disables the convergence criterion, and execution will only
+ *        stop after \c maxNumIterations iterations.
+ *
+ * @return TODO
```
haradh1: Can you fill it?

src/modules/stats/cox_prop_hazards.cpp
```diff
@@ -0,0 +1,316 @@
...
+    if (!state.hessian.is_finite() || !state.grad.is_finite())
+        throw NoSolutionFoundException("Over- or underflow in intermediate "
+            "calulation. Input data is likely of poor numerical condition.");
+
+    SymmetricPositiveDefiniteEigenDecomposition<Matrix> decomposition(
+        state.hessian, EigenvaluesOnly, ComputePseudoInverse);
+		Matrix inverse_of_hessian = decomposition.pseudoInverse();
+
+		// Fixed Step size of 1		
+		state.coef = state.coef - state.hessian.inverse()*state.grad;
```
haying: 1. Why using "state.hessian.inverse()" instead of "inverse_of_hessian"? 
2. I think we actually prefer to expose the stepsize instead of hard-coded. Do you have difficulty in doing it?

src/ports/postgres/modules/stats/test/cox_proportional_hazards.sql_in
```diff
@@ -0,0 +1,58 @@
...
+(21,1,2.42,4),
+(22,1,4.01,3),
+(23,1,4.91,2),
+(24,1,5,1);
+
+ 
+SELECT * from madlib.cox_prop_hazards(
+    'leukemia', 'ARRAY[grp, wbc]', 'timeDeath', 20, 'newton',  0.001
+);
+
```
haying, line 119: Could you check the expected output coefficients? Is that not converging because of the hard-coded stepsize?

src/modules/stats/cox_prop_hazards.cpp
```diff
@@ -0,0 +1,316 @@
...
+		
+		double xc = trans(state.coef)*x;
+		double s = std::exp(xc);
+		
+		state.S += s;
+		state.H += s*x;		
+		state.V += x * trans(x) * s;
+
+		state.grad += x - state.H/state.S;
+		state.hessian += (state.H * trans(state.H))/(state.S*state.S) - state.V/state.S;
```
haying: Why do you still need to compute hessian if you choose gradient descent?

srikris: @haying: Small typo on the comment. That should be Newton method.

src/ports/postgres/modules/stats/test/cox_proportional_hazards.sql_in
srikris, line 119: I have checked the coefficients. They are correct. Newton method with step length 1 works pretty well on most problems.

src/modules/stats/cox_prop_hazards.cpp
haying: I see. I did not get an update from your design document which said that hessian is not advisable for Cox. So, this means you somehow find a way to do it using Newton's method?

src/ports/postgres/modules/stats/test/cox_proportional_hazards.sql_in
haying, line 119: Could you add the assert statement here, similar to regress/test/linear.sql_in, to check the coefficients?

src/ports/postgres/modules/stats/cox_prop_hazards.py_in
```diff
@@ -0,0 +1,218 @@
...
+        DROP TABLE IF EXISTS _madlib_iterative_alg;
+        CREATE TEMPORARY TABLE _madlib_iterative_alg (
+            _madlib_iteration INTEGER PRIMARY KEY,
+            _madlib_state {stateType}
+        );
+				DROP TABLE IF EXISTS _cox_intermediate_state;
+				CREATE TEMPORARY TABLE _cox_intermediate_state (
+			    _cox_dependant_variable FLOAT8,
+			    _cox_iState {intermediateStateType}
+				) DISTRIBUTED BY (_cox_dependant_variable);
```
haradh1, line 109: You have to use m4 macro to avoid DISTRIBUTED BY clause in postgres.  I hit an error in PG8.4 install-check.

haradh1:
Merged.  In passing, I removed the eigen file, added m4 macros for ordered aggregates, and reverted eigen md5 value.

PR #196, opened 2012-08-12
Generate a readable output rules for Apriori

aojwang:
JIRA: [MADLIB-411](http://jira.madlib.net/browse/MADLIB-411) [MADLIB-451](http://jira.madlib.net/browse/MADLIB-451)
The old output wasn't readable, since it use sparse vector to represent the rules and contains some useless columns, such as hash_list, hash_key. As per discussion, for a rule A=>B, where A/B is the itemset, we will use an array to kept the original items in the itemset. For example, for rule i: (a, b) => (c), we will have two columns: one is pre, which stores {a, b}; the other is post, which stores {c}. Details please refer to my comments of the bug.

### DIFF
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
@@ -406,7 +406,7 @@ SET client_min_messages= warning;
   -- the auxiliary table to generate the readable rules
   EXECUTE 'DROP TABLE IF EXISTS assoc_rules_aux_tmp';
   EXECUTE 
-    'CREATE TABLE assoc_rules_aux_tmp 
+    'CREATE TEMP TABLE assoc_rules_aux_tmp 
     (
         ruleId      SERIAL,
         pre         MADLIB_SCHEMA.svec,
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
@@ -403,19 +403,35 @@ SET client_min_messages= warning;
   m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (hash_key)')  
   ';
 
--- Association rules output 
+  -- the auxiliary table to generate the readable rules
+  EXECUTE 'DROP TABLE IF EXISTS assoc_rules_aux_tmp';
+  EXECUTE 
+    'CREATE TABLE assoc_rules_aux_tmp 
+    (
+        ruleId      SERIAL,
+        pre         MADLIB_SCHEMA.svec,
+        post        MADLIB_SCHEMA.svec,
+        support     FLOAT8,
+        confidence  FLOAT8,
+        lift        FLOAT8,
+        conviction  FLOAT8
+    )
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -403,19 +403,35 @@ SET client_min_messages= warning;
   m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (hash_key)')  
   ';
 
--- Association rules output 
+  -- the auxiliary table to generate the readable rules
+  EXECUTE 'DROP TABLE IF EXISTS assoc_rules_aux_tmp';
+  EXECUTE 
+    'CREATE TABLE assoc_rules_aux_tmp
```
cwelton: Should either be a temporary table or go into the specified output schema.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -685,6 +701,39 @@ LOOP
...
+            (   
+                SELECT 
+                    ruleId,
+                    unnest(MADLIB_SCHEMA.svec_nonbase_positions(post, 0)) as post_id
+                FROM assoc_rules_aux_tmp
+            ) s1, assoc_prod_uniq s2
+         WHERE s1.post_id = s2.prod_id
+         GROUP BY ruleId
+         ) t3
+     WHERE t1.ruleId = t2.ruleId AND t1.ruleId = t3.ruleId';
```
cwelton, line 733: 1. Output theoretically has unstable sort order since the array_agg is not ordered.  This wouldn't produce wrong results, but could make testing more difficult.
2. Looks like we are using an svec as a compressed bitmap here.  I'd have to look at the rest of the AR code, but if we're treating the svec as a bitmap universally then we have room for an optimization.

cwelton:
The temporary table is an issue that needs to be fixed prior to pull.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang: Should be a temp table. Done. Thanks.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang, line 733: Thanks.
1. As per discussion, QA can sort the pre or post column to get the stable result for each running.
2. Yes. And the performance of compressed bitmap will be better than svec. We will think about that.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -403,19 +403,35 @@ SET client_min_messages= warning;
...
+    ( 
+        ruleId      INT,
+        pre         TEXT[],
+        post        TEXT[],
+        support     FLOAT8,
+        confidence  FLOAT8,
+        lift        FLOAT8,
+        conviction  FLOAT8
+    )
+  m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (ruleId)')';
```
haradh1, line 434: What is the difference between __GREENPLUM__ and GREENPLUM?  I think I saw __GREENPLUM__ in dt code.

PR #197, opened 2012-08-14
Added ridge regression, LASSO, linear SVM and logit

haying:
This pull request demonstrates how to add new modules based on generic convex module.

Again, over 7k lines are for a data set, that still leaves LOC ~ 4k, roughly 1k for each module. Modules that have more than 1 optimizers may have more LOC. Duplicate code is largely on defining SQL functions and transition states.

Review commit by commit is suggested.

Adding a baseline logistic regression based on linear svm is no much more than "sed 's/linear_svm/logit/g'. I have it done simply because it is too easy... It should offer some information to understand the generic convex module. If we really do not like to have duplicate logistic regress implementation, this can be ignored when pushing to madlib/madlib. But I suggest to keep it in madlib/madlib using a different SQL function name, since it seems the best way we (for now, Florian and I) can keep track of it.


haying:
Move Caleb's comment here, loss after rebasing:
Currently we have the following functions:

```
\df madlib.l?norm
                           List of functions
 Schema |  Name  | Result data type |   Argument data types    |  Type  
--------+--------+------------------+--------------------------+--------
 madlib | l1norm | double precision | madlib.svec, madlib.svec | normal
 madlib | l2norm | double precision | madlib.svec, madlib.svec | normal
(2 rows)
```

I'm fine with either 'l1norm' or 'l1_norm', but would like MADlib to be consistent on the topic.

Overall I think l1_norm is a little cleaner, but its counter to the existing precedent.

haying:
Two TODOs: 1. change install-check to use plpgsql for storing cross-query model id; 2. fix user doxygen doc

haying:
Three things done by the commits (a) "Use plpgsql to store cross-query info", (b) "Fix doxygen for linear SVM, ridge, LASSO": 1. fix install-check by not using INSERT INTO ... AS; 2. remove the l1_norm; 3. added doxygen. No outstanding issues that I am aware of for now.

cwelton:
Currently deferring until next available release due to lack of resources to review.

src/ports/postgres/modules/regress/linear.sql_in
```diff
@@ -48,14 +48,14 @@ Let
 
 Maximizing the likelihood is equivalent to maximizing the log-likelihood
 \f$ \sum_{i=1}^n \log f(y_i \mid \boldsymbol x_i) \f$, which simplifies to
-maximizing the <b>residual sum of squares</b> \f$ RSS \f$ (also called sum of
+minimizing the <b>residual sum of squares</b> \f$ RSS \f$ (also called sum of
```
haradh1, line 51: Hm, that's nice.  Is this something we should pull separately for this release?

src/ports/postgres/modules/convex/logit_newton.py_in
```diff
@@ -0,0 +1,67 @@
```
haradh1, line 0: I thought this and igd were to be integrated into existing logistic.py_in.  Isn't it possible?

haying:
For those who may review this PR: I tried to rebase my branch on the current master, but the commits of new kmeans break the install-check of this convex module ("madpack install-check -t convex" works before rebasing; not after). The bug is not clear to me yet. First guess is that this is related to the python drivers.

cwelton:
Status:
- Ridge Regression & Lasso Regression largely subsumed by the Elastic Net work, but there might be something to keep.  Needs evaluation.
- Linear SVM is something that can be worthwhile.

Delaying resolution of this PR until we can get additional input from Aaron.

PR #198, opened 2012-08-15
Compatibility: INSERT INTO ... SELECT

haying:
Removed temp table for storing returned id, use plpgsql local variable instead

### DIFF
--- a/src/ports/postgres/modules/convex/test/lmf.sql_in
+++ b/src/ports/postgres/modules/convex/test/lmf.sql_in
@@ -100016,28 +100016,38 @@ COPY mlens100k (user_id, movie_id, rating) FROM STDIN;
 12	203	3
 \.
 
-DROP TABLE IF EXISTS _temp_model_id CASCADE;
-CREATE TEMPORARY TABLE _temp_model_id (id INTEGER);
+-- GPDB < 4.2.1 has issues of executing "CREATE TABLE .. AS SELECT" or 
+-- "INSERT INTO ... AS SELECT"
+-- To be compatiable with historical versions of GPDB, we use plpgsql 
+-- local variables instead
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/convex/test/lmf.sql_in
```diff
@@ -100016,28 +100016,34 @@ COPY mlens100k (user_id, movie_id, rating) FROM STDIN;
 12	203	3
 \.
 
-DROP TABLE IF EXISTS _temp_model_id CASCADE;
-CREATE TEMPORARY TABLE _temp_model_id (id INTEGER);
+CREATE FUNCTION check_rmse()
```
haradh1: Can you put a comment to explain why we need a function?

haying: Forgot to check it in. Fixed

cwelton:
This raises some longer term questions for me around model management.  

Linear regression:  We return the results directly, can be used as an aggregate.
Logistic regression: We return the results directly, but cannot be used as an aggregate.
K-means: Results written out to a user specified output table
Association Rules: Results written to system created tables in a user specified output schema
Decision Trees: Results written to user specified table with system generated auxiliary tables and tracked in a global model table by the user specified id.
SVD: Results written to a user specified output table returning a system generated id that references the model.

Which is to say... we have no consistency in terms of expected API.

haying:
I noticed this. Apparently we do not have enough guidelines for developers currently. My feeling is that some developer guidelines on this is needed, although some constraints like most main modules have to call a fixed set of API or SQL macros to produce user output could be better.

cwelton:
Agreed.  That was what I was trying to kick off early this release process, but it fell a little flat.  We'll have to take another stab at it in the next release.

PR #199, opened 2012-08-16
add linear-chain conditional random field for NLP

livingstream:
Dear all,
          I finished the linear-chain crf learning for natural language processing. However, documentation and unit tests are missing, I will work on that. Thanks.


src/modules/crf/linear_crf.cpp
```diff
@@ -0,0 +1,840 @@
...
+    double stp1, ftol, stp, sq, yr, beta;
+    int  iflag, iter, nfun, point, ispt, iypt, maxfev, info, bound, npt, cp, nfev, inmc, iycn, iscn;
+    //shared varibles in mcscrch
+    int infoc;
+    double dg, dgm, dginit, dgtest, dgx, dgxm, dgy, dgym, finit, ftest1, fm, fx, fxm, fy, fym, p5, p66, stx, sty, stmin, stmax, width, width1, xtrapf;
+    bool brackt, stage1, finish;
+
+    Eigen::VectorXd w;
+    Eigen::VectorXd x;
+    Eigen::VectorXd diag;
```
haying: Since you have been "using namespace dbal::eigen_integration", "ColumnVector" instead of "Eigen::VectorXd" is preferred. Also "Matrix" instead of "Eigen::MatrixXd".

src/modules/crf/linear_crf.cpp
```diff
@@ -0,0 +1,840 @@
...
+    typename HandleTraits<Handle>::ColumnVectorTransparentHandleMap lbfgs_state;
+    typename HandleTraits<Handle>::ColumnVectorTransparentHandleMap mcsrch_state;
+};
+
+class LBFGS {
+    //shared variables in the lbfgs
+public:
+    //shared variable in lbfgs
+    double stp1, ftol, stp, sq, yr, beta;
+    int  iflag, iter, nfun, point, ispt, iypt, maxfev, info, bound, npt, cp, nfev, inmc, iycn, iscn;
```
haying, line 172: Comment to explain these variables is needed.

src/modules/crf/linear_crf.cpp
```diff
@@ -0,0 +1,840 @@
...
+
+void compute_log_Mi(int num_labels, Eigen::MatrixXd &Mi, Eigen::VectorXd &Vi) {
+    // take exponential operator
+    for (int m = 0; m < num_labels; m++) {
+        Vi(m) = std::exp(Vi(m));
+        for (int n = 0; n < num_labels; n++) {
+            Mi(m, n) = std::exp(Mi(m, n));
+        }
+    }
+}
```
haying, line 659: This function seems to compute exp_Mi instead of log_Mi which is the name of this function. A name that describes what this function is actually doing would be helpful.

haying:
I have 2 high-level comments:
1. test/crf.sql_in needs a complete test case to verify the correctness of the model trained by this implementation: at least if you use the trained model to perform Viterbi inference on the training data itself, you should get good prediction results, i.e., small train error. An example can be found here: http://hazy.cs.wisc.edu/hazy/victor/demo/bismarck_crf.php
2. It is good that you follow the way we program an iterative algorithm in MADlib. But I would find it helpful if you put your implementation logic in a more abstract way. I try to do a little bit it for you here, please see whether it matches what you actually coded.
We code an iterative module in MADlib like this:

```
interState = Start(args)
Repeat
    In parallel for each segment:   
        intraState = Initialize(interState)
        For each row
            intraState = Transit(intraState, row)
    For each intraState:
        intraState = Merge(oldIntraState, intraState)
    interState = Finalize(intraState)
Until Converged(interState) 
Return End(interState)
```

In your LBFGS, my guess is that you have been computing the "gradient" in the Transit() function, and updating the LBFGS state using the new gradient in the Finalize() function. If you can abstract out the many lines of code in the lincrf_lbfgs_step_transition::run() as a function that takes (state, tuple) and output gradient and some other necessary things (perhaps normalization factor Z). That could make the code a lot cleaner. If you do not want to make them a separate function, please add some comment to say which block of code is to compute the gradient.

For now, my understanding of the code is this, but I guess it is not entirely correct.

```
Transit(state, tuple) {
    state.Z = crf_fwdbkwd(state.coef, tuple);
    state.gradient = crf_gradient(state, tuple);
    state.loglik = crf_gradient(state.Z);
}

FInalize(state) {
    state.lbfgs_state = LBFGS::lbfgs(state);
}
```

haying:
@livingstream Are you done with your update and ready for review?

src/modules/crf/linear_crf.cpp
livingstream, line 659: Done! compute_log_Mi is changed to compute_exp_Mi. Thanks.

livingstream:
Aaron, The code to compute loglikelihood,gradient is hooked up 
together. Cannot be separated.
So I abstract out all the code to compute the normalizer, 
log-likelihood,gradient using a function "compute_logli_graident"

src/modules/crf/linear_crf.cpp
livingstream: Do I need to replace all the Eigen vector, matrix to columvector and 
Matrix?

livingstream:
Aaron, some of your reviews are addressed. I am working on the left 
comments such as crf learning unit test and documentation/code comments.

src/modules/crf/linear_crf.cpp
haying: better than not, but not a "have-to". :)

haying:
Thanks! @livingstream

livingstream:
Aaron, I finished the unit test for the linear-chain crf learning. Let me know if you have more comments/suggestions. @haying @cwelton

cwelton:
Current plan is to stabilize the current (v0.5) release without this functionality and include it within the next release so that we can give it proper review and testing.

livingstream:
@cwelton, Caleb, we only have some documentation left and I already verified the correctness the crf by comparing the test results with other crf java package. Would we catch up the v0.5 release?

haying:
Although I won't be able to get a line-by-line review to verify the correctness of the code due to lack of design document, I should be able to verify the install-check some time today. I think whether this will go in for v0.5 may depend on whether we have immediate user demand for it?

cwelton:
Aaron and Grace are examining the code and doing some preliminary tests.  I will work with QA to see if we can fit it into the schedule.  I'll let you know progress as we go.  

Things that will help with testing and review: a) documentation, b) a design document.

src/ports/postgres/modules/crf/crf.sql_in
```diff
@@ -0,0 +1,236 @@
...
+    /* previous_state */ DOUBLE PRECISION[]) (
+    
+    STYPE=DOUBLE PRECISION[],
+    SFUNC=MADLIB_SCHEMA.lincrf_lbfgs_step_transition,
+    m4_ifdef(`GREENPLUM',`prefunc=MADLIB_SCHEMA.lincrf_lbfgs_step_merge_states,')
+    FINALFUNC=MADLIB_SCHEMA.lincrf_lbfgs_step_final,
+    INITCOND='{0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0}'
+);
+
+CREATE AGGREGATE MADLIB_SCHEMA.array_union(anyarray) (
```
haradh1, line 192: Look for src/ports/postgres/modules/stats/hypothesis_tests.sql_in for the GPDB keyword issue.

```
m4_ifdef(<!__HAS_ORDERED_AGGREGATES__!>,<!
CREATE
m4_ifdef(<!__GREENPLUM__!>,<!ORDERED!>)
AGGREGATE MADLIB_SCHEMA.ks_test(
```

Make sure install-check doesn't run it if ORDERED aggregate is not supported.

src/ports/postgres/modules/crf/test/crf_train.sql_in
```diff
@@ -0,0 +1,142 @@
...
+        CREATE TABLE train_featureset(f_index integer, f_name text, feature integer[]);
+
+        PERFORM MADLIB_SCHEMA.crf_train_fgen('train_segmenttbl', 'train_regex', 'train_dictionary', 'train_featuretbl','train_featureset');
+
+        CREATE TABLE train_crf_feature (id integer,name text,prev_label_id integer,label_id integer,weight float);
+
+        PERFORM MADLIB_SCHEMA.lincrf('train_featuretbl','sparse_r','dense_m','sparse_m','f_size',45, 'train_featureset','train_crf_feature', 20);
+
+        -- Expected feature table
+        CREATE TABLE expected_crf_feature(id integer,name text,prev_label integer,label integer,weight float);
```
geeg, line 64: Hi Kun, Where did these values come from?  Since they are used for the accuracy check, it would be important to document.

src/ports/postgres/modules/crf/crf_feature_gen.sql_in
```diff
@@ -77,11 +74,152 @@ label sequence, we only need a single lookup to get the feature weight.
...
+                           array_cat(fset.feature, ARRAY[f_index,start_pos,1])
+                    FROM   tmp1_feature, """ + featureset + """ fset
+                    WHERE  start_pos > 0 AND tmp1_feature.f_name = fset.f_name AND tmp1_feature.feature = fset.feature AND fset.f_name = 'E.';""")
+
+    plpy.execute("""INSERT INTO dense_mtbl(doc_id, dense_m)
+                    SELECT doc_id, MADLIB_SCHEMA.array_union(feature::integer[] order by start_pos)
+                    FROM   tmp_dense_mtbl
+                    GROUP BY doc_id;""")
+
+    plpy.execute("""INSERT INTO sparse_mtbl(sparse_m)
```
geeg, line 208: Hi Kun,

Just making sure that you intentionally meant to store all edge features (from all documents) into one single array.  This one array is duplicated in each row of featuretbl below (lines 213-216).

src/ports/postgres/modules/crf/test/crf_test.sql_in
```diff
@@ -22,16 +22,16 @@ DECLARE
...
         (1,E'Chancellor  of  the  Exchequer  Nigel  Lawson  ''s  restated  commitment  to  a  firm monetary  policy  has  helped  to  prevent  a  freefall  in  sterling  over  the  past  week  .'),
         (2,E'But  analysts  reckon  underlying  support  for  sterling  has  been  eroded  by  the  chancellor ''s  failure  to  announce  any  new  policy  measures  in  his  Mansion  House  speech  last  Thursday  .'),
         (3,E'His actions prevent disaster.');
-	analyze textfex_document;
+	analyze crf_document;
         
 	-- Features table
-	CREATE TABLE textfex_feature (id integer,name text,prev_label_id integer,label_id integer,weight float);
-        INSERT INTO textfex_feature VALUES
+	CREATE TABLE crf_feature_test (id integer,name text,prev_label_id integer,label_id integer,weight float);
```
geeg, line 33: Hi Kun,

Similar to my comment about the values of expected_crf_feature, could you explicitly document where these hard-coded values for crf_feature_test and expected_extraction came from since they are used to check against the MADlib crf output? Thanks!

livingstream, line 33: Hi Grace,

```
expected_extraction in the crf_test.sql_in and expected_crf_feature 
```

in the crf_train.sql_in are
produced from the Dr. Sunita's java CRF package 
http://crf.sourceforge.net/ with the same inputs.
I used these results as the correct result to compare with the our 
results.

livingstream, line 33: I just add some comments in the code about the standard output.

src/ports/postgres/modules/crf/crf.sql_in
livingstream, line 192: Hi Hitoshi, something like the following? But it looks like that the postgres cannot find this aggregate to use.

m4_ifdef(<!**HAS_ORDERED_AGGREGATES**!>,<!
CREATE
m4_ifdef(<!**GREENPLUM**!>,<!ORDERED!>)
AGGREGATE MADLIB_SCHEMA.array_union(anyarray) (
    SFUNC = array_cat,
    STYPE = anyarray
);
!>)

geeg:
Hi Kun,

Thanks for adding a design document, it will be very helpful for code review.  I'm not sure if you had planned to work on it some more, but these are a few initial comments I have regarding the design doc in the current commit:
- Variables/symbols need to be defined when they are used (e.g. "M is the total number of features" and "F(y_k, x_k) = ..." in the Formulation section, and "Mi is defined as a |Y| x |Y| matrix, where |Y| is the size of the tag space, and each element of Mi represents...." in the Forward-backward Algorithm section, etc.)
- For ease of reading, make sure that variables/symbols are used consistently throughout the doc.  For example, it seems like "n" was used as the label position in the first half of the doc, but in the latter half of the doc, "i" is used.
- I don't understand why the gradient of the log-likelihood (in the Formulation section) still has lambda in the first term.  Shouldn't that lambda be dropped through the derivative?
- Your notation for the feature function f_m(y_n, y_n-1, x_n) seems to imply that the features only depend on the current word (x_n); however, can't the features depend on any part of the whole sentence (x)? Though I suppose the six features implemented only make use of the current word.  (If there are plans to incorporate more global features or allowing users to add their own features, it would be better to switch x_n to x.)
- The Algorithm:Finalization in the Parallel CRF Training section doesn't add to my understanding.  I'm not sure if you meant to include it (it looks like an exact copy of Algorithm: CRF training from the previous page). However, the explanatory paragraph afterwards ("Feeding the current solution, gradient, ...") is good, but probably needs bullet points (e.g. 1. Penalize the log likelihood... 2. Negate the gradient... etc)
- I think there needs to be more explanation for why certain algorithms and formats were used, specifically in the Training Data Feature Extraction section.
- Isn't the lambda_i in the Forward-backward Algorithm section supposed to be beta_i?
- For ease of reading, tables should be placed after their explanations (e.g. Table 6.1 and Table 6.2 should be placed after section 6.4.3 Learned Model).  They should also be referenced somewhere in the explanations (e.g. "An example of a feature table and dictionary table extracted from training data is provided in Table 6.1 and Table 6.2") -- otherwise, the tables are just floating around.
- There are a few grammatical errors that need to be edited, but I can help out later (low priority)

livingstream:
Hi Grace,
    Thanks for the valuable comments.  I may randomly copy and paste with changing the actual contents. I am keeping revising it. 
Also please check out the commit I just pushed.  I will address your comments gradually.

All the best,
Kun

livingstream:
Hi Grace, I partially address your comment 1 in my latest commit.  Looks ok to you?

Commit name "partially address Grace's comment 1"

geeg:
Hi Kun,

Yes, that's exactly the idea I was going for.

haradh1:
Hi Kun,
So is this done?

livingstream:
Hi Hitoshi,
     Sorry for late reply, I was hoping I can finish it at the time of replying your email.  But ...

Al though I have finished 11 pages of crf modules design document and it looks like the doc contains every components it supposed to have .  
 But I am sure it needs to be polished a lot especially the feature extraction part. 
At the current stage, I think we are better off if we can polish the document and the code at the same time to move on. 

All the best,
Kun

haradh1:
Merged with my modification around online document index and m4 macros for ORDERED aggregate syntax.

PR #200, opened 2012-08-17
Fix wrong results issue with Sketch Estimators

renyi533:
Jira: http://jira.madlib.net/browse/MADLIB-567

Issue was with the merge function when the size of the input was large enough that we switch from exact calculations to sketch based estimators.  The merge function was failing to correctly account for the internal structure of the sketch data and OR-ing the wrong byteas.

### DIFF
--- a/methods/sketch/src/pg_gp/fm.c
+++ b/methods/sketch/src/pg_gp/fm.c
@@ -378,7 +378,10 @@ Datum __fmsketch_merge(PG_FUNCTION_ARGS)
         fmtransval *newval;
         tblob_big = fm_new(transval1);
         newval = (fmtransval *)VARDATA(tblob_big);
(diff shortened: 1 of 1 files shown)


cwelton:
The change itself seems good.  However reviewing the code I do have a couple concerns that we should followup on.
1.  This code is spoofable.  Which is to say the user can give us bytea's and we will assume that we can safely follow a user supplied pointer.  This is not introduced by this PR, so I don't insist on fixing it before we merge, but it is something that we will need to address.
2.  I'm not convinced by the `char storage[];` in the fmtransval.  For one it is a used as a bytea not a char[].  Second more typically we use `char storage[1]` for this sort of use which should be more portable - though per point 1 including pointers in user supplied bytea structures must be avoided anyways.

cwelton:
This looks safe to pull to me.

PR #201, opened 2012-08-17
Decision Tree: Fix RLE column compression version check.

haradh1:
4.2.0. was a bit too early to use RLE_TYPE for the column compression.


cwelton:
Accept.

PR #202, opened 2012-08-18
Added user doxygen for LMF

haying:
and minor fixes of design doc


cwelton:
Doc only change, safe for pull.

PR #203, opened 2012-08-20
Change the return type of rf_train from RECORD to rf_train_result.

aojwang:
JIRA: MADLIB-631
I tracked the history of the file in our local repository. It seems we made a mistake for replacement. The RECORD type will affect the cases which check the field in the result type.


cwelton:
Okay.

PR #204, opened 2012-08-20
Add unit test for checking the variables in the returned type of train API

aojwang:
JIRA: MADLIB-633
Current, there are no unit tests for checking the returned type. It would be better that if we have the checking so that we can find the changes during development.


cwelton:
It looks like you're just trying to confirm that the definition of the returned tuple hasn't changed?  Wouldn't that be more reliable to actually investigate the returned datatype's definition rather than trying to lookup items in the tuple within plpgsql?

cwelton:
Example query to get the information from the catalog:

```
select attname from pg_attribute where attrelid='madlib.name_of_my_type'::regclass order by attnum;
 attname 
---------
 a
 b
(2 rows)
```

PR #205, opened 2012-08-20
The wrong encoding result cases the scoring API failure.

aojwang:
JIRA: MADLIB-632
During encoding, we should not btrim the attribute value, since we will use it as join key between encoded table and raw training table. The second problem is we always use 'explict' option during encoding in classification/score. We should use the passed parameter.

### DIFF
--- a/methods/cart/src/pg_gp/dt_preproc.sql_in
+++ b/methods/cart/src/pg_gp/dt_preproc.sql_in
@@ -1079,7 +1079,11 @@ m4_changequote(>>>`<<<, >>>'<<<)
     IF (h2hmv_routine_id = 1) THEN
         where_txt = ' WHERE NULLIF(NULLIF(btrim(fval, '' ''), ''?''), '''') IS NOT NULL';
     ELSE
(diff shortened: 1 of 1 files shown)


methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -1079,7 +1079,11 @@ m4_changequote(>>>`<<<, >>>'<<<)
     IF (h2hmv_routine_id = 1) THEN
         where_txt = ' WHERE NULLIF(NULLIF(btrim(fval, '' ''), ''?''), '''') IS NOT NULL';
     ELSE
-        fval_txt  = ' NULLIF(NULLIF(btrim(fval, '' ''), ''?''), '''')';
+        fval_txt  = ' CASE WHEN NULLIF(NULLIF(btrim(fval, '' ''), ''?''), '''') IS NULL THEN
```
haradh1, line 1082: I prefer to use COALESCE in this case, but not significant difference.

PR #206, opened 2012-08-20
Association Rules (Apriori) Rewrite

aojwang:
JIRA: MADLIB-638, MADLIB-435, MADLIB-462, MADLIB-475, MADLIB-449, MADLIB-461, MADLIB-442, MADLIB-443

Details please refer to the corresponding JIRA.


methods/svec/src/pg_gp/operators.c
```diff
@@ -555,6 +555,34 @@ Datum svec_mult(PG_FUNCTION_ARGS)
...
+	}
+
+	if (PG_ARGISNULL(0) && !PG_ARGISNULL(1))
+	{
+		PG_RETURN_SVECTYPE_P(PG_GETARG_SVECTYPE_P_COPY(1));
+	}
+
+	if (!PG_ARGISNULL(0) && PG_ARGISNULL(1))
+	{
+		PG_RETURN_SVECTYPE_P(PG_GETARG_SVECTYPE_P_COPY(0));
```
cwelton: Copy not needed when returning the first parameter (state) of an aggregate function.

methods/svec/src/pg_gp/operators.c
```diff
@@ -555,6 +555,34 @@ Datum svec_mult(PG_FUNCTION_ARGS)
...
+	}
+
```
cwelton: And when both parameters are NULL?

cwelton:
Is there any functional change from this work, or is it only performance?

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+        plpy.execute("TRUNCATE TABLE assoc_rule_sets_loop");
+        plpy.execute("""
+            INSERT INTO assoc_rule_sets_loop(set_list, support)
+            SELECT 
+              set_list, float8div(cnt, {0}) as support
+            FROM 
+            (
+              SELECT
+              {1}.svec_cast_positions_float8arr
+               (array_agg(t1.item), array_agg(1), {2}, 0)
```
haradh1: What's array_agg(1)?  What about ARRAY[1]?

cwelton:
Comment from Jianwang : There are some cases where the results will be different.

We need to have this explained in more detail.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+                 SELECT id, 
+                    unnest({5}.svec_nonbase_positions(set_list, 0)) 
+                    as item
+                 FROM  assoc_loop_aux
+               ) t1,
+              assoc_item_svec u
+              WHERE t1.item = u.item_id
+              GROUP BY id
+            )  t
+            WHERE float8ge(cnt::FLOAT8, {6})
```
haradh1: Why isn't this just '>='?  Operator has more information in the optimizer than a raw function.

cwelton:
What is the complete set of jiras that you think this is addressing?

We need to know what is intended as changes in behavior, and what is simply performance work.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang: I see. But for float number, we can't count on = for equal operation.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang: We need an array with size N rather than 1.

methods/svec/src/pg_gp/operators.c
aojwang: Thanks. Done.

methods/svec/src/pg_gp/operators.c
aojwang: Should be return NULL. Done.

src/ports/postgres/dbconnector/UDF_impl.hpp
```diff
@@ -45,22 +88,25 @@
 UDF::call(FunctionCallInfo fcinfo) {
```
aojwang, line 131: Currently, C++ AL doesn't support a function returns SETOF. We add this function here to support this feature.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+            support is not None and 
+            confidence is not None and
+            tid_col is not None and 
+            item_col is not None and
+            input_table is not None and 
+            verbose is not None,
+            "none of the parameters except the output_schema should be null"
+            );
+
+    __assert(
```
aojwang, line 75: Related to JIRA: [MADLIB-442](http://jira.madlib.net/browse/MADLIB-442) and [MADLIB-443](http://jira.madlib.net/browse/MADLIB-443) for parameter checking.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+         m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (tid)')
+         """); 
+    
+    if verbose :
+        plpy.info("finished encoding input table: {0}".format(
+                time.time() - begin_step_exec));
+   
+    begin_step_exec = time.time();
+    
+    plpy.execute("DROP TABLE IF EXISTS assoc_item_svec");
```
aojwang: This table stores the item id and transaction IDs for this id. We use svec to represent the transaction IDs.  Therefore, we can easily get the common transactions of two or more items by multiply theirs svec. The old version doesn't have this table.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+            time.time() - begin_step_exec));
+        plpy.info("Beginning iteration #1"); 
+
+
+    begin_step_exec = time.time();
+    
+    # this table keeps the patterns in each iteration.
+    plpy.execute("DROP TABLE IF EXISTS assoc_rule_sets");
+    plpy.execute("""
+        CREATE TEMP TABLE assoc_rule_sets
```
aojwang, line 229: we use a text to unique identify an itemset (svec) rather than its hash key.
For example,  we have three items: 'a', 'b' and 'c', we assign 1, 2 and 3 to them respectively. for an itemset '{3}:{1}', the text (hash_text) is '1,2,3'. That's we first get the positions of the elements whose value is 1, and then concat them to a string with comma. This string is the hash text, and it's unique.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+                ) as set_list
+             FROM assoc_rule_sets_loop t1, 
+                  rule_set_rel t2, 
+                  assoc_rule_sets_loop t3
+             WHERE t1.id = t2.sid and t2.did = t3.id
+           ) t
+           WHERE {4}.svec_l1norm(set_list)::INT = {5}
+           """.format(madlib_schema, madlib_schema, madlib_schema, 
+                      madlib_schema, madlib_schema, iter + 1)
+           );
```
aojwang, line 368: The above code has the same functionality with the following code. However, according to my previous testing, the performance and scalability of cross join is bad. Therefore, we use a table to store the join relation, and then join the three table together. This fix is part of MADLIB-475 and MADLIB-462.

``` sql
           INSERT INTO assoc_loop_aux(set_list)
           SELECT DISTINCT ON({0}.svec_to_string(set_list)) set_list
           FROM (
             SELECT
                {1}.svec_minus(
                    {2}.svec_plus(t1.set_list, t3.set_list), 
                    {3}.svec_mult(t1.set_list, t3.set_list)
                ) as set_list
             FROM assoc_rule_sets_loop t1 cross join
                  assoc_rule_sets_loop t3
             WHERE t3.id > t1.id
           ) t
           WHERE {4}.svec_l1norm(set_list)::INT = {5}
```

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+            INSERT INTO assoc_rule_sets_loop(set_list, support)
+            SELECT 
+              set_list, float8div(cnt, {0}) as support
+            FROM 
+            (
+              SELECT
+              {1}.svec_cast_positions_float8arr
+               (array_agg(t1.item), array_agg(1), {2}, 0) 
+               as set_list,
+              {3}.svec_l1norm({4}.svec_agg_mult(u.tids))
```
aojwang: This line code is used to get the support count for an itemset. We use the new format, assoc_item_svec, to calculate this count. The performance is far better than the old one, which use cross join between assoc_loop_aux and the encoded input table. This is another part fix for MADLIB-475 and MADLIB-462.
According to my test, this operation still occupies more than 1/2 of the total running time. The reasonis, that the performance of svec_agg_mult is not good, especially for a large dataset (i means there are a large number of transactions).  I think, in the next release, we can implement a compressed bitmap for this case. Then the performance should be good.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
cwelton: The suggestion was for >= not =.  Plus calling the function directly instead of by operator doesn't actually change equality comparisons...

src/ports/postgres/dbconnector/UDF_impl.hpp
```diff
@@ -36,6 +36,49 @@
...
+    funcctx = SRF_PERCALL_SETUP();
+
+    result = Function(fcinfo).SRF_next(
+            	static_cast<SystemInformation*>(funcctx->user_fctx)->user_fctx,
+            	&is_last_call);
+
+    if (is_last_call)
+    	SRF_RETURN_DONE(funcctx);
+
+    SRF_RETURN_NEXT(funcctx, result.getAsDatum(fcinfo));
```
haying, line 122: A checking of result.isNull() is probably needed.

src/ports/postgres/dbconnector/UDF_impl.hpp
```diff
@@ -45,22 +88,25 @@
...
-            ->functionInformation(fcinfo->flinfo->fn_oid)->cxx_func
-            = invoke<Function>;
+        if (fcinfo->flinfo->fn_retset) {
+            return SRF_invoke<Function>(fcinfo);
+        }
+        else {
+        	AnyType args(fcinfo);
+            SystemInformation::get(fcinfo)
+                ->functionInformation(fcinfo->flinfo->fn_oid)->cxx_func
+                = invoke<Function>;
```
haying, line 145: This setting of `functionInformation` is missing if the call is set returning function.

haying:
Overview of reviewing adding Set Returning Function (SRF) to C++ AL:
1. The overall functionality added looks sane, except that I made 2 comments which indicate some mismatches between calling a SRF and non-SRF. They should be addressed.
2. I somehow feel the code organization can be improved, but given my limited experiences on SRF and C++ AL, I don't think I am the best person to give suggestions for this. The total amount of code is not too much, so I am not sure whether it is significant.
3. I am not sure whether GPDB has compatibility issues of executing SRF or not. This is actually my main concern.

src/ports/postgres/dbconnector/UDF_impl.hpp
aojwang, line 122: But the question is, that the result may be a NULL element in some cases. As it's a function in top level, we should allow users to return NULL element.

src/ports/postgres/dbconnector/UDF_impl.hpp
aojwang, line 145: Actually, we consider this situation. The problem is that the our function `SRF_invoke` needs to return Datum due to the two MACROs `SRF_RETURN_NEXT` and `SRF_RETURN_DONE`. Therefore, the function signature is different with the `cxx_func`. We search the code, and find that it only be cached. We didn't have a case to use it. So, we didn't set the function pointer. If later, we have the scenarios use this pointer, then we can simply add a new function pointer for the SRF function. Thanks.

aojwang:
What do you mean here about the compatibility issues? Could u please give me an example. Thanks.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang: After look at the code of float8ge. It seems the function doesn't set any margin for the equal comparison. I will change to >-=. Thanks.

src/ports/postgres/dbconnector/UDF_impl.hpp
haying, line 122: If resilt.isNull() == true, I am not sure result.getAsDatum() will be valid or not?

haying:
I am not sure whether there are compatible issues or not, so I cannot quite give examples. What are the GPDB versions that you have this code tested?

aojwang:
It runs well on our supported platforms: GPDB411, GPDB422, Postgres8.4, Postgres9.0 and Postgres9.1.

src/ports/postgres/dbconnector/UDF_impl.hpp
aojwang, line 122: I think we may need to enhance the AnyType when isNull() is true. That's we should map the NULL AnyType instance (I mean  the Null() function in the following code) to a NULL Datum ( PG/GP did like this `#define PG_RETURN_NULL()  do { fcinfo->isnull = true; return (Datum) 0; } while (0)`).  With this enhancement, we didn't to care the NULL AnyType and it would be convenient for the invokers. 

``` c
inline
AnyType
Null() {
    return AnyType();
}


inline
AnyType::AnyType()
  : mContent(Null),
    mDatum(0),
    fcinfo(NULL),
    mSysInfo(NULL),
    mTupleHeader(NULL),
    mTypeID(InvalidOid),
    mTypeName(NULL),
    mIsMutable(false)
    { }


inline
Datum
AnyType::getAsDatum(FunctionCallInfo inFnCallInfo,
    Oid inTargetTypeID) const {

    consistencyCheck();

    // The default value to return in case of Null is 0. Note, however, that
    // 0 can also be a perfectly valid (non-null) Datum. It is the caller's
    // responsibility to call isNull() separately.
    if (isNull())
        return 0;
    ...
}


inline
void
AnyType::consistencyCheck() const {
    const char *kMsg("Inconsistency detected while converting between "
        "PostgreSQL and C++ types.");

    madlib_assert(mContent != Null || (mDatum == 0 && fcinfo == NULL &&
        mSysInfo == NULL && mTupleHeader == NULL && mTypeID == InvalidOid &&
        mTypeName == NULL && mChildren.empty()),
        std::logic_error(kMsg));
    ...
}
```

aojwang, line 122: What do you think? Maybe we can ask Florian for some suggestions, since he did the design.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+        CREATE TEMP TABLE assoc_item_uniq (item_id, item_text, cnt) AS
+        SELECT row_number() OVER (ORDER BY cnt DESC)::BIGINT,
+           item::TEXT,
+           cnt::FLOAT8
+        FROM (
+           SELECT item AS item, count(tid) as cnt
+           FROM assoc_input_unique
+           GROUP BY item
+        ) t
+        WHERE cnt::FLOAT8 >= {0}
```
aojwang, line 197: We use >= here for fixing MADLIB-449: AR: We choose rules which is larger than minimum support instead of larger and equal than minimun support.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+                 SELECT id, 
+                    unnest({5}.svec_nonbase_positions(set_list, 0)) 
+                    as item
+                 FROM  assoc_loop_aux
+               ) t1,
+              assoc_item_svec u
+              WHERE t1.item = u.item_id
+              GROUP BY id
+            )  t
+            WHERE cnt::FLOAT8 >= {6}
```
aojwang: We use >= here for fixing MADLIB-449: AR: We choose rules which is larger than minimum support instead of larger and equal than minimun support.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+            FROM (
+                SELECT  
+                    {0}.gen_rules_from_cfp(hash_text, iteration) as item, 
+                    support as support_xy
+                FROM assoc_rule_sets
+                WHERE iteration > 1
+            ) t, assoc_rule_sets x, assoc_rule_sets y
+            WHERE t.item[1] = x.hash_text AND
+                  t.item[2] = y.hash_text AND
+                  (t.support_xy / x.support) >= {1}
```
aojwang, line 431: We use >= here for fixing MADLIB-449: AR: We choose rules which is larger than minimum support instead of larger and equal than minimun support.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+                    {0}.svec_nonbase_positions(set_list, 0), ','), 
+                    set_list, 
+                    support, {1}
+             FROM assoc_rule_sets_loop""".format(madlib_schema, iter)
+             );
+                     
+        plpy.execute("TRUNCATE TABLE assoc_loop_aux");
+        plpy.execute(""" 
+           INSERT INTO assoc_loop_aux(set_list)
+           SELECT DISTINCT ON({0}.svec_to_string(set_list)) set_list
```
aojwang: The fix for MADLIB-461: AR: Result is incorrect when running on dataset mushroom. 

The old version use the hash key of a svec to unique identify an association rule. However, as we know, the hash key is not unique, which cause the result is unstable. Especially, there are a large number of rules. Here, we define the table as below. As we know, the serial type will generate the unique number for each inserted row. Furthermore, we use the code `DISTINCT ON({0}.svec_to_string(set_list))` to select the unique rules (set_list, which is a svec). We can't distinct on the svec directly, since it's not hashable. And the group by on svec also has problem (MPP-18127).

``` python
    plpy.execute(""" 
         CREATE TEMP TABLE assoc_loop_aux 
            (
            id              BIGSERIAL,
            set_list        {0}.svec
            )
         m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (id)')
         """.format(madlib_schema));
```

aojwang:
<b> I add comments to the main changes of the PR. In summary:

MADLIB-435: Function call assoc_make_svec_sfunc in MADlib.assoc_rules does not contain schema name
`Fixed. We will not use this function any more.`

MADLIB-462: After AR run almost several hours, the server is slow and reinstall madlib will be halt
`Fixed: We will not change svec any more, and all the aggregates and time critical functions were written by C`

MADLIB-475: AR processing time does not decrease when segment number increase
`Three reasons cause this issue: 1) change svec again and again. 2) use a lot of cross join. 3) traverse the table row by row in PLPGSQL. Now, fixed.`

MADLIB-449: We choose rules which is larger than minimum support instead of larger and equal than minimun support
`By the definition of support and confidence, we should use >= rather than >. Now, fixed. Our behavior is the same as Weka.`

MADLIB-461: Result is incorrect when running on dataset mushroom
`It is cause by the hash function of svec. Now, fixed. We didn't use the svec any more.`

MADLIB-442: Invalid value for i_support and i_confidence can not throw error exception
`The range should be (0, 1]`

MADLIB-443:When support = 0, AR does not return result on limited time slot on some data set.
`The range should be (0, 1].`

methods/svec/src/pg_gp/svec.sql_in
```diff
@@ -977,6 +977,22 @@ CREATE AGGREGATE MADLIB_SCHEMA.svec_median_inmemory (float8) (
...
+CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.svec_agg_mult
+    (
+    MADLIB_SCHEMA.svec,
+    MADLIB_SCHEMA.svec
+    ) 
+RETURNS MADLIB_SCHEMA.svec AS 'MODULE_PATHNAME', 'svec_agg_mult'
+LANGUAGE C STRICT IMMUTABLE; 
+
+-- The aggregate to multiple all the svec in a group
+CREATE AGGREGATE MADLIB_SCHEMA.svec_agg_mult(MADLIB_SCHEMA.svec) (
```
haradh1: We have svec_sum().  In Excel, the multiplication version of SUM() is called PRODUCT(). It seems that's the de-fact name when googling around.

haradh1:
Can you rebase on the master?  I got a merge-conflict.

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,188 @@
...
+        	char 				*p_post;
+        	char				*p_begin;
+        	char				*p_cur;
+        	bool 				is_cont;
+        	Datum				*result;
+        	char				**p_sel_pos;
+        	int					len = 0;
+
+        	if (!is_last_call)
+        		elog(ERROR, "the paramter is_last_class should not be null");
```
haradh1: I think elog() cannot be used in C++ without PG_TRY()

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,188 @@
```
haradh1, line 0: This file has a lot of inconsistent indentation.  As always, please make it consistent.  Especially in C++ code, we use 4 spaces for one level.  Make sure your editor visualizes spaces and tabs, trailing white spaces.

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,188 @@
+/*
+ *
+ * @file assoc_rules.c
```
haradh1: cpp

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,188 @@
...
+			myfctx = (perm_fctx *) palloc(sizeof(perm_fctx));
+			myfctx->positions 	= positions;
+			myfctx->pos_len  	= strlen(positions);
+			myfctx->num_elems 	= args[1].getAs<int32>();
+			myfctx->num_calls	= (1 << myfctx->num_elems) - 2;
+			myfctx->flags 		= (bool*) palloc0(sizeof(bool) *
+											myfctx->num_elems);
+
+			/* return type id is TEXTOID, get the related information */
+			get_typlenbyvalalign
```
haradh1: Use MADLIB_WRAP_PG_FUNC(), as this function may cause ERROR.

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,188 @@
...
+					myfctx->flags[i] = true;
+					break;
+				}
+				else
+				{
+					myfctx->flags[i] = false;
+				}
+			}
+
+			pre_text 	= (char*) palloc0(sizeof(char) * myfctx->pos_len);
```
haradh1: palloc0 is also in danger.

src/ports/postgres/dbconnector/UDF_impl.hpp
haradh1, line 122: Actually, I have never seen such a SRF that returns NULL value (not a tuple cotaining NULL value).  I'm not even sure if it's expected.  With SRF_RETURN_NEXT(), you always return the Datum.  if you set fcinfo->isnull to true before SRF_RETURN_NEXT(), it might be possible to do it, but again it's not common.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+                   "item_col must be in the input table");
+ 
+    if output_schema is None :
+        rv = plpy.execute("SELECT current_schema() as s");
+        output_schema = rv[0]["s"];
+    else :
+        rv = plpy.execute("""
+            SELECT count(*) as cnt 
+            FROM pg_namespace
+            where nspname = lower(btrim('{0}', ' '))
```
haradh1: This is not correct if nspname is capitalized.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+    # note: duplicated records will be removed.
+    plpy.execute("DROP TABLE IF EXISTS assoc_input_unique"); 
+    rv = plpy.execute("""
+        SELECT max({0}) as x, min({1}) as i, count({2}) as c
+        FROM {3}""".format(tid_col, tid_col, tid_col, input_table));
+    if rv[0]["i"] != 1 or rv[0]["x"] - rv[0]["i"] != rv[0]["c"] :
+        plpy.execute("""
+             CREATE TEMP TABLE assoc_input_unique AS
+             SELECT 
+                dense_rank() OVER (ORDER BY tid)::BIGINT as tid,
```
haradh1, line 159: Do we really need continuous value?  If you want some id, I guess serial can do.  Wouldn't it acceptable to have invalid id?

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
haradh1: Isn't DISTINCT ON slow?  Tried GROUP BY instead?

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
```diff
@@ -0,0 +1,542 @@
...
+        plpy.execute(""" 
+             INSERT INTO assoc_rules_aux_tmp 
+                (pre, post, support, confidence, lift, conviction)
+             SELECT
+                t.item[1], 
+                t.item[2],
+                t.support_xy,
+                float8div(t.support_xy, x.support),
+                float8div(t.support_xy, x.support * y.support),
+                CASE WHEN float8eq(float8div(t.support_xy, x.support), 1) THEN
```
haradh1: Also wonder why you don't use operators but functions.

src/modules/assoc_rules/assoc_rules.cpp
aojwang: Could u please tell me the reason. Thanks.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang, line 159: We need the continuous value. If it's not continuous, then the svec maybe very large, since we use the IDs as the positions of the svec. Say, we only have two transactions, and the IDs are 1, 100 respectively. If we don't make them continuous, then the dimension of the svec is 100. If we do, then the dimension is only 2.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang: It has a bug if we group by svec (see the jira: MPP-18127. Caleb ask me open to you when i show this bug two weeks ago :)).
However, according to the test, the time for this operation is very less, compared to the total time.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
haradh1, line 159: Is that the main motivation of svec?  The repeating zero between 1 and 100 will be encoded efficiently in svec.  So [3, 4] and [3, 0 ..., 4] doesn't make big difference, though I know it adds zero in between.  I concern more about the anti-scalability of the single-node window function.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
haradh1: But you used svec_to_string() anyway.  Text type shouldn't have bug, no?

src/ports/postgres/dbconnector/UDF_impl.hpp
aojwang, line 122: The direct case is, that unnest an array with null elements. What i mean here is, that it would be better and consistent if the NULL AnyType (anytype.isNull is true) is corresponding to the NULL Datum.

``` sql
testdb=# select unnest(array[1, 2, null, 3, null, 4]);
 unnest 
--------
      1
      2

      3

      4
(6 rows)

```

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang: Removed the lower and btrim.

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang, line 159: One more case is, that continuous IDs will make the svec more compact. For example. we have  6 transaction IDs, 1, 3, 5, 7, 9 and 15. All of which contains item 'a'. Then the svec for the 'a' is {1, 1, 1, 1, 1, 1, 1, 1, 1, 5, 1}:{1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1}'. If the IDs are continuous (the IDs will be 1, 2, 3, 4, 5 and 6), then the svec is {6}:{1}. Especially, if the original dataset has a lot gaps, then it needs a lot of space. 

I understand the scalability of the window function. As you see, i will first check if they are continuous, then we will not invoke the window function.  As this encoding is one time operation, i would expect it will not spend too much time. The following log is from running ar_pos dataset on 40 segments, which has 515597 transaction IDs. The time for this step is 5.83, while the total time is 56.25.

```
INFO:  finished checking parameters
CONTEXT:  PL/Python function "assoc_rules"
INFO:  finished removing duplicates
CONTEXT:  PL/Python function "assoc_rules"
INFO:  finished encoding items
CONTEXT:  PL/Python function "assoc_rules"
INFO:  finished encoding input table: 5.83728909492
CONTEXT:  PL/Python function "assoc_rules"
INFO:  finished vertical table transforming: 0.648931026459
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Beginning iteration #1
CONTEXT:  PL/Python function "assoc_rules"
INFO:  141 Frequent itemsets found in this iteration
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Completed iteration # 1. Time: 0.130474805832
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Beginning iteration # 2
CONTEXT:  PL/Python function "assoc_rules"
INFO:  time of generating rules: 0.39672088623
CONTEXT:  PL/Python function "assoc_rules"
INFO:  389 Frequent itemsets found in this iteration
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Completed iteration # 2. Time: 3.85302996635
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Beginning iteration # 3 
CONTEXT:  PL/Python function "assoc_rules"
INFO:  time of generating rules: 0.451709985733
CONTEXT:  PL/Python function "assoc_rules"
INFO:  410 Frequent itemsets found in this iteration
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Completed iteration # 3. Time: 24.4801518917
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Beginning iteration # 4 
CONTEXT:  PL/Python function "assoc_rules"
INFO:  time of generating rules: 0.130084037781
CONTEXT:  PL/Python function "assoc_rules"
INFO:  151 Frequent itemsets found in this iteration
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Completed iteration # 4. Time: 16.4897310734
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Beginning iteration # 5
CONTEXT:  PL/Python function "assoc_rules"
INFO:  time of generating rules: 0.119781017303
CONTEXT:  PL/Python function "assoc_rules"
INFO:  8 Frequent itemsets found in this iteration
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Completed iteration # 5. Time: 4.15427207947
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Beginning iteration # 6
CONTEXT:  PL/Python function "assoc_rules"
INFO:  time of generating rules: 0.421010971069
CONTEXT:  PL/Python function "assoc_rules"
INFO:  0 Frequent itemsets found in this iteration
CONTEXT:  PL/Python function "assoc_rules"
INFO:  Completed iteration # 6. Time: 0.524311065674
CONTEXT:  PL/Python function "assoc_rules"
INFO:  begin to generate the final rules
CONTEXT:  PL/Python function "assoc_rules"
INFO:  5592 Total association rules found. Time: 0.133746862411
CONTEXT:  PL/Python function "assoc_rules"
 output_schema | output_table | total_rules |   total_time
---------------+--------------+-------------+-----------------
 test          | assoc_rules  |        5592 | 00:00:56.254781

```

src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
aojwang: From my current test, the result is stable. I don't know if the text has bug or not. That's what i can figure out for now. Any suggestions are greatly appreciated.

src/modules/assoc_rules/assoc_rules.cpp
aojwang, line 0: Done.

src/modules/assoc_rules/assoc_rules.cpp
aojwang: Done.

haradh1:
My understanding is this pull request is discarded based on the change of our approach to address the issues separately.  Please close this if I'm correct.

aojwang:
It's so strange that The following query can sometimes run successfully  on GPDB411 if table "t3" has a large number of rows, but sometimes it causes GPDB crash. GPDB422 doesn't have such problem. I will investigate more. Any comments and suggestions are welcome.

```
testdb=# explain analyze create table t4 as
testdb-#               SELECT id,
testdb-# 
testdb-#               madlib.svec_l1norm(madlib.svec_agg_mult(tids))
testdb-#               AS cnt
testdb-#              from t3
testdb-#               GROUP BY id;
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named 'id' as the Greenplum Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
NOTICE:  Error on receive from seg9 gpdb1.delta.sm.greenplum.com:40001 pid=30463: server closed the connection unexpectedly
DETAIL:  
    This probably means the server terminated abnormally
    before or while processing the request.
NOTICE:  Releasing gangs to finish aborting the transaction.
ERROR:  GPDB performed segment reconfiguration. (cdbgang.c:1648)
```

```
testdb=# explain analyze create table t4 as                                                                                                                
SELECT id,                                                                                                                                   
madlib.svec_l1norm(madlib.svec_agg_mult(tids))                                                                                              
 AS cnt                                                                                                                                      
from t3                                                                                                                                      
 GROUP BY id;
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named 'id' as the Greenplum Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
                                                                      QUERY PLAN                                                             

---------------------------------------------------------------------------------------------------------------------------------------------
----------
 HashAggregate  (cost=12397.65..12412.65 rows=63 width=40)
   Group By: t3.id
   Rows out:  Avg 1034.6 rows x 16 workers.  Max 1036 rows (seg0) with 6953 ms to first row, 7867 ms to end, start offset by 55 ms.
   Executor memory:  77576K bytes avg, 79838K bytes max (seg1).
   Work_mem used:  64737K bytes avg, 66566K bytes max (seg1).
   Work_mem wanted: 45546K bytes avg, 51142K bytes max (seg13) to lessen workfile I/O affecting 16 workers.
   (seg0)   1036 groups total in 32 batches; 1 overflows; 1671 spill groups.
   (seg0)   Hash chain length 1.0 avg, 2 max, using 2060 of 4325376 buckets.
   ->  Redistribute Motion 16:16  (slice1; segments: 16)  (cost=12362.65..12382.65 rows=63 width=40)
         Hash Key: t3.id
         Rows out:  Avg 4891.7 rows x 16 workers at destination.  Max 4912 rows (seg4) with 26 ms to first row, 4057 ms to end, start offset 
by 56 ms.
         ->  HashAggregate  (cost=12362.65..12362.65 rows=63 width=40)
               Group By: t3.id
               Rows out:  Avg 4891.7 rows x 16 workers.  Max 8256 rows (seg10) with 599 ms to first row, 1065 ms to end, start offset by 65 m
s.
               Executor memory:  49077K bytes avg, 81667K bytes max (seg10).
               ->  Seq Scan on t3  (cost=0.00..8807.43 rows=44441 width=40)
                     Rows out:  Avg 5173.1 rows x 16 workers.  Max 9150 rows (seg10) with 0.068 ms to first row, 12 ms to end, start offset b
y 66 ms.
 Slice statistics:
   (slice0)  * Executor memory: 78734K bytes avg x 16 workers, 81054K bytes max (seg1).  Work_mem: 66566K bytes max, 51142K bytes wanted.
   (slice1)    Executor memory: 49952K bytes avg x 16 workers, 82823K bytes max (seg10).
 Statement statistics:
   Memory used: 128000K bytes
   Memory wanted: 102684K bytes
 Total runtime: 8678.445 ms
(24 rows)
```

```
testdb=# explain analyze create table t4 as                                                                                                                
SELECT id,                                                                                                                                   
madlib.svec_l1norm(madlib.svec_agg_mult(tids))                                                                                               
AS cnt                                                                                                                                      
from t3                                                                                                                                      
 GROUP BY id;
NOTICE:  Table doesn't have 'DISTRIBUTED BY' clause -- Using column(s) named 'id' as the Greenplum Database data distribution key for this table.
HINT:  The 'DISTRIBUTED BY' clause determines the distribution of data. Make sure column(s) chosen are the optimal data distribution key to minimize skew.
NOTICE:  Error on receive from seg1 gpdb5.delta.sm.greenplum.com:40001 pid=15974: server closed the connection unexpectedly
DETAIL:  
    This probably means the server terminated abnormally
    before or while processing the request.
NOTICE:  Releasing gangs to finish aborting the transaction.
ERROR:  GPDB performed segment reconfiguration. (cdbgang.c:1648)
```

haradh1:
What is the query plan from GPDB422?  Does it use HashAgg, too?

Thanks,
Hitoshi

aojwang:
The plan on GPDB422. It seems the plan is the same, while it's faster than GPDB411

```
testdb=# explain analyze create table t4 as                                                                                                                SELECT id,                                                                                                                                   madlib.svec_l1norm(madlib.svec_agg_mult(tids))                                                                                              
 AS cnt                                                                                                                                      
from t3                                                                                                                                       
GROUP BY id;                                                                        
                                                                        QUERY PLAN                                                           

---------------------------------------------------------------------------------------------------------------------------------------------
--------------
 HashAggregate  (cost=3932.47..3947.47 rows=63 width=16)
   Group By: t3.id
   Rows out:  Avg 340.4 rows x 16 workers.  Max 341 rows (seg2) with 2153 ms to first row, 2169 ms to end, start offset by 1412 ms.
   Executor memory:  42489K bytes avg, 43713K bytes max (seg0).
   ->  Redistribute Motion 16:16  (slice1; segments: 16)  (cost=3897.47..3917.47 rows=63 width=40)
         Hash Key: t3.id
         Rows out:  Avg 1496.8 rows x 16 workers at destination.  Max 1510 rows (seg7) with 9.819 ms to first row, 792 ms to end, start offse
t by 1415 ms.
         ->  HashAggregate  (cost=3897.47..3897.47 rows=63 width=40)
               Group By: t3.id
               Rows out:  Avg 1496.8 rows x 16 workers.  Max 2580 rows (seg10) with 253 ms to first row, 267 ms to end, start offset by 1430 
ms.
               Executor memory:  17671K bytes avg, 37318K bytes max (seg10).
               ->  Seq Scan on t3  (cost=0.00..2776.65 rows=14011 width=40)
                     Rows out:  Avg 1552.0 rows x 16 workers.  Max 2772 rows (seg10) with 0.153 ms to first row, 4.411 ms to end, start offse
t by 1433 ms.
 Slice statistics:
   (slice0)    Executor memory: 43555K bytes avg x 16 workers, 44549K bytes max (seg0).
   (slice1)    Executor memory: 18537K bytes avg x 16 workers, 38498K bytes max (seg10).
 Statement statistics:
   Memory used: 128000K bytes
 Total runtime: 3606.786 ms
(19 rows)


```

aojwang:
I think the problem is the svec. However, i didn't get the root cause. Is it related to send/recv of svec?
NOTE: the type of tids is svec.

``` sql
testdb=# explain analyze create table t4 as                                                                                                                
SELECT id,                                                                                                                                   
madlib.svec_l1norm(madlib.svec_agg_mult(tids))                                                                                               
AS cnt                                                                                                                                      
from t3                                                                                                                                      
 GROUP BY id;
```

aojwang:
The running results on different platform are below. Is it a DB problem?
- Result on GPDB4113

```
testdb=# select * from madlib.assoc_rules(0.05, 0.25, 'tran_id', 'product', 'ar_adult', 'test', true);
...
ERROR:  Unexpected internal error: Segment process received signal SIGSEGV (postgres.c:3360)  (seg4 slice3 gpdb5.delta.sm.greenplum.com:40004 pid=9835) (cdbdisp.c:1457)
CONTEXT:  SQL statement "
            INSERT INTO assoc_rule_sets_loop(set_list, support)
            SELECT 
              set_list, cnt::FLOAT8/32561 as support
            FROM 
            (
              SELECT
              madlib.svec_cast_positions_float8arr
               (array_agg(t1.item), array_agg(1), 36, 0) 
               as set_list,
              madlib.svec_l1norm(madlib.svec_agg_mult(u.tids))
              AS cnt
             FROM 
               (
                 SELECT id, 
                    unnest(madlib.svec_nonbase_positions(set_list, 0)) 
                    as item
                 FROM  assoc_loop_aux
               ) t1,
              assoc_item_svec u
              WHERE t1.item = u.item_id
              GROUP BY id
            )  t
            WHERE cnt::FLOAT8 >= 1628.05
            "
```
- Result on GPDB422

```
testdb=# select * from madlib.assoc_rules(0.05, 0.25, 'tran_id', 'product', 'ar_adult', 'test', true);
...
 output_schema | output_table | total_rules |   total_time    
---------------+--------------+-------------+-----------------
 test          | assoc_rules  |      364473 | 00:00:39.317404
```
- Result on PG9.1

```
select * from madlib.assoc_rules(0.05, 0.25, 'tran_id', 'product', 'ar_adult', 'test', true);
...
 output_schema | output_table | total_rules |   total_time    
---------------+--------------+-------------+-----------------
 test          | assoc_rules  |      364473 | 00:06:29.294343
(1 row)
```

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,190 @@
...
+            char                *p_post;
+            char                *p_begin;
+            char                *p_cur;
+            bool                is_cont;
+            Datum               *result;
+            char                **p_sel_pos;
+            int                 len = 0;
+
+            if (!is_last_call)
+                ereport (
```
haradh1: Well, ereport doesn't solve the problem...

One way to easily to do that for short term is to make a postgres-ish function and wrap it by PG_TRY() ... PG_END_TRY() and return AnyType from there.  An example should look like

```
AnyType SRF_next() {
  Datum result;
  MADLIB_PG_TRY()  {
    result = call_your_c_function(fcinfo);
    return AnyType(result);
  }
  MADLIB_PG_DEFAULT_CATCH_AND_END_TRY;
}
```

and in your c function, you can call elog()/ereport()/palloc() and everything you are familiar with.

haradh1:
For the crash issue, find stack trace.

aojwang:
Sorry. It's not crash, but report error. The stack is below:

```
(gdb) backtrace
#0  0x99d9faa2 in __semwait_signal ()
#1  0x99ddebe8 in pthread_join$UNIX2003 ()
#2  0x007ed503 in MPPnoticeReceiver ()
#3  0x003143cb in ExecEvalVar ()
#4  0x002f2ce0 in spill_hash_table ()
#5  0x00359e73 in last_value_generic ()
#6  0x0035c56e in appendToFrameBuffer ()
#7  0x014053f6 in PLy_spi_execute_query ()
#8  0x01405dfb in PLy_spi_execute ()
#9  0x014ce49c in PyEval_EvalFrameEx ()
#10 0x014cf1f9 in PyEval_EvalFrameEx ()
#11 0x014cf1f9 in PyEval_EvalFrameEx ()
#12 0x014d034d in PyEval_EvalCodeEx ()
#13 0x014d0467 in PyEval_EvalCode ()
#14 0x0140b0df in PLy_function_handler ()
#15 0x0140b801 in plpython_call_handler ()
#16 0x0030645f in slot_get_partition ()
#17 0x00329936 in process_ordered_aggregate_multi ()
#18 0x0030bad1 in ExecEvalCase ()
#19 0x003297d1 in process_ordered_aggregate_multi ()
#20 0x002fd0a8 in EvalPlanQual ()
#21 0x002fa8fe in invertedAttrMap ()
#22 0x002fc650 in ExecCheckPlanOutput ()
#23 0x00555601 in PageIndexMultiDelete ()
#24 0x00546a89 in DeadLockCheckRecurse ()
#25 0x0054ad72 in DeadLockCheckRecurse ()
#26 0x0049fcc8 in walk_plan_node_fields ()
#27 0x004a9558 in ServiceListenLoop ()
#28 0x0039b32c in PQescapeBytea ()
#29 0x00001d16 in memtuple_has_misaligned_attribute ()
```

aojwang:
The following logs were attached to [MADLIB-638](http://jira.madlib.net/browse/MADLIB-638).
- Log from pg_log on master:

```
 52 2012-08-28 01:34:08.301936 PDT,"zhangh25","testdb",p16831,th1080621376,"[local]",,2012-08-28 01:33:58 PDT,45902,con7,cmd28,seg-1,,dx3,x45    902,sx1,"LOG","00000","Dispatcher encountered connection error on seg4 gpdb5.delta.sm.greenplum.com:40004 pid=16842: server closed the co    nnection unexpectedly
 53     This probably means the server terminated abnormally
 54     before or while processing the request.
 55 ",,,,,,"
 56             INSERT INTO assoc_rule_sets_loop(set_list, support)
 57             SELECT 
 58               set_list, cnt::FLOAT8/32561 as support
 59             FROM 
 60             (
 61               SELECT id,
 62               madlib.svec_l1norm(madlib.svec_agg_mult(u.tids))
 63               AS cnt
 64              FROM 
 65                (
 66                  SELECT id, 
 67                     unnest(madlib.svec_nonbase_positions(set_list, 0)) 
 68                     as item
 69                  FROM  assoc_loop_aux
 70                ) t1,
 71               assoc_item_svec u
 72               WHERE t1.item = u.item_id
 73               GROUP BY id
 74             )  t1, assoc_loop_aux t2
 75             WHERE t1.id = t2.id AND cnt::FLOAT8 >= 1628.05
 76             ",0,,,,
 77 2012-08-28 01:34:10.495027 PDT,"zhangh25","testdb",p16831,th1431225760,"[local]",,2012-08-28 01:33:58 PDT,45902,con7,cmd28,seg-1,,dx3,x45    902,sx1,"WARNING","01000","PL/Python: in PL/Python function ""assoc_rules""","<class 'plpy.SPIError'>: unrecognized error in PLy_spi_exec    ute_query",,,,,"select * from madlib.assoc_rules(0.05, 0.25, 'tran_id', 'product', 'ar_adult', 'test', true);",0,,"plpython.c",3170,
 78 2012-08-28 01:34:10.495055 PDT,"zhangh25","testdb",p16831,th1431225760,"[local]",,2012-08-28 01:33:58 PDT,45902,con7,cmd28,seg-1,,dx3,x45    902,sx1,"ERROR","58M01","Error on receive from seg4 slice3 gpdb5.delta.sm.greenplum.com:40004 pid=16960: server closed the connection une    xpectedly","
 79     This probably means the server terminated abnormally
 80     before or while processing the request.",,,,"SQL statement ""
 81             INSERT INTO assoc_rule_sets_loop(set_list, support)
 82             SELECT 
 83               set_list, cnt::FLOAT8/32561 as support
 84             FROM 
 85             (
 86               SELECT id,
 87               madlib.svec_l1norm(madlib.svec_agg_mult(u.tids))
 88               AS cnt
 89              FROM 
 90                (
 91                  SELECT id, 
 92                     unnest(madlib.svec_nonbase_positions(set_list, 0)) 
 93                     as item
 94                  FROM  assoc_loop_aux
 95                ) t1,
 96               assoc_item_svec u
 97               WHERE t1.item = u.item_id
 98               GROUP BY id
 99             )  t1, assoc_loop_aux t2
100             WHERE t1.id = t2.id AND cnt::FLOAT8 >= 1628.05
101             ""","select * from madlib.assoc_rules(0.05, 0.25, 'tran_id', 'product', 'ar_adult', 'test', true);",0,,"cdbdisp.c",1457,
102 2012-08-28 01:34:10.500661 PDT,"zhangh25","testdb",p16831,th1431225760,"[local]",,2012-08-28 01:33:58 PDT,45902,con7,cmd28,seg-1,,dx3,x45    902,sx1,"LOG","00000","Segment connection failed: allocateWriterGang attempted to return a bad gang.",,,,,,,0,,"cdbgang.c",1647,
103 2012-08-28 01:34:10.527357 PDT,"zhangh25","testdb",p16831,th1431225760,"[local]",,2012-08-28 01:33:58 PDT,45902,con7,cmd28,seg-1,,dx3,x45    902,sx1,"ERROR","XX000","GPDB performed segment reconfiguration. (cdbgang.c:1648)",,,,,,,0,,"cdbgang.c",1648,"Stack trace:
104 1    0xa3349a postgres <symbol not found> (elog.c:454)
105 2    0xa35208 postgres elog_finish (elog.c:1365)
106 3    0xb22f8a postgres allocateWriterGang (cdbgang.c:1651)
107 4    0xaefda3 postgres cdbdisp_dispatchDtxProtocolCommand (cdbdisp.c:1263)
108 5    0xb8f782 postgres <symbol not found> (cdbtm.c:1942)
109 6    0xb95721 postgres rollbackDtxTransaction (cdbtm.c:785)
110 7    0x4ee818 postgres AbortTransaction (xact.c:3420)
111 8    0x4f1435 postgres AbortCurrentTransaction (xact.c:3964)
112 9    0x8e42cc postgres PostgresMain (postgres.c:4406)
113 10   0x84f8d1 postgres <symbol not found> (postmaster.c:6436)
114 11   0x856bc5 postgres PostmasterMain (postmaster.c:2272)
115 12   0x7655ea postgres main (main.c:212)
116 13   0x32a341d994 libc.so.6 __libc_start_main (??:0)
117 14   0x475039 postgres <symbol not found> (??:0)
118 "
119 2012-08-28 01:34:10.528408 PDT,"zhangh25","testdb",p16831,th1431225760,"[local]",,2012-08-28 01:33:58 PDT,45902,con8,,seg-1,,dx3,x45902,s    x1,"LOG","00000","The previous session was reset because its gang was disconnected (session id = 7). The new session id = 8",,,,,,,0,,"cd    bgang.c",2772,

```
- pg_log on the segment reported error.

```
 17 *** glibc detected *** postgres: port 40005, zhangh25 testdb 10.5.212.121(47331) con7 seg5 cmd33 slice3 MPPEXEC INSERT: malloc(): memory 
    corruption: 0x0000000009785450 ***",,,,,,,,"SysLoggerMain","syslogger.c",517, 18 2012-08-28 01:52:29.423394 PDT,,,p17536,th-1904560736,,,,0,,,seg-1,,,,,"LOG","00000","3rd party error log:
 19 ======= Backtrace: =========

```

src/modules/assoc_rules/assoc_rules.cpp
aojwang: As this function was already in a try...catch blog, should we do that? 

```
UDF::call(FunctionCallInfo fcinfo) {
    int sqlerrcode;
    char msg[2048];
    try {
        if (fcinfo->flinfo->fn_retset) {
            return SRF_invoke<Function>(fcinfo);
        }
        ...
        }
    } catch (std::bad_alloc &) {
```

haradh1: You should make sure any possible ERROR being in PG_TRY() block without C++ try block.

aojwang: Done.

cwelton: When in C++ code the recommended way of raising errors is to throw a C++ exception rather than wrapping things in PG_TRY blocks and calling ereport.  See the logistic regression code for examples.

aojwang: Done. Thanks.

aojwang:
According to my test, the latest changes can run successfully on PG8.4/9.0/9.1, GPDB4.1 series and GPDB4.2. Some performance tests are available [here](http://confluence.greenplum.com/x/jaHtAg).

src/modules/assoc_rules/assoc_rules.cpp
haradh1: No, it's not done.  It is desirable to do it correctly in C++, but as far as you don't understand why palloc is dangerous here, you should write the SRF in your traditional way like in C, and do whatever you know including ereport().  Then wrap it in PG_TRY() and throw a C++ exception by MADLIB_PG_TRY() block so the error can correctly be handled.

At least the current code is worse than before, as new'ed pointers are not delet'ed, for example.

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    myfctx->flags[i] = true;
+                    break;
+                }
+                else
+                {
+                    myfctx->flags[i] = false;
+                }
+            }
+
+            pre_text  = new char[myfctx->pos_len];
```
haradh1, line 119: OK.  So NewDelete.cpp overloads new with palloc.  But I'm not quite sure if this new calls palloc...

aojwang, line 119: the call hierarchy: new -> Allocator::allocate -> Allocator::internalAllocate -> Allocator::internalPalloc (inside in PG_TRY) -> palloc

For convenience, i paste the code of  internalPalloc here. Thanks.

```
template <dbal::ZeroMemory ZM>
inline
void *
Allocator::internalPalloc(size_t inSize) const {
#if MAXIMUM_ALIGNOF >= 16
    return (ZM == dbal::DoZero) ? palloc0(inSize) : palloc(inSize);
#else
    if (inSize > std::numeric_limits<size_t>::max() - 16)
        return NULL;

    /* Precondition: inSize <= std::numeric_limits<size_t>::max() - 16 */
    const size_t size = inSize + 16;
    void *raw = (ZM == dbal::DoZero) ? palloc0(size) : palloc(size);
    return makeAligned(raw);
#endif
}
```

src/ports/postgres/dbconnector/UDF_impl.hpp
```diff
@@ -36,6 +36,50 @@
...
+template <class Function>
+inline
+Datum
+UDF::SRF_invoke(FunctionCallInfo fcinfo) {
+    FuncCallContext *funcctx = NULL;
+    MemoryContext oldcontext;
+    bool is_last_call = false;
+    AnyType result;
+
+    if (SRF_IS_FIRSTCALL()) {
```
haradh1: Hmm, SRF_IS_FIRSTCALL() and SRF_PERCALL_SETUP() could also call ERROR.  It is unsafe to use them outside of PG_TRY().

aojwang: Since i am not familiar with the SRF details, my question is:I find all the functions, such as pg_ls_dir and array_agg, with returning setof in pg_proc.sql didn't use PG_TRY for that. Those two functions you mentioned are very straightforward,  when they will raise error. Thank.

haradh1: The problem is issuing ERROR cannot catch the C++'s try block and if the ERROR passed the C++'s try block, the process hangs.  As far as we live within the backend C world, those functions are safe.

aojwang: Ok. Added the MADLIB_PG_TRY.

haradh1: And now the problem is SRF_next throws C++ exception, it should not intersect with PG_TRY() block.

aojwang: We just use PG_TRY for the two functions, what do you think?

```
UDF::SRF_invoke(FunctionCallInfo fcinfo) {
    FuncCallContext *funcctx = NULL;
    MemoryContext oldcontext;
    bool is_last_call = false;
    bool is_first_call = true;
    AnyType result;

    MADLIB_PG_TRY{
        is_first_call = SRF_IS_FIRSTCALL();
    } MADLIB_PG_DEFAULT_CATCH_AND_END_TRY;

    if (is_first_call) {
        /* create a function context for cross-call persistence */
        funcctx = SRF_FIRSTCALL_INIT();
        oldcontext = MemoryContextSwitchTo(funcctx->multi_call_memory_ctx);

        // we must construct the argument here, since it needs SRF_FIRSTCALL_INIT
        // to init the pointer: fn_extra
        AnyType args(fcinfo);
        args.mSysInfo->user_fctx = Function(fcinfo).SRF_init(args);
        MemoryContextSwitchTo(oldcontext);
    }

    MADLIB_PG_TRY{
        funcctx = SRF_PERCALL_SETUP();
    } MADLIB_PG_DEFAULT_CATCH_AND_END_TRY;

        // the invoker function will handle the exceptions from this function
    result = Function(fcinfo).SRF_next(
        static_cast<SystemInformation*>(funcctx->user_fctx)->user_fctx,
        &is_last_call);

    if (is_last_call)
        SRF_RETURN_DONE(funcctx);

    SRF_RETURN_NEXT(funcctx, result.getAsDatum(fcinfo));
}
```

haradh1: You want to replace the second SRF_IS_FISTCALL() with is_first_call.  But I'd like to see more elegant way.

aojwang: you mean wrap them in a single pg_try?

aojwang: For each MACRO, i define a function to wrap the macro inside the PG_TRY block, and use another macro to call this function. Therefore, it's easy to use in SRF_invoke.
As I don't know is it the way you want, i add the TODO comments. If you find a better way, please replace them. Thanks.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -50,608 +54,284 @@ S (X) = \frac{Total X}{Total transactions}
...
+  This will generate all association rules that meet a minimum support of <em>support</em> and confidence of <em>confidence</em>.
   
 - The results containing the rules, support, confidence, lift, and conviction are stored in the table assoc_rules in the schema specified by <em>output_schema</em>.
 <pre>
-	 Table "output_schema.assoc_rules"
-       Column   |       Type       | Modifiers 
+     Table "output_schema.assoc_rules"
+       Column | Type | Modifiers
     ------------+------------------+-----------
-     ruleid     | integer          |
```
haradh1, line 109: This change doesn't make sense.  I'll revert these format changes.

src/ports/postgres/dbconnector/UDF_impl.hpp
```diff
@@ -36,6 +75,53 @@
...
+    funcctx = MADLIB_SRF_PERCALL_SETUP();
+
+    // the invoker function will handle the exceptions from this function
+    result = Function(fcinfo).SRF_next(
+        static_cast<SystemInformation*>(funcctx->user_fctx)->user_fctx,
+        &is_last_call);
+
+    if (is_last_call)
+        SRF_RETURN_DONE(funcctx);
+
```
haradh1, line 121: @haying was right.  You need to check result.isNull() to return NULL.  I'll add it.

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include <dbconnector/dbconnector.hpp>
+
+#include "assoc_rules.hpp"
+
+namespace madlib
+{
+  namespace modules
+  {
+    namespace ar
```
haradh1, line 19: s/ar/assoc_rules/

haradh1:
Merged with my modification.  C++ support for SRF needs improvements.  I'll discuss with @fschopp

PR #208, opened 2012-08-21
insert the information of the columns in which all the values are null to the meta table

aojwang:
JIRA: [MADLIB-635](http://jira.madlib.net/browse/MADLIB-635)
During encoding process, if all the values of a column is null, then we will remove the column from breakup table in 'ignore' mode. Therefore, besides insert the columns' information in the breakup table to the metatable, we need to insert those columns in which all the values are null to the meta table. This PR is used to track the issue.

When i fix this bug, I found that we didn't change the file Module.yml. As now we invoke some functions from array_ops, we need to depend on that module. As it's a minor problem, i combine to this PR. Thanks.

### DIFF
--- a/methods/cart/src/pg_gp/dt_preproc.sql_in
+++ b/methods/cart/src/pg_gp/dt_preproc.sql_in
@@ -1633,6 +1633,29 @@ m4_changequote(>>>`<<<, >>>'<<<)
 
     EXECUTE curstmt;
 
+    IF (h2hmv_routine_id = 1) THEN
+        -- retrieve the information of the columns (all the values in those 
+        -- columns are missing), and insert them to the meta table. 
+        curstmt = MADLIB_SCHEMA.__format
+            (
+                'INSERT INTO %
+                 SELECT id, (ARRAY[''%''])[id] as column_name,
+                        ''f'' as column_type, ''t'', NULL, 0
+                 FROM (
+                     SELECT generate_series(1, %) id
+                     EXCEPT
+                     SELECT id FROM % WHERE column_type = ''f'' 
+                 ) t',
+                ARRAY[
+                    meta_tbl_name,
+                    array_to_string(attr_col_names, ''','''),
+                    array_upper(attr_col_names, 1)::TEXT,
+                    meta_tbl_name
+                ]
+            );
+        EXECUTE curstmt;
+    END IF;
+    
     -- no need this table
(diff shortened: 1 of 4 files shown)


methods/cart/src/pg_gp/dt_preproc.sql_in
```diff
@@ -1633,6 +1633,29 @@ m4_changequote(>>>`<<<, >>>'<<<)
...
+                 SELECT id, (ARRAY[''%''])[id] as column_name,
+                        ''f'' as column_type, ''t'', NULL, 0
+                 FROM (
+                     SELECT generate_series(1, %) id
+                     EXCEPT
+                     SELECT id FROM % WHERE column_type = ''f'' 
+                 ) t',
+                ARRAY[
+                    meta_tbl_name,
+                    array_to_string(attr_col_names, ''','''),
```
haradh1: Do we care empty  or NULL array?

cwelton: The `array_to_string(attr_col_names, ''',''')` probably doesn't work correctly with column names that require quote_ident(), however I strongly suspect that this is just one of many for the DT code, so I doubt that breaks anything new.

aojwang: It should not be null here. Null means no features. I add the check in function "MADLIB_SCHEMA.__check_training_table" before encoding.

We have 4 places need to use quote_ident and quote_literal. All of places were changed. I use the following table to do the test, it works well for training, classification, score and display. Please help double check the correctness of the changes. If I am wrong, please correct me. Thanks.

``` sql
testdb=# \d golf_2
           Table "public.golf_2"
   Column    |       Type       | Modifiers 
-------------+------------------+-----------
 id          | integer          | 
 out,look    | text             | 
 temperature | double precision | 
 humi'dity   | double precision | 
 win'dy      | text             | 
 class       | text             | 
Distributed by: (id)

```

src/config/Modules.yml
```diff
@@ -15,6 +15,7 @@ modules:
     - name: data_profile
       depends: ['sketch']
     - name: cart
+      depends: ['array_ops']
```
haradh1: Do we need this?

cwelton:
Looks good.

src/config/Modules.yml
aojwang: Sorry. It's my mistake. I double check with Yi and the src code. We needn't this anymore. Done. Thanks.

PR #210, opened 2012-08-21
Usability: Improved LMF output format

haying:
JIRA: MADLIB-578

After helping Sarah from the DS team to use this the first time, some usability is improved, mostly on making the output easier to be interpreted.


haradh1:
Is this for v0.5?  Do we have a jira?

haying:
I would count MADLIB-578 for this. Do we have to reopen it or create a new jira for this?

haradh1:
I just wanted to make sure the status of this PR.  If you think it's worth recording additional changes as a history that we can look up later, go for jira.  Especially for bugs (, though I don't see this is a bug)

haying:
This is more like making the feature complete than a bug. So I added 578 to the desc of this PR.

PR #211, opened 2012-08-22
Code Format Beautify

KevinYL:
Replace all tabs with 4 spaces. Using 4 spaces as indent character other than tabs. 
MADLIB-601 will be fixed with another pr.

### DIFF
--- a/src/ports/postgres/modules/quantile/quantile.sql_in
+++ b/src/ports/postgres/modules/quantile/quantile.sql_in
@@ -64,162 +64,162 @@ Module grp_countmin for an approximate quantile implementation.
  */
 CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.quantile_big(table_name TEXT, col_name TEXT, quantile FLOAT) RETURNS FLOAT AS $$
 declare
-    size FLOAT[];
-    increment INT := 0;
-    curr_old FLOAT;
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/quantile/quantile.sql_in
```diff
@@ -73,7 +73,7 @@ declare
   quantile_size FLOAT;
   full_size FLOAT;
   delta FLOAT;
-  rows_removed INT := 0;
+  rows_removed FLOAT := 0;
```
haradh1: Why not BIGINT?

KevinYL: Both the meaning and the content is the same as the fourth element of an float array:  last_values FLOAT[].

cwelton: Why can't we separate the float[] into separate variables so that we aren't lossy on the datatype.  We shouldn't be casting the bigint count(*) to a float8.

cwelton:
No changes other than whitespace, looks fine.

haradh1:
If you clean up the spaces, I'd suggest removing trailing spaces, too.

haradh1:
I'm going to merge this along with PR213.

PR #212, opened 2012-08-23
Fix MADLIB-641: Error information should be enhanced on data set without any rows

renyi533:
This pull request fix madlib-641. Please see http://jira.madlib.net/browse/MADLIB-641?focusedCommentId=11892#comment-11892 for details.


cwelton:
Looks fine.

PR #213, opened 2012-08-23
BugFix: MADLIB-601

KevinYL:
quantile_big uses INT/FLOAT to store the number of rows in the table.
If the number of rows in the table exceeds 2^31, the integer will
overflow.  At the same time, cast count(*) to FLOAT may cause issues.

To  fix it, we use INT8 instead.

### DIFF
--- a/src/ports/postgres/modules/quantile/quantile.sql_in
+++ b/src/ports/postgres/modules/quantile/quantile.sql_in
@@ -127,14 +127,12 @@ Begin
         
             
         IF(last_count=rows_removed) THEN
-            RAISE INFO 'first';
             /* 
                 If there are no more rows left, we exit.
             */
             EXIT;        
         
         ELSIF((increment > 0)AND(curr_old = last_count)) THEN
-            RAISE INFO 'second';
             /*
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/quantile/quantile.sql_in
```diff
@@ -64,162 +64,162 @@ Module grp_countmin for an approximate quantile implementation.
...
+            
+        IF(last_count=rows_removed) THEN
+            RAISE INFO 'first';
+            /* 
+                If there are no more rows left, we exit.
+            */
+            EXIT;        
+        
+        ELSIF((increment > 0)AND(curr_old = last_count)) THEN
+            RAISE INFO 'second';
```
haradh1: What is this for?

haradh1:
What is the status of this PR?  If you want me to remove RAISE sentences, I'll do it on merge.  Anyway give updates.

src/ports/postgres/modules/quantile/quantile.sql_in
KevinYL: I'll fix it later.

PR #214, opened 2012-08-23
score function return 'divide by zero' on madelon, since the class column is malformat.

aojwang:
JIRA: MADLIB-632.
The root cause is, that the dataset 'madelon_test' is wrong. The class labels of that dataset were missing. However, from dev perspective, we need to handle this exception. With the fix and the dataset, then the score should be zero, since the expected class labels are not the same with the classified labels at all.

```
testdb=# select distinct class from dt_madelon;
 class 
-------
 -1
 1
(2 rows)

testdb=# select distinct class from dt_madelon_test;
 class 
-------

(1 row)
```

We use left join to ensure all the rows of classification/scoring table were breakup, even the class labels in that table were not in training table. The class labels which don't appear in the training dataset will be treat as 0 (encoded value). Therefore, during scoring, those samples with 0 class label will be regarded misclassified.

### DIFF
--- a/methods/cart/src/pg_gp/c45.sql_in
+++ b/methods/cart/src/pg_gp/c45.sql_in
@@ -421,7 +421,7 @@ BEGIN
             verbosity,
             'tree'
         );
-    RAISE INFO '%', tree_table_name;
+        
     train_rs = MADLIB_SCHEMA.__encode_and_train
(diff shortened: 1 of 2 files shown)


cwelton:
Looks fine

PR #216, opened 2012-08-24
Make the result stable.

aojwang:
JIRA: MADLIB-461 and MADLIB-449.
We use the MADLIB_SCHEMA.svec_to_string(svec) as the unique identifier for a svec, so that the results are stable now. According to the definition of support and confidence, the support and confidence of the generated association rules should be >= give values rather than >. We also fix that. Furthurmore, we remove the dead code of the implementation.

### DIFF
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
@@ -213,7 +213,7 @@ select * from assoc_prod_uniq order by 1;
 The association rules are stored in the myschema.assoc_rules: 
 
 \code 
-testdb=# select * from myschema.assoc_rules order by support desc;
+select * from myschema.assoc_rules order by support desc;
  ruleid |       pre       |      post      |      support      |    confidence     |       lift        |    conviction     
 --------+-----------------+----------------+-------------------+-------------------+-------------------+-------------------
       4 | {diapers}       | {beer}         | 0.714285714285714 |                 1 |                 1 |                 0
@@ -331,23 +331,23 @@ DECLARE
   prod_min int; 
   prod_max int;
   r  MADLIB_SCHEMA.assoc_rules_results; 
-  total_rules int;  
+  total_rules int;
 BEGIN
-SET client_min_messages= warning; 
+SET client_min_messages= warning;
 
--- Stores data as one row per transaction (transactions saved as MADLIB_SCHEMA.svec type) 
-  EXECUTE 'DROP TABLE IF EXISTS assoc_trans_svec'; 
-  EXECUTE 'CREATE TEMPORARY TABLE assoc_trans_svec (trans_id text, products MADLIB_SCHEMA.svec)'; 
+-- Stores data as one row per transaction (transactions saved as MADLIB_SCHEMA.svec type)
+  EXECUTE 'DROP TABLE IF EXISTS assoc_trans_svec';
+  EXECUTE 'CREATE TEMPORARY TABLE assoc_trans_svec (trans_id text, products MADLIB_SCHEMA.svec)';
 
 -- Frequently occurring itemsets 
   EXECUTE 'DROP TABLE IF EXISTS assoc_rule_sets';
   EXECUTE 'CREATE TEMPORARY TABLE assoc_rule_sets (
     set_name  text 
     , set_list MADLIB_SCHEMA.svec
-    , hash_key TEXT
+    , text_svec TEXT
     , support FLOAT8 
     , iteration int) 
-  m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (hash_key)')  
+  m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (text_svec)')  
   ';
 
   -- the auxiliary table to generate the readable rules
@@ -498,24 +498,26 @@ LOOP
 
     EXECUTE 'DROP TABLE IF EXISTS list_b'; 
     EXECUTE 'CREATE TEMPORARY TABLE list_b AS
-    SELECT hash_key, set_list, tot_products
+    SELECT  MADLIB_SCHEMA.svec_to_string(set_list) text_svec, 
+            set_list, ' || n + 1 || ' as iteration
     FROM (
         SELECT
-            MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)) as set_list,
-            MADLIB_SCHEMA.svec_l1norm(MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list))) as tot_products, 
-            MADLIB_SCHEMA.svec_to_string(MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list))) as hash_key
+            MADLIB_SCHEMA.svec_minus(
+                MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), 
+                MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)
+            ) as set_list
         FROM
             assoc_rule_sets a 
        CROSS JOIN
             assoc_rule_sets b
        WHERE
-          a.hash_key <> b.hash_key
+          a.text_svec <> b.text_svec
           AND a.iteration = ' || n || ' 
           AND b.iteration = ' || n || '
-          AND MADLIB_SCHEMA.svec_l1norm(MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)))= ' || n+1 || '  
     ) t
+    WHERE MADLIB_SCHEMA.svec_l1norm(set_list) = ' || n + 1 || '
     GROUP BY 1, 2, 3
-    m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (hash_key)') 
+    m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (text_svec)') 
     '; 
 
   IF p_verbose is true THEN
@@ -524,28 +526,27 @@ LOOP
   END IF;  
 
 
-  EXECUTE 'INSERT INTO assoc_rule_sets (set_list, hash_key, support, iteration)
+  EXECUTE 'INSERT INTO assoc_rule_sets (set_list, text_svec, support, iteration)
   SELECT 
     c.set_list
-    , a.hash_key
+    , a.text_svec
     , support
     , iteration
   FROM (
     SELECT
-      a.hash_key
+      a.text_svec
       , (SUM(CASE WHEN MADLIB_SCHEMA.svec_contains(u.products,a.set_list) is TRUE THEN 1 ELSE 0 END))/b.tot::FLOAT8  AS support 
-      , MADLIB_SCHEMA.svec_l1norm(a.set_list) as iteration 
     FROM
       assoc_trans_svec u
     CROSS JOIN
       list_b a 
     CROSS JOIN
       max_t b 
-    GROUP BY 1,3, b.tot
+    GROUP BY 1, b.tot
     ) a
   JOIN
     list_b c
-     ON a.hash_key=c.hash_key
+     ON a.text_svec=c.text_svec
   WHERE
      support >= ' || i_support ;
   
@@ -582,7 +583,7 @@ LOOP
       a.set_list
       , b.set_list as product_x 
       , MADLIB_SCHEMA.svec_minus(a.set_list, b.set_list) as product_y 
-      , a.hash_key
+      , a.text_svec
       , a.support as support_xy 
       , b.support as support_x 
     FROM
@@ -595,7 +596,7 @@ LOOP
   ) c 
   JOIN
     assoc_rule_sets d
-      ON MADLIB_SCHEMA.svec_to_string(c.product_y)=d.hash_key
+      ON MADLIB_SCHEMA.svec_to_string(c.product_y)=d.text_svec
   WHERE
     c.support_xy/c.support_x >= ' || i_confidence || ' 
      AND MADLIB_SCHEMA.svec_l1norm(c.set_list)>=2'; 
@@ -631,17 +632,17 @@ LOOP
          ) t3
      WHERE t1.ruleId = t2.ruleId AND t1.ruleId = t3.ruleId';        
 
-  --EXECUTE 'DROP TABLE assoc_rules_aux_tmp';
+    EXECUTE 'DROP TABLE assoc_rules_aux_tmp';
 
-  IF p_verbose is true THEN 
-    EXECUTE 'SELECT count(*) FROM ' || output_schema || '.assoc_rules' into l; 
-    RAISE INFO '% Total association rules found', l; 
-  END IF; 
+ IF p_verbose is true THEN
+    EXECUTE 'SELECT count(*) FROM ' || output_schema || '.assoc_rules' into l;
+    RAISE INFO '% Total association rules found', l;
+  END IF;
 
   EXECUTE 'SELECT count(*) from ' || output_schema || '.assoc_rules' into total_rules;
-  EXECUTE 'SELECT '''|| output_schema || '''::text, ''assoc_rules''::text, ' || total_rules || '::int ' into r; 
+  EXECUTE 'SELECT '''|| output_schema || '''::text, ''assoc_rules''::text, ' || total_rules || '::int ' into r;
 
-  RETURN r ; 
+  RETURN r ;
 
 END;
 $$
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
@@ -104,19 +104,19 @@ The algorithm will map the product names to consective integer ids starting at 1
 - The results containing the rules, support, confidence, lift, and conviction are stored in the table assoc_rules in the schema specified by <em>output_schema</em>.
 <pre>
 	 Table "output_schema.assoc_rules"
-	    Column     |    Type     | Modifiers 
-	---------------+-------------+-----------
-	 set_list      | madlib.svec | 
-	 hash_key      | integer     | 
-	 subset_x      | madlib.svec | 
-	 subset_y      | madlib.svec | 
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -549,35 +496,27 @@ LOOP
...
-  FROM
-    list_a
-  GROUP BY 1,2,3
-  '; 
+    EXECUTE 'DROP TABLE IF EXISTS list_b'; 
+    EXECUTE 'CREATE TEMPORARY TABLE list_b AS
+    SELECT hash_key, set_list, tot_products
+    FROM (
+        SELECT
+            MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)) as set_list,
```
haradh1: Why don't you use '+' instead of svec_plus, '-' instead of svec_minus?

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -732,7 +631,7 @@ LOOP
          ) t3
      WHERE t1.ruleId = t2.ruleId AND t1.ruleId = t3.ruleId';        
 
-  EXECUTE 'DROP TABLE assoc_rules_aux_tmp';
+  --EXECUTE 'DROP TABLE assoc_rules_aux_tmp';
```
haradh1: If we don't need this, simply remove it.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -213,17 +213,17 @@ select * from assoc_prod_uniq order by 1;
...
- {1,1,1}:{1,0,1} |  -640301634 | {2,1}:{0,1}     | {1,2}:{1,0}     | 0.7142857142857142 | 1.0000000000000000 | 1.0000000000000000 |                  0 
- {1,1,1}:{1,0,1} |  -640301634 | {1,2}:{1,0}     | {2,1}:{0,1}     | 0.7142857142857142 | 0.7142857142857571 | 1.0000000000000000 | 1.0000000000000000
- {2,1}:{1,0}     |  1126614205 | {1,1,1}:{0,1,0} | {1,2}:{1,0}     | 0.4285714285714285 | 1.0000000000000000 | 1.0000000000000000 |                  0
- {3}:{1}         |  2106950333 | {2,1}:{1,0}     | {2,1}:{0,1}     | 0.2857142857142857 | 0.6666666666666667 | 0.9333333333333333 | 0.8142857142857142
- {3}:{1}         |  2106950333 | {1,1,1}:{0,1,0} | {1,1,1}:{1,0,1} | 0.2857142857142857 | 0.6666666666666667 | 0.9333333333333333 | 0.8142857142857142
- {3}:{1}         |  2106950333 | {1,2}:{0,1}     | {1,2}:{1,0}     | 0.2857142857142857 | 1.0000000000000000 | 1.0000000000000000 |                  0 
- {1,2}:{0,1}     | -2063121219 | {1,1,1}:{0,1,0} | {2,1}:{0,1}     | 0.2857142857142857 | 0.6666666666666667 | 0.9333333333333333 | 0.8142857142857142
+testdb=# select * from myschema.assoc_rules order by support desc;
```
haradh1: Remove "testdb=# "

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -549,35 +496,27 @@ LOOP
...
+            assoc_rule_sets a 
+       CROSS JOIN
+            assoc_rule_sets b
+       WHERE
+          a.hash_key <> b.hash_key
+          AND a.iteration = ' || n || ' 
+          AND b.iteration = ' || n || '
+          AND MADLIB_SCHEMA.svec_l1norm(MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)))= ' || n+1 || '  
+    ) t
+    GROUP BY 1, 2, 3
```
haradh1, line 527: Using column name or expression makes more stable.
HINT: but don't duplicate complicated expression here.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
cwelton: Well, if madlib isn't in the path then it needs to be a qualified operator, so:
`(a.set_list OPERATOR(madlib.+) b.set_list) OPERATOR(madlib.-) (a.set_list OPERATOR(madlib.*) b.set_list)`

Which I guess is a little more succinct than calling the svec_minus functions directly, but I can understand the reticence around schema qualified operators...

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -381,24 +339,13 @@ SET client_min_messages= warning;
...
-   '; 
-
 -- Frequently occurring itemsets 
   EXECUTE 'DROP TABLE IF EXISTS assoc_rule_sets';
   EXECUTE 'CREATE TEMPORARY TABLE assoc_rule_sets (
     set_name  text 
     , set_list MADLIB_SCHEMA.svec
-    , hash_key int
-    , support numeric 
+    , hash_key TEXT
```
cwelton: Calling a text representation of the sparse vector a hash_key seems somewhat misleading.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
```diff
@@ -634,21 +533,21 @@ LOOP
...
     SELECT
       a.hash_key
-      , (SUM(CASE WHEN MADLIB_SCHEMA.svec_contains(u.products,a.set_list) is TRUE THEN 1 ELSE 0 END))/b.tot::numeric  AS support 
+      , (SUM(CASE WHEN MADLIB_SCHEMA.svec_contains(u.products,a.set_list) is TRUE THEN 1 ELSE 0 END))/b.tot::FLOAT8  AS support 
       , MADLIB_SCHEMA.svec_l1norm(a.set_list) as iteration 
     FROM
       assoc_trans_svec u
     CROSS JOIN
-      list_c a 
+      list_b a
```
cwelton, line 550: Removing list_c in its entirety seems to be a rather substantial change here.  Can you provide additional explanation?

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang: It seems the svec module doesn't provide such operators.

``` sql
testdb=# select '{2, 1}:{0, 1}'::madlib.svec - '{1,1,1}:{1, 0, 1}'::madlib.svec;
ERROR:  operator does not exist: madlib.svec - madlib.svec
LINE 1: select '{2, 1}:{0, 1}'::madlib.svec - '{1,1,1}:{1, 0, 1}'::m...
                                            ^
HINT:  No operator matches the given name and argument type(s). You may need to add explicit type casts.
testdb=# select '{2, 1}:{0, 1}'::madlib.svec + '{1,1,1}:{1, 0, 1}'::madlib.svec;
ERROR:  operator does not exist: madlib.svec + madlib.svec
LINE 1: select '{2, 1}:{0, 1}'::madlib.svec + '{1,1,1}:{1, 0, 1}'::m...
```

According to the source code, it defines a lot of operators. However, they all starts with MADLIB_SCHEMA as schema name. I don't know the reason why we do that. If we want to make "+", "-", etc. work, we need to add the schema "MADLIB_SCHEMA" to the search path or remove the schema from those operator. I think we can't force the users to set the MADLIB_SCHEMA to the search path, so i prefer the latter one. What do you think.

``` sql
CREATE OPERATOR MADLIB_SCHEMA.- (
    LEFTARG = MADLIB_SCHEMA.svec,
    RIGHTARG = MADLIB_SCHEMA.svec,
    PROCEDURE = MADLIB_SCHEMA.svec_minus
);
CREATE OPERATOR MADLIB_SCHEMA.+ (
    LEFTARG = MADLIB_SCHEMA.svec,
    RIGHTARG = MADLIB_SCHEMA.svec,
    PROCEDURE = MADLIB_SCHEMA.svec_plus
);
```

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang: Should be removed.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang: Done.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang, line 550: Two reasons:
1) There is no need to generate the subsets of the itemsets in the current iteration, since we will generate all of them after the end of the loop. As you see from the code, we did self-join on assoc_rule_sets.
2) I did a full search for the text "assoc_set_list_parts", "tot_parts", "tot_parts1" and "tot_parts2", there are no reference on them. And list_c has the values in column "set_list" as the column in list_b.

```
  EXECUTE 'SELECT MADLIB_SCHEMA.assoc_svec_list_parts(''set_list '', '' list_b'', ''assoc_set_list_parts'')';

  EXECUTE 'DROP TABLE IF EXISTS list_c'; 
  EXECUTE 'CREATE TEMP TABLE list_c AS 
  SELECT
    d.set_list
    , c.*
  FROM (
    SELECT
      p.hash_key
      , count(*) as tot_parts 
      , count(a.hash_key) as tot_part1 
      , count(b.hash_key) as tot_part2 
    FROM
      assoc_set_list_parts p
    LEFT JOIN
      assoc_rule_sets a 
       ON MADLIB_SCHEMA.svec_hash(p.subset_x)=a.hash_key
    LEFT JOIN
      assoc_rule_sets b 
       ON MADLIB_SCHEMA.svec_hash(p.subset_y)=b.hash_key 
    WHERE
      MADLIB_SCHEMA.svec_l1norm(p.set_list)= '|| n+1 || '
    GROUP BY 1
    HAVING 
      count(*)=count(a.hash_key)
      AND count(*)=count(b.hash_key)
    ) c 
  JOIN
    list_b d 
      ON c.hash_key=d.hash_key 
  m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (hash_key)') 
  ';
```

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang: Right. It will be changed to text_svec.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang, line 527: We can not refer to the column name here, since the column is an expression `MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list))`.
One solution is, that we use a subquery to get a column, and we refer the column in the top query.  Please refer to the following code. But i don't know if it will bring to additional performance penalty. I will do a test for that. 

``` sql
    EXECUTE 'CREATE TEMPORARY TABLE list_b AS
    SELECT text_svec, set_list, tot_products
    FROM (
        SELECT MADLIB_SCHEMA.svec_to_string(set_list),
               set_list,
               tot_products
        FROM (
            SELECT
                MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)) as set_list,
                MADLIB_SCHEMA.svec_l1norm(MADLIB_SCHEMA.svec_minus(MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list))) as tot_products, 
            FROM
                assoc_rule_sets a 
           CROSS JOIN
                assoc_rule_sets b
           WHERE
              a.text_svec <> b.text_svec
              AND a.iteration = ' || n || ' 
              AND b.iteration = ' || n || '
        ) t
        WHERE MADLIB_SCHEMA.svec_l1norm(set_list)= ' || n+1 || '  
    ) t
    GROUP BY 1, 2, 3
    m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (text_svec)') 
    '; 
```

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang: Ok. I will use `OPERATOR(madlib.+)`. Thanks.

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
aojwang, line 527: It seems the optimizer will automatically merge the filter conditions to the subquery (please see the following query plan). As the subquery is more readable, i will use it.
- The query plan using subquery.

```
testdb-#     SELECT text_svec, set_list, tot_products
testdb-#     FROM (
testdb(#         SELECT madlib.svec_to_string(set_list) as text_svec,
testdb(#                set_list,
testdb(#                tot_products
testdb(#         FROM (
testdb(#             SELECT
testdb(#                 madlib.svec_minus(madlib.svec_plus(a.set_list, b.set_list), madlib.svec_mult(a.set_list, b.set_list)) as set_list,
testdb(#                 madlib.svec_l1norm(madlib.svec_minus(madlib.svec_plus(a.set_list, b.set_list), madlib.svec_mult(a.set_list, b.set_list))) as tot_products
testdb(#             FROM
testdb(#                 assoc_rule_sets a 
testdb(#            CROSS JOIN
testdb(#                 assoc_rule_sets b
testdb(#            WHERE
testdb(#               a.text_svec <> b.text_svec
testdb(#               AND a.iteration = 3 
testdb(#               AND b.iteration = 3
testdb(#         ) t1
testdb(#         WHERE madlib.svec_l1norm(set_list)= 4  
testdb(#     ) t2
testdb-#     GROUP BY 1, 2, 3
testdb-#     DISTRIBUTED BY (text_svec);
                                                                                                      QUERY PLAN                                                                                                       
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Redistribute Motion 1:4  (slice3; segments: 1)  (cost=8.41..8.48 rows=1 width=72)
   Hash Key: text_svec
   Rows out:  Avg 108.5 rows x 4 workers at destination.  Max 121 rows (seg1) with 78 ms to first row, 80 ms to end, start offset by 52 ms.
   ->  GroupAggregate  (cost=8.41..8.48 rows=1 width=72)
         Group By: "?column3?", "?column4?", "?column5?"
         Rows out:  434 rows with 85 ms to first row, 87 ms to end, start offset by 44 ms.
         ->  Sort  (cost=8.41..8.41 rows=4 width=64)
               Sort Key: "?column3?", "?column4?", "?column5?"
               Rows out:  2608 rows with 85 ms to end, start offset by 44 ms.
               Executor memory:  4169K bytes.
               Work_mem used:  4169K bytes.
               ->  Gather Motion 4:1  (slice2; segments: 4)  (cost=4.02..8.38 rows=1 width=64)
                     Rows out:  2608 rows at destination with 1.948 ms to first row, 63 ms to end, start offset by 45 ms.
                     ->  Nested Loop  (cost=4.02..8.21 rows=1 width=64)
                           Join Filter: madlib.svec_l1norm(madlib.svec_minus(madlib.svec_plus(a.set_list, b.set_list), madlib.svec_mult(a.set_list, b.set_list))) = 4::double precision AND a.text_svec <> b.text_svec
                           Rows out:  Avg 652.0 rows x 4 workers.  Max 1051 rows (seg2) with 0.670 ms to first row, 66 ms to end, start offset by 43 ms.
                           ->  Seq Scan on assoc_rule_sets a  (cost=0.00..3.96 rows=1 width=64)
                                 Filter: iteration = 3
                                 Rows out:  Avg 33.8 rows x 4 workers.  Max 55 rows (seg2) with 0.207 ms to first row, 0.279 ms to end, start offset by 43 ms.
                           ->  Materialize  (cost=4.02..4.06 rows=1 width=64)
                                 Rows out:  Avg 4556.2 rows x 4 workers.  Max 7425 rows (seg2) with 0.059 ms to first row, 2.223 ms to end of 55 scans, start offset by 112 ms.
                                 ->  Broadcast Motion 4:4  (slice1; segments: 4)  (cost=0.00..4.01 rows=1 width=64)
                                       Rows out:  Avg 135.0 rows x 4 workers at destination.  Max 135 rows (seg0) with 0.742 ms to first row, 2.098 ms to end, start offset by 42 ms.
                                       ->  Seq Scan on assoc_rule_sets b  (cost=0.00..3.96 rows=1 width=64)
                                             Filter: iteration = 3
                                             Rows out:  Avg 33.8 rows x 4 workers.  Max 55 rows (seg2) with 0.171 ms to first row, 0.220 ms to end, start offset by 44 ms.
 Slice statistics:
   (slice0)    Executor memory: 345K bytes avg x 4 workers, 345K bytes max (seg0).
   (slice1)    Executor memory: 441K bytes avg x 4 workers, 441K bytes max (seg0).
   (slice2)    Executor memory: 359K bytes avg x 4 workers, 381K bytes max (seg1).
   (slice3)    Executor memory: 4581K bytes (seg0).  Work_mem: 4169K bytes max.
 Statement statistics:
   Memory used: 128000K bytes
 Total runtime: 133.371 ms
(34 rows)

```

src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
haradh1: Ugh.  I don't like `OPERATOR(madlib.+)`.  It doesn't improve readability.

Off topic but I always wonder why we don't set search_path to madlib at the beginning of the driver function.

aojwang: I can set the search path at the beginning of function and restore it at the end. Please see the following code.

``` sql
    EXECUTE 'SHOW search_path' INTO searchpath_text;
    EXECUTE 'SET search_path = ' || searchpath_text || ', MADLIB_SCHEMA';
    ......
    EXECUTE 'SET search_path = ' || searchpath_text;
```
- As we need to concat the string,  we should consider the special chars. I try to add schema "a'b" to the search path, however, it failed. Any suggestions?

``` sql
testdb=# set search_path="a'b", public;
ERROR:  syntax error at or near "b', '"  (seg0 aojwang:40000 pid=12297)

testdb=# set search_path="a,b", public;
SET

```

haradh1: Ah, wait.  I meant it's just a thought.  If we do it, we should probably do it for all modules, possibly in abstraction layer.  Need a right design.  Thus, I'm ok with the current one.

PR #218, opened 2012-08-27
Association Rules: Avoid hash aggregate on svec

haradh1:
Jira: MADLIB-461

GPDB apparently chooses hash aggregate although svec doesn't support it.
Use text type to make the plan correct.

### DIFF
--- a/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
+++ b/src/ports/postgres/modules/assoc_rules/assoc_rules.sql_in
@@ -507,7 +507,9 @@ LOOP
     EXECUTE 'DROP TABLE IF EXISTS list_b'; 
     EXECUTE 'CREATE TEMPORARY TABLE list_b AS
     SELECT  MADLIB_SCHEMA.svec_to_string(set_list) text_svec, 
-            set_list, ' || n + 1 || ' as iteration
+            MADLIB_SCHEMA.svec_from_string(
+                MADLIB_SCHEMA.svec_to_string(set_list)) set_list,
+            ' || n + 1 || ' as iteration
     FROM (
         SELECT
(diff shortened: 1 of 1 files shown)


haradh1:
Assign @aojwang to look at the fix.

aojwang:
How about we avoid hash aggregate on svec by converting it to text in subquery. Please refer to the following code, it will invoke svec_to_string once and invoke svec_from_string twice. Thanks. I will check if the PR-206 has the same problem.

```
    EXECUTE 'DROP TABLE IF EXISTS list_b'; 
    EXECUTE 'CREATE TEMPORARY TABLE list_b AS
    SELECT  text_svec, 
            MADLIB_SCHEMA.svec_from_string(text_svec) as set_list, 
            ' || n + 1 || ' as iteration
    FROM (
        SELECT
            MADLIB_SCHEMA.svec_to_string(
                MADLIB_SCHEMA.svec_minus(
                    MADLIB_SCHEMA.svec_plus(a.set_list, b.set_list), 
                    MADLIB_SCHEMA.svec_mult(a.set_list, b.set_list)
                ) 
            ) as text_svec
        FROM
            assoc_rule_sets a 
       CROSS JOIN
            assoc_rule_sets b
       WHERE
          a.text_svec <> b.text_svec
          AND a.iteration = ' || n || ' 
          AND b.iteration = ' || n || '
    ) t
    WHERE MADLIB_SCHEMA.svec_l1norm(MADLIB_SCHEMA.svec_from_string(text_svec)) = ' || n + 1 || '
    GROUP BY text_svec
    m4_ifdef(`GREENPLUM',`DISTRIBUTED BY (text_svec)') 
    ';
```

haradh1:
I see.  I'll make a subquery.

PR #222, opened 2012-08-31
Add the check for the maximum number of features in the training table.

aojwang:
JIRA: MADLIB-602.
If the number of features of a dataset is greater than 102400, then we will report error. If the table is empty, we will report an error too.


haradh1:
We'd better mention the limitation in the documentation.

methods/kernel_machines/src/pg_gp/online_sv.py_in
```diff
@@ -23,8 +38,10 @@ def svm_regression( madlib_schema, input_table, model_table, parallel, kernel_fu
     # Output error if model_table already exist
     # if __check_rel_exist(model_table):
     #    plpy.error('Table ' + model_table + ' exists; please use a different model table or drop ' + model_table + ' before calling this function.');
-
     # plpy.execute('drop table if exists ' + model_table);
+
+    __validate_input_table(input_table);
```
aojwang, line 43: SVM regression should also have such limitation.

PR #228, opened 2012-09-07
Replace GREENPLUM macros in m4 with __GREENPLUM__

iyerr3:
JIRA: MADLIB 639

Files updated: 
 src/ports/postgres/modules/assoc_rules/assoc_rules.py_in
 src/ports/postgres/modules/bayes/bayes.sql_in
 src/ports/postgres/modules/convex/lmf.sql_in
 src/ports/postgres/modules/kmeans/kmeans.sql_in
 src/ports/postgres/modules/regress/linear.sql_in
 src/ports/postgres/modules/regress/logistic.sql_in
 src/ports/postgres/modules/regress/multilogistic.sql_in
 src/ports/postgres/modules/svd_mf/svdmf.py_in

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -15,7 +15,6 @@ function(define_m4_macros OUT_M4_CMD_LINE OUT_M4_CODE)
     set(IN_FEATURES ${ARGN})
 
     set(MACROS
-        "${PORT_UC}"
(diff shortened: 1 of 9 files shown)


haradh1:
I believe somewhere we define "GREENPLUM" in cmake files.  I'd like to remove it so that someone won't introduce it again in the future.

PR #231, opened 2012-09-14
Build System: Replace GREENPLUM macro, take 2

haradh1:
Jira: MADLIB-666


cwelton:
Looks good!   If this get's everything can we remove the historical GREENPLUM macro definition as well?

fschopp:
Hmm, the old definition of GREENPLUM had already been removed with commit 7f82cab. Interesting that it worked at all.

haradh1:
Most of the case were for the better performance in GPDB.

fschopp:
I think you forgot to the m4_include(`SQLCommon.m4')

haradh1:
In which file?

PR #233, opened 2012-09-20
SVM: fix madlib 677. 

renyi533:
In parallel mode, we build one sub-model for each segment. In the combination step, we should skip the segments with no source data since we do not build sub-model at all if one segment has no source data.


cwelton:
Looks reasonable.  The larger questions about using segments as a logical unit of calculation still need to be addressed, but we can handle that at a later point.

PR #235, opened 2012-09-21
mullogistic: fix madlib 670.

renyi533:
Add more error checking for negative test cases. 

Make the algorithm's behavior consistent with the document.

### DIFF
--- a/src/modules/regress/multilogistic.cpp
+++ b/src/modules/regress/multilogistic.cpp
@@ -237,8 +237,8 @@ mlogregr_irls_step_transition::run(AnyType &args) {
         if (numCategories < 1)
                 throw std::domain_error("Number of cateogires must be at least 2");
 
-        if (category > numCategories)
-                throw std::domain_error("You have entered a category > numCategories"
+        if (category > numCategories || category<0)
+                throw std::domain_error("You have entered an invalid category. "
                     "Categories must be of values {0,1... numCategories-1}");
 
         // Init the state (requires x.size() and category.size())
--- a/src/modules/regress/multilogistic.cpp
+++ b/src/modules/regress/multilogistic.cpp
@@ -237,10 +237,6 @@ mlogregr_irls_step_transition::run(AnyType &args) {
         if (numCategories < 1)
                 throw std::domain_error("Number of cateogires must be at least 2");
 
-        if (category > numCategories || category<0)
-                throw std::domain_error("You have entered an invalid category. "
-                    "Categories must be of values {0,1... numCategories-1}");
-
         // Init the state (requires x.size() and category.size())
         state.initialize(*this, static_cast<uint16_t>(x.size())
                                                     , static_cast<uint16_t>(numCategories));
@@ -252,8 +248,13 @@ mlogregr_irls_step_transition::run(AnyType &args) {
         }
     }
 
-
-
+    /*  
+     * This check should be done for each iteration. Only checking the first 
+     * run is not enough.
+     */ 
+    if (category > numCategories || category < 0)
+        throw std::domain_error("You have entered an invalid category. "
+            "Categories must be of values {0,1... numCategories-1}");
 
     // Now do the transition step
     state.numRows++;
--- a/src/ports/postgres/modules/regress/multilogistic.sql_in
+++ b/src/ports/postgres/modules/regress/multilogistic.sql_in
@@ -380,7 +380,7 @@ BEGIN
     EXECUTE stmt INTO rec;
 
     IF (rec IS NOT NULL) THEN
-        RAISE EXCEPTION 'you have entered an invalid category outside of [%, %]', 
+        RAISE EXCEPTION 'category must be in [%, %]', 
             0, numcategories-1;
     END IF;
     
--- a/src/ports/postgres/modules/regress/multilogistic.sql_in
+++ b/src/ports/postgres/modules/regress/multilogistic.sql_in
@@ -372,18 +372,7 @@ BEGIN
     IF (rec IS NULL) THEN
         RAISE EXCEPTION 'the type of (%) in (%) must be integer', depvar, source;
     END IF;
-    
-    stmt= 'SELECT *  
-    FROM '||source||'  
-    WHERE '||depvar||'<0 OR '||depvar||'>'||numcategories-1||
-    ' LIMIT 1;';
-    EXECUTE stmt INTO rec;
-
-    IF (rec IS NOT NULL) THEN
-        RAISE EXCEPTION 'category must be in [%, %]', 
-            0, numcategories-1;
-    END IF;
-    
+        
     theIteration := (
         SELECT MADLIB_SCHEMA.compute_mlogregr($1, $2, $3, $4, $5, $6, $7)
     );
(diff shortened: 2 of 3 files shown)


src/ports/postgres/modules/regress/multilogistic.sql_in
```diff
@@ -346,7 +346,44 @@ DECLARE
...
+                            ]
+                    )
+          ) AND 
+          attname IN ('||quote_literal(depvar)||') LIMIT 1;';
+
+    EXECUTE stmt INTO rec;
+
+    IF (rec IS NULL) THEN
+        RAISE EXCEPTION 'the type of (%) in (%) must be integer', depvar, source;
+    END IF;
```
cwelton: This seems like an overly complex method of type verification, I do not understand why you can't simply rely on the databases type verification.   E.g. both "smallint" and "int" have built in casts to "bigint", no extra code written.

If the problem is that there is no automatic cast from "bigint" to "int" that is because it is a lossy conversion and you shouldn't force that to be automatic for the same reason that forcing the cast from float is a bad idea.

src/ports/postgres/modules/regress/multilogistic.sql_in
```diff
@@ -346,7 +346,44 @@ DECLARE
...
+    
+    stmt= 'SELECT *  
+    FROM '||source||'  
+    WHERE '||depvar||'<0 OR '||depvar||'>'||numcategories-1||
+    ' LIMIT 1;';
+    EXECUTE stmt INTO rec;
```
cwelton: Couldn't this check be part of the transition function so that we don't have to add another full scan of the table?

renyi533: OK. After the modification, we can check it within transition function. Initially, we only check the first run. Actually, we should check the category for each iteration. I will fix this logic error.

src/ports/postgres/modules/regress/multilogistic.sql_in
renyi533: The big problem is related with the error message's information. If I remove the code above and rely on the type checking of the system. The error message will be like below.

WARNING:  PL/Python: in PL/Python function "compute_mlogregr"
DETAIL:  <class 'plpy.SPIError'>: unrecognized error in PLy_spi_execute_query
CONTEXT:  SQL statement "SELECT  ( SELECT madlib.compute_mlogregr( $1 ,  $2 ,  $3 ,  $4 ,  $5 ,  $6 ,  $7 ) )"
PL/pgSQL function "mlogregr" line 32 at assignment
ERROR:  function madlib.mlogregr_irls_step(double precision, integer, double precision[], double precision[]) does not exist
LINE 6:             madlib.mlogregr_irls_step(
                    ^
HINT:  No function matches the given name and argument types. You may need to add explicit type casts.
QUERY:  
        INSERT INTO _madlib_iterative_alg
        SELECT
            1,

```
        madlib.mlogregr_irls_step(
            (y),
            (3),
            (x)::FLOAT8[],
            (st._madlib_state)
        )

    FROM
        _madlib_iterative_alg AS st,
        madlibtestdata.mlogr_incorrecty_decimal_observations AS src
    WHERE
        st._madlib_iteration = 1 - 1
```

CONTEXT:  SQL statement "SELECT  ( SELECT madlib.compute_mlogregr( $1 ,  $2 ,  $3 ,  $4 ,  $5 ,  $6 ,  $7 ) )"
PL/pgSQL function "mlogregr" line 32 at assignment

If we check with the code above, the error message will be much more helpful for users. If you think the first error message is good enough, I will remove the checking logic above.

cwelton: What version are you running in when you get the above error?  This should have been fixed by some improvements to plpython in later versions of GP and Postgres.

renyi533: I used gpdb 4.1 in my laptop. I will try gpdb 4.2 and check the error message again.

renyi533: I also tried gpdb 4.2.2 in redhat server. The error message is still the same. Please see the message below. The error message is not very helpful for users.

ERROR:  plpy.SPIError: function madlib.mlogregr_irls_step(double precision, integer, double precision[], double precision[]) does not exist (plpython.c:4648)
LINE 6:             madlib.mlogregr_irls_step(
                    ^
HINT:  No function matches the given name and argument types. You may need to add explicit type casts.
QUERY:  
        INSERT INTO _madlib_iterative_alg
        SELECT
            1,

```
        madlib.mlogregr_irls_step(
            (y),
            (3),
            (x)::FLOAT8[],
            (st._madlib_state)
        )

    FROM
        _madlib_iterative_alg AS st,
        madlibtestdata.mlogr_incorrecty_decimal_observations AS src
    WHERE
        st._madlib_iteration = 1 - 1
```

CONTEXT:  Traceback (most recent call last):
  PL/Python function "compute_mlogregr", line 27, in <module>
    return multilogistic.compute_mlogregr(**globals())
  PL/Python function "compute_mlogregr", line 178, in compute_mlogregr
  PL/Python function "compute_mlogregr", line 96, in __runIterativeAlg
PL/Python function "compute_mlogregr"
SQL statement "SELECT  ( SELECT madlib.compute_mlogregr( $1 ,  $2 ,  $3 ,  $4 ,  $5 ,  $6 ,  $7 ) )"
PL/pgSQL function "mlogregr" line 32 at assignment

haradh1: I wonder why your quote doesn't show explicit cast, although compute_mlogregr says

```
        updateExpr = """
            {schema_madlib}.mlogregr_{optimizer}_step(
                ({depvar})::INTEGER,
                ({numcategories})::INTEGER,
                ({indepvar})::FLOAT8[],
                {{state}}
            )
```

haradh1: And why do you care only depvar?  Is this same argument for numcategories, indepvar, too?

I agree the error message is not friendly, but this could solve part of the problem, I guess, so I'm fine with the unpleasant error message rather than adding a hack to check the type.

renyi533: In my local copy, I temporarily removed the explicit cast. So I can get the system reported error.

If we think the system error message is acceptable, I will remove the checking code and the explicit cast.

src/modules/regress/multilogistic.cpp
```diff
@@ -252,8 +248,13 @@ class MLogRegrIRLSTransitionState {
         }
     }
 
-
-
+    /*  
+     * This check should be done for each iteration. Only checking the first 
+     * run is not enough.
+     */
```
fschopp, line 254: True,  though this is rather a commit comment than a source-code comment.

src/modules/regress/multilogistic.cpp
```diff
@@ -252,8 +248,13 @@ class MLogRegrIRLSTransitionState {
...
 
-
-
+    /*  
+     * This check should be done for each iteration. Only checking the first 
+     * run is not enough.
+     */ 
+    if (category > numCategories || category < 0)
+        throw std::domain_error("You have entered an invalid category. "
+            "Categories must be of values {0,1... numCategories-1}");
```
fschopp, line 257: This error message is unlike all other error messages we have. We do not use "you" in error messages.

iyerr3, line 257: Please let me know if you're planning to address Florian's comments so that I wait before merging.

fschopp:
This was merged with commit 8bc6c95.

PR #237, opened 2012-09-23
Build system: Option LIBSTDCXX_COMPAT for compatibility with an old libstdc++

fschopp:
Jira: MADLIB-263

This commit follows ideas proposed here: http://glandium.org/blog/?p=1901
Unfortunately, MADlib seems to use libstdc++ to a greater extend than Firefox 4.0
did, so we need to do a bit more.

The declarations and definitions in this commit make it possible to build MADlib
with the following versions of gcc (please add to the list), while continuing to
only rely on libstdc++.so.6.0.8 (which corresponds to gcc 4.1.2, and labels
GLIBCXX_3.4.8, CXXABI_1.3.1).

Tested with the following versions of gcc:
- gcc 4.4.2
- gcc 4.5.4
- gcc 4.6.2

It should most likely also work with other recent gcc versions.

To use, add the option "-DLIBSTDCXX_COMPAT=40102" when running cmake.

### DIFF
--- a/ReadMe_Build.txt
+++ b/ReadMe_Build.txt
@@ -134,6 +134,21 @@ root directory) for more options, after having run `cmake` the first time.
     Note: If no `GREENPLUM<...>_PG_CONFIG` is specified, the build script will
     look for `/usr/local/greenplum-db/bin/pg_config`.
 
+- `LIBSTDCXX_COMPAT` (default: *empty*)
+
+    If GNU gcc is used to build MADlib and link against the GNU libstdc++, this
+    option may be used to set the maximum version of libstdc++ acceptable as a
+    runtime dependency (not supported on Mac OS X). E.g., if MADlib should
+    require no more than the libstdc++ shipped with gcc 4.1.2, call
+    `./configure` with `-DLIBSTDCXX_COMPAT=40102`.
+    
+    The current minimum value supported for option `LIBSTDCXX_COMPAT` is
+    `40102`, and the latest version of gcc supported when setting this option is
+    gcc 4.6.x.
+    
+    Setting this option will enable workarounds in
+    `src/utils/libstdcxx-compatibility.cpp`.
+
 - `BOOST_TAR_SOURCE` (default: *empty*)
 
     If no recent version of Boost is found (>= 1.46), Boost is downloaded
--- a/license/third_party/_M_widen_init.txt
+++ b/license/third_party/_M_widen_init.txt
@@ -1,0 +1,20 @@
+std::ctype<char>::_M_widen_init() is a function authored by Jerry Quinn
+<jlquinn@optonline.net>, which was added to libstdc++ with revision 74662 on
+Dec 16, 2003 [1].
+
+With permission from Jerry (thankfully received on Oct 9, 2012), we include a
+copy of this function in the MADlib repository. The sole intention is to allow
+compiling MADlib with recent versions of gcc while still keeping the runtime
+dependencies limited to earlier versions of libstdc++. Technical details are
+given in src/utils/libstdcxx-compatibility.cpp.
+
+Revision 74662 of the libstdc++-v3 file include/bits/locale_facets.h, where
+std::ctype<char>::_M_widen_init() has been copied from, also included the
+following notice in the file header [2]:
+
+// As a special exception, you may use this file as part of a free software
+// library without restriction. [...]
+
+Links:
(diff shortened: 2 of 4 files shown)


src/ports/linux/utils/libstdc++.so.6.0.8-compatibility.cpp
```diff
@@ -0,0 +1,149 @@
...
+ * following.
+ *
+ * The type __cxxabiv1::__forced_unwind was only introduced in May 2007 (see
+ * attachments to the previous bug report) and thus after the release of
+ * gcc 4.1.2 (Feb 13, 2007, see http://gcc.gnu.org/releases.html).
+ *
+ * As http://gcc.gnu.org/onlinedocs/gcc/Vague-Linkage.html explains:
+ * <blockquote>
+ *     If the class declares any non-inline, non-pure virtual functions, the
+ *     first one is chosen as the “key method” for the class, and the vtable is
```
fschopp: Change “” to "".

src/ports/linux/utils/libstdc++.so.6.0.8-compatibility.cpp
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * @brief Write a value to an ostream
+ *
+ * In recent versions of libstdc++, \c _M_insert contains the implementation for
+ * the various operator<<() overloads. Now, as http://glandium.org/blog/?p=1901
+ * explains, newer libstdc++ versions contain various instantiations for
+ * \c _M_insert, even though <bits/ostream.tcc> contains a general (template)
+ * definition.
+ *
+ * libstdc++.so.6.0.8 did not contain implementations for for \c _M_insert, so
```
fschopp: remove 2nd "for"

cwelton:
All of Kevin's comments have been addressed, based on his assessment it looks good.

fschopp:
Another possible legal concern is the function `_M_widen_init`. Technically, it is a few lines of code from libstdc++. Obviously, it is trivial, short, and was previously an inline function, i.e., it was _always_ compiled into earlier versions of C++ programs. Also, when this function was first added to the code base, it [contained the following header information](http://gcc.gnu.org/viewcvs/trunk/libstdc%2B%2B-v3/include/bits/locale_facets.h?diff_format=h&r1=74662&r2=74661&pathrev=74662):

```
// As a special exception, you may use this file as part of a free software
// library without restriction.
```

So from my perspective, including this code, which is otherwise under the GPL, is _not_ a problem. Rewriting is of course not a sensible option because this function deals with an internal data structure, and thus could not be rewritten without looking at libstdc++ source code.

cwelton:
Good point on _M_widen_init.  We should include the full text of the license exception with the file:

```
Portions of this file are copied from /trunk/libstdc++-v3/include/bits/locale_facets.h and are subject
to the Gnu Public License (GPL) with the following exception:

// As a special exception, you may use this file as part of a free software
// library without restriction. Specifically, if other files instantiate
// templates or use macros or inline functions from this file, or you compile
// this file and link it with other files to produce an executable, this
// file does not by itself cause the resulting executable to be covered by
// the GNU General Public License. This exception does not however
// invalidate any other reasons why the executable file might be covered by
// the GNU General Public License. 
```

fschopp:
In my understanding, the [current libstdc++ license exception](http://gcc.gnu.org/onlinedocs/libstdc++/manual/license.html) does _not_ immediately and unambiguously allow the use of libstdc++ code in other products, as it only applies to "Eligible Compilation Processes". I see three options instead:
1. Use older code which still had the cited notice in the header file.
2. Ask the original author of `_M_widen_init` ([Jerry Quinn, see this link](http://gcc.gnu.org/ml/libstdc++/2003-12/msg00165.html)) for permission.
3. (Currently chosen.) Take the viewpoint that the older code explicitly allowed the use in free software, that the modifications are trivial, that the code is short and trivial altogether, and that therefore `_M_widen_init` bears no license restrictions.

In any case, I agree that the license question should be explicitly addressed in the source file.

fschopp:
Of course, 'trivial' in my previous posts refers to the code copied, not the original change to libstdc++ by Jerry Quinn.

fschopp:
I am very happy that Jerry Quinn gave explicit permission to use the `_M_widen_init` function from [his code](http://gcc.gnu.org/viewcvs?diff_format=h&view=revision&revision=74662). Thanks a lot!

I updated the pull request accordingly and also added a file in the license directory, because that is the place where we store all references to third-party code. Comments?

PR #240, opened 2012-09-29
Install create_table_as() workaround also on PostgreSQL

fschopp:
Jira: MADLIB-691

It is convenient to have a uniform interface for CREATE TABLE ... AS, e.g.,
when writing test cases that should run unmodified on all supported
platforms. Obviously, we then can rely on nothing more than the least common
denominator – which is that CTAS is defunct, and our SELECT create_table_as()
workaround has to be used.

In other words, the compatibility module should also be installed for the
PostgreSQL port.


haradh1:
Looks good

PR #241, opened 2012-10-02
Build support for PostgreSQL 9.2

fschopp:
Jira: MADLIB-662

The decision-tree code contained the function table_exists(), which calls
RangeVarGetRelid(), which used to be a two-parameter function, but is now a
three-paremter macro.

### DIFF
--- a/methods/cart/src/pg_gp/dt.c
+++ b/methods/cart/src/pg_gp/dt.c
@@ -2334,7 +2334,12 @@ Datum table_exists(PG_FUNCTION_ARGS)
     input = PG_GETARG_TEXT_PP(0);
 
     names = textToQualifiedNameList(input);
-    relid = RangeVarGetRelid(makeRangeVarFromNameList(names), true);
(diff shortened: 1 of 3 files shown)


haradh1:
Looks good.

methods/cart/src/pg_gp/dt.c
```diff
@@ -2334,7 +2334,12 @@ Datum table_exists(PG_FUNCTION_ARGS)
     input = PG_GETARG_TEXT_PP(0);
 
     names = textToQualifiedNameList(input);
-    relid = RangeVarGetRelid(makeRangeVarFromNameList(names), true);
+    relid = RangeVarGetRelid(
+        makeRangeVarFromNameList(names),
+#if PG_VERSION_NUM >= 90200
+        NoLock,
+#endif
```
fschopp, line 2341: Actually, this is not the right thing to do and will give the following warning on PostgreSQL 9.2: `warning: embedding a directive within macro arguments is not portable`.

PR #243, opened 2012-10-05
Design document updates, k-means naming improvements

fschopp:
Changes:
- Fix compilation warning introduced with commit 845210b
- With k-means allowing to supply arbitrary distance functions and recentering
  aggregates, the name k-means has become a bit of a misnomer. E.g., by supplying
  the 1-norm as distance function and the element-wise median as recentering
  aggregate we would compute a local minimum for the k-medians problem.
  Due to the previous narrow focus on only the Euclidean case, the distance
  parameter used to be called 'fn_squared_dist' and the recentering parameter
  'agg_mean'. I replace them by 'fn_dist' and 'agg_centroid'.
- Along the same lines, I removed the distance functions 'squared_dist_1norm',
  'squared_angle', and 'squared_tanimoto', and only kept 'dist_1norm', 'dist_angle',
  and 'dist_tanimoto'.
- Added default parameters to madlib.closest_column
- Added author and version information to some modules in the design document

Jiras:
- MADLIB-692, MADLIB-662

### DIFF
--- a/doc/design/design.tex
+++ b/doc/design/design.tex
@@ -34,6 +34,7 @@
 \usepackage{xspace}                    % Correct spaces after \newcommand definitions
 \usepackage[noend]{algpseudocode}      % algorithm environment
 \usepackage{listings}                  % Code snippets
+\usepackage{bbding}
 
 
 % BEGIN Doc Layout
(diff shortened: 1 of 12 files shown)


iyerr3:
Everything looks good.

PR #246, opened 2012-10-09
Added check for invalid kmeans input parameters

geeg:
Check for invalid max iteration, convergence threshold, and k values.

### DIFF
--- a/src/ports/postgres/modules/kmeans_new/kmeans_new.sql_in
+++ b/src/ports/postgres/modules/kmeans_new/kmeans_new.sql_in
@@ -390,6 +390,12 @@ BEGIN
         FROM pg_proc WHERE oid = proc_agg_centroid) THEN
         RAISE EXCEPTION 'Mean aggregate has wrong signature or is not an aggregate.';
     END IF;
+    IF (min_frac_reassigned < 0) OR (min_frac_reassigned > 1) THEN
+        RAISE EXCEPTION 'Convergence threshold is not a valid value (must be a fraction between 0 and 1).';
+    END IF;
+    IF (max_num_iterations < 0) THEN
+        RAISE EXCEPTION 'Number of iterations must be a non-negative integer.';
+    END IF;
 
     -- We first setup the argument table. Rationale: We want to avoid all data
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/kmeans_new/kmeans_new.sql_in
```diff
@@ -390,6 +390,12 @@ BEGIN
         FROM pg_proc WHERE oid = proc_agg_centroid) THEN
         RAISE EXCEPTION 'Mean aggregate has wrong signature or is not an aggregate.';
     END IF;
+    IF (min_frac_reassigned < 0) OR (min_frac_reassigned > 1) THEN
+        RAISE EXCEPTION 'Convergence threshold is not a valid value (must be a fraction between 0 and 1).';
+    END IF;
+    IF (max_num_iterations < 0) THEN
+        RAISE EXCEPTION 'Number of iterations must be a non-negative integer.';
+    END IF;
```
cwelton, line 398: Is 0 a valid number of iterations?

cwelton:
Looks like a fix for MADLIB-361 and MADLIB-359, updating the jiras accordingly would be good.

geeg:
Yep, about to update the JIRAs.

I think it may be left up to debate about whether or not 0 is a valid number of iterations (technically kmeans could return with some sort of seeding of centroids w/o actually going through a single iteration.  That was my understanding of Florian's code.).

src/ports/postgres/modules/kmeans_new/kmeans_new.sql_in
iyerr3, line 398: Technically, the seeding happens before any iterations - so 0 iterations should return back only the seeding output.

fschopp, line 398: 0 would be the result immediately after the seeding phase. No need to disallow that.

iyerr3:
Madmark tests for bad input returned the right result. Amended the commit and merged: https://github.com/madlib/madlib/commit/aa1f657d46222f60b64ac509bce6d799866c4d93

PR #247, opened 2012-10-10
CRF module: fix digit word features and add install-check tests

geeg:
Jira: MADLIB-699 (http://jira.madlib.net/browse/MADLIB-699)
Fix CRF module to only have one 'W_DIGIT' word feature to represent any digit, and add additional install-check tests


cwelton:
Grace please read https://github.com/madlib/madlib/wiki/Contribution-Guide and update the description with jira number and description of issue.

cwelton:
Kun can you supply some review for these changes?

geeg:
I will also edit the code so that the feature generation function doesn't explicitly change the original segment table that is provided by the user.  Otherwise, the user will lose the actual digits in the segment table, which will be replaced by the word 'DIGIT.'

src/ports/postgres/modules/crf/test/crf_train_small.sql_in
```diff
@@ -149,6 +177,5 @@ $$ language plpgsql;
 -- Test: 
 ---------------------------------------------------------------------------
 SELECT crf_train_install_test();
```
fschopp: Hmm. Is is really necessary to hide the test in a pgsql function, and then call that function?

fschopp: The much preferred way of writing unit tests is [like here](https://github.com/madlib/madlib/blob/master/src/ports/postgres/modules/regress/test/linear.sql_in).

src/ports/postgres/modules/crf/crf_feature_gen.sql_in
```diff
@@ -122,10 +122,18 @@ $$
     plpy.execute("DROP TABLE IF EXISTS MADLIB_SCHEMA.tmp_featureset;" + \
                  "CREATE TABLE MADLIB_SCHEMA.tmp_featureset(f_name text, feature integer[]);")
```
fschopp: I'm just seeing this now. We really should not create tables in the madlib schema.

src/ports/postgres/modules/crf/test/crf_test_large.sql_in
```diff
@@ -0,0 +1,766 @@
...
+	SELECT INTO result CASE WHEN (result_count = 0) THEN 'PASS' ELSE 'FAIL' END;
+
+	IF result = 'FAIL' THEN
+	   RAISE EXCEPTION 'Failed install check with % mislabels.', result_count;
+	END IF;
+
+	RETURN result;
+
+END
+$$ language plpgsql;
```
fschopp: This should be rewritten in plain SQL using assert.

src/ports/postgres/modules/crf/viterbi.sql_in
```diff
@@ -57,9 +57,16 @@ MADLIB_SCHEMA.vcrf_label(segtbl text, factor_mtbl text, factor_rtbl text, labelt
   $$
   m_factors = "_m_factors"
   r_factors = "_r_factors"
+  segtbl_digits = "_segtbl_digits"
   resulttbl_raw = resulttbl + "_raw"
 
-  plpy.execute("DROP TABLE IF EXISTS " + m_factors +","+ r_factors + "," + resulttbl_raw +" CASCADE;")
+  plpy.execute("DROP TABLE IF EXISTS " + m_factors +","+ r_factors + "," + resulttbl_raw + "," + segtbl_digits + " CASCADE;")
```
fschopp: Hmm, really calling DROP TABLE CASCADE should be a means of last resort. Is it really necessary? If not, it should probably only be called with schema-qualified names. Otherwise, this is potentially very dangerous.

fschopp:
Once everybody agrees that this is ready, this pull request should probably be melded into one commit before merging.

geeg:
Thanks for your comments, Florian.  Will work on integrating your suggestions tonight.  Will comment when I put up a fixed commit for final review.

geeg:
I have merged all commits to a single one for the pull request, and have integrated Kun's and Florian's comments.  Due to the suggestion to NOT create tables in the MADlib schema, I tried a workaround that created temporary tables (with hopefully unique enough names) and used "drop if exists" (no cascade).  If there is a better standard, then please discuss with me.  Otherwise, the rest of the code should be final.  Thanks for the help.

src/ports/postgres/modules/crf/crf_feature_gen.sql_in
```diff
@@ -100,120 +100,132 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.crf_train_fgen(
...
  
-    plpy.execute("""INSERT INTO tmp_featureset(f_name, feature) 
+    plpy.execute("""INSERT INTO """ + tmp_featureset + """(f_name, feature) 
                     SELECT DISTINCT f_name, feature
-                    FROM   tmp1_feature;""")
+                    FROM   """ + tmp1_feature + """;""")
 
     plpy.execute("""DROP SEQUENCE IF EXISTS seq;
-                    CREATE TEMP SEQUENCE seq START 1 INCREMENT 1;""")
+                    CREATE  SEQUENCE seq START 1 INCREMENT 1;""")
```
fschopp, line 185: Wouldn't a column of type SERIAL be easier than creating a sequence?

src/ports/postgres/modules/crf/crf_feature_gen.sql_in
```diff
@@ -100,120 +100,132 @@ CREATE OR REPLACE FUNCTION MADLIB_SCHEMA.crf_train_fgen(
...
+    plpy.execute("""CREATE TABLE """ + tmp_featureset + """(f_name text, feature integer[]);""")
+    plpy.execute("""CREATE TABLE """ + tmp_segmenttbl + """(start_pos int,doc_id int,seg_text text,label int,max_pos int);""")
+
+    plpy.execute("""SET client_min_messages TO """ + str(origClientMinMessages[0]['setting']) + """;""")
+
+    # replace digits with "DIGIT" keyword
+    plpy.execute("""INSERT INTO """ + tmp_segmenttbl + """ SELECT start_pos,doc_id,seg_text,label,max_pos FROM """ + segmenttbl + """ WHERE 
+			NOT (seg_text ~ E'^[-+]?([0-9]{1,3}[,]?)*[0-9]{1,3}$' OR seg_text ~ E'^[-+]?[0-9]*[.][0-9]+$');""")
+    plpy.execute("""INSERT INTO """ + tmp_segmenttbl + """ SELECT start_pos,doc_id,'DIGIT',label,max_pos FROM """ + segmenttbl + """ WHERE  
+		    seg_text ~ E'^[-+]?([0-9]{1,3}[,]?)*[0-9]{1,3}$' OR seg_text ~E'^[-+]?[0-9]*[.][0-9]+$';""")
```
fschopp, line 135: This is fine for now, but in general I find this code messy. There are lots of plpy.execute invocations where there only needs to be one. Also it would be preferable to use the Python format function instead of string concatenation.
Eventually, of course, this code should be adapted to the forthcoming Python abstraction layer and all the PG/GP-specific parts should be factored out.

cwelton:
Reassigning to Hai for review.

src/ports/postgres/modules/crf/crf_feature_gen.sql_in
geeg, line 185: The featureset table is already created before crf_feature_gen is called.  It is either created by the user, or by the crf_data_loader code.  Either way, the f_index column of featureset is already of type integer before this line of code, and the only way that I know of to convert an integer column to a serial column is essentially creating this sequence.  

However, we could convert other parts of the CRF code and documentation so that f_index is initially created as a serial column, if you think that is necessary.

fschopp, line 185: No, I don't think it's really important at the moment. But if the featureset table is already created by the user, does it have to be empty? If not, there could be collisions.

geeg, line 185: Yes, it is assumed that the featureset table is empty, though this doesn't seem to be addressed in the CRF documentation.  Perhaps updating the documentation reflecting this assumption would be sufficient?  Otherwise, we could have the CRF module create the featureset table itself and let the user know where it is stored (I think Kun let the user specify it in the first place for simplicity, since it contains useful information for the user to look at and can be passed into the CRF test function easily).

iyerr3:
@walkingsparrow I know you have a message on JIRA but could you please give me a final confirmation here before I merge this?

walkingsparrow:
I have done the test and confirm that this problem is resolved by this commit.

walkingsparrow:
It is merged.

PR #248, opened 2012-10-11
 SVM: Fix the SVM classification issue in parallel mode

renyi533:
Jira: Madlib-690
In parallel mode, classification/regression will fail if there is no data distributed to certain segments.
The root cause is that we always request a model for each segment. With this fix, we firstly check the data
distribution and omit the segments with no data.


iyerr3:
Please update the commit messages as described in the contribution guide: https://github.com/madlib/madlib/wiki/Contribution-Guide . It would be helpful to have a description of the fix in the commit message.

renyi533:
Thanks for your comments. I updated the commit message.

iyerr3:
Everything looks good. I will merge this today.

iyerr3:
Merged with commit: https://github.com/madlib/madlib/commit/a20fbcdbe2299f3962a7c713e5bae4fb00fc2d77

PR #249, opened 2012-10-11
DT Test: Removed use of madlib schema for test

iyerr3:
Jira: MADLIB-684

There is no need to use a temporary table for
test results since a temporary schema is created
before the test starts. Further, any DB object
should be created w/o schema prefix, since the
test is executed in the temporary schema context.
Drop statements should also be avoided since the
default schema will be cleaned outside.

Other changes:
- Removed all trailing whitespace characters. 
  Trailing whitespace should be removed since it 
  adds unnecessary diff lines in the commit.

### DIFF
--- a/methods/cart/src/pg_gp/sql/dt_test.sql_in
+++ b/methods/cart/src/pg_gp/sql/dt_test.sql_in
@@ -1,7 +1,6 @@
 m4_include(`SQLCommon.m4')
 
-DROP TABLE IF EXISTS MADLIB_SCHEMA.golf_dt_test;
-CREATE TABLE MADLIB_SCHEMA.golf_dt_test (
+CREATE TABLE golf_dt_test (
     id          INT,
     outlook     TEXT,
     temperature DOUBLE PRECISION,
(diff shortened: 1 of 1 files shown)


cwelton:
Looks good to me.  Should be safe for merging.

fschopp:
There is no need to create temporary tables in the unit tests, as unit tests are run with a temporary schema first in the `search_path`.

However, creating tables in the MADlib schema is always a bad idea. The DT unit tests still do that.

fschopp:
While we are at it, the DT unit tests should probably be fixed in the sense of my previous comment.

iyerr3:
@fschopp Please check if this is per the rules. The three commits will be melded before merge, but I wanted to run this by you quickly. There are still create functions statements that use madlib schema, as well as some tables for checking for the results. Do you suggest I remove those as well?

fschopp:
I'd say it's fine for now, though at some point the testing should be moved out of the pgsql functions and performed in plain SQL.

iyerr3:
Will appreciate someone testing this with PG 9.2 and GP to ensure nothing got broken. ( @cwelton Hoping you could assign this to someone with open cycles)

iyerr3:
@cwelton Should I go ahead with the merge on this?

cwelton:
Good to merge.

iyerr3:
Merged with commit: https://github.com/madlib/madlib/commit/6f61346486dcf3f05e9442be470362cb7e30089d

PR #252, opened 2012-10-16
BugFix:MADLIB-702

KevinYL:
Eliminate warnings on gcc 4.6.

There are two kinds of warnings.

First, type conversion, from uint64_t to double, between signed and
unsigned types. For type conversion, we add cast explicitly.

Second, unused auto variables. We just remove them.

### DIFF
--- a/src/modules/linalg/average.cpp
+++ b/src/modules/linalg/average.cpp
@@ -48,7 +48,7 @@ public:
 
         mStorage = inAllocator.allocateArray<double, dbal::AggregateContext,
             dbal::DoZero, dbal::ThrowBadAlloc>(
-                arraySize(inNumDimensions));
+                static_cast<uint32_t>(arraySize(inNumDimensions)));
         rebind(inNumDimensions);
         numDimensions = inNumDimensions;
     }
@@ -107,7 +107,7 @@ avg_vector_transition::run(AnyType& args) {
     MappedColumnVector x = args[1].getAs<MappedColumnVector>();
 
     if (state.numRows == 0)
-        state.initialize(*this, x.size());
+        state.initialize(*this, static_cast<uint16_t>(x.size()));
     else if (x.size() != state.sumOfVectors.size()
         || state.numDimensions !=
             static_cast<uint32_t>(state.sumOfVectors.size()))
@@ -141,7 +141,7 @@ avg_vector_final::run(AnyType& args) {
 
     MutableNativeColumnVector avgVector(allocateArray<double>(
         state.sumOfVectors.size()));
-    avgVector = state.sumOfVectors / state.numRows;
+    avgVector = state.sumOfVectors / static_cast<int32_t>(state.numRows);
     return avgVector;
 }
 
@@ -152,7 +152,7 @@ normalized_avg_vector_transition::run(AnyType& args) {
     MappedColumnVector x = args[1].getAs<MappedColumnVector>();
 
     if (state.numRows == 0)
-        state.initialize(*this, x.size());
+        state.initialize(*this, static_cast<uint16_t>(x.size()));
     else if (x.size() != state.sumOfVectors.size()
         || state.numDimensions !=
             static_cast<uint32_t>(state.sumOfVectors.size()))
@@ -170,7 +170,7 @@ normalized_avg_vector_final::run(AnyType& args) {
 
     MutableNativeColumnVector avgVector(allocateArray<double>(
         state.sumOfVectors.size()));
-    avgVector = (state.sumOfVectors / state.numRows).normalized();
+    avgVector = (state.sumOfVectors / static_cast<int32_t>(state.numRows)).normalized();
     return avgVector;
 }
 
--- a/src/modules/linalg/matrix_agg.cpp
+++ b/src/modules/linalg/matrix_agg.cpp
@@ -64,15 +64,16 @@ public:
             MatrixAggState oldSelf = *this;
             mStorage = inAllocator.allocateArray<double, dbal::AggregateContext,
                 dbal::DoZero, dbal::ThrowBadAlloc>(
-                    arraySize(numRows, numColsReserved));
+                    static_cast<uint32_t>(arraySize(numRows, numColsReserved)));
             rebind(numRows, numColsReserved);
             numRows = oldSelf.numRows;
             numCols = oldSelf.numCols;
-            matrix.leftCols(numCols) = oldSelf.matrix.leftCols(numCols);
+            matrix.leftCols(static_cast<int>(numCols)) = 
+                oldSelf.matrix.leftCols(static_cast<int>(numCols));
         }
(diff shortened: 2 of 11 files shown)


src/modules/linalg/average.cpp
```diff
@@ -141,7 +141,7 @@ class AvgVectorState {
 
     MutableNativeColumnVector avgVector(allocateArray<double>(
         state.sumOfVectors.size()));
-    avgVector = state.sumOfVectors / state.numRows;
+    avgVector = state.sumOfVectors / static_cast<int32_t>(state.numRows);
```
fschopp, line 144: int32_t seems small here. I'll check.

src/modules/linalg/average.cpp
```diff
@@ -170,7 +170,7 @@ class AvgVectorState {
 
     MutableNativeColumnVector avgVector(allocateArray<double>(
         state.sumOfVectors.size()));
-    avgVector = (state.sumOfVectors / state.numRows).normalized();
+    avgVector = (state.sumOfVectors / static_cast<int32_t>(state.numRows)).normalized();
```
fschopp, line 173: See above.

src/modules/linalg/matrix_agg.cpp
```diff
@@ -64,15 +64,16 @@ class MatrixAggState {
             MatrixAggState oldSelf = *this;
             mStorage = inAllocator.allocateArray<double, dbal::AggregateContext,
                 dbal::DoZero, dbal::ThrowBadAlloc>(
-                    arraySize(numRows, numColsReserved));
+                    static_cast<uint32_t>(arraySize(numRows, numColsReserved)));
```
fschopp, line 67: size_t?

src/modules/linalg/matrix_agg.cpp
```diff
@@ -64,15 +64,16 @@ class MatrixAggState {
...
             numRows = oldSelf.numRows;
             numCols = oldSelf.numCols;
-            matrix.leftCols(numCols) = oldSelf.matrix.leftCols(numCols);
+            matrix.leftCols(static_cast<int>(numCols)) = 
+                oldSelf.matrix.leftCols(static_cast<int>(numCols));
         }
 
         rebind(numRows, numCols + 1);
-        return matrix.col(numCols++);
+        return matrix.col(static_cast<int>(numCols++));
```
fschopp, line 76: We don't use int.

fschopp:
I'm going to merge.

src/modules/regress/multilogistic.cpp
```diff
@@ -219,9 +219,9 @@ class MLogRegrIRLSTransitionState {
     MappedColumnVector x = args[3].getAs<MappedColumnVector>();
 
     // Get the category & numCategories as integer
-    int16_t category = args[1].getAs<int>();
+    int32_t category = args[1].getAs<int>();
```
fschopp, line 222: also need to change the int on the right-hand side.

src/modules/linalg/average.cpp
fschopp, line 144: The warning was

```
/Users/schopf/Documents/Development/fschopp-madlib/src/modules/linalg/average.cpp:173:45: warning: conversion to 'Eigen::MatrixBase<Eigen::Map<const Eigen::Matrix<double, -0x00000000000000001, 1>, 0, Eigen::Stride<0, 0> > >::Scalar {aka double}' from 'long long unsigned int' may alter its value [-Wconversion]
```

Converting to int32_t is certainly not going to make things better. The correct way is of course still to convert to double. The warning is because double only has 52 bits of precision, while uint64_t has 64 bit of precision for integers.

src/modules/linalg/average.cpp
```diff
@@ -48,7 +48,7 @@ class AvgVectorState {
 
         mStorage = inAllocator.allocateArray<double, dbal::AggregateContext,
             dbal::DoZero, dbal::ThrowBadAlloc>(
-                arraySize(inNumDimensions));
+                static_cast<uint32_t>(arraySize(inNumDimensions)));
```
fschopp, line 51: No, this should be size_t.

src/modules/linalg/matrix_agg.cpp
```diff
@@ -64,15 +64,16 @@ class MatrixAggState {
...
             mStorage = inAllocator.allocateArray<double, dbal::AggregateContext,
                 dbal::DoZero, dbal::ThrowBadAlloc>(
-                    arraySize(numRows, numColsReserved));
+                    static_cast<uint32_t>(arraySize(numRows, numColsReserved)));
             rebind(numRows, numColsReserved);
             numRows = oldSelf.numRows;
             numCols = oldSelf.numCols;
-            matrix.leftCols(numCols) = oldSelf.matrix.leftCols(numCols);
+            matrix.leftCols(static_cast<int>(numCols)) = 
+                oldSelf.matrix.leftCols(static_cast<int>(numCols));
```
fschopp, line 72: Eigen::Index

src/modules/linalg/matrix_agg.cpp
```diff
@@ -95,7 +96,8 @@ class MatrixAggState {
     void rebind(uint64_t inNumRows, uint64_t inNumCols) {
         numRows.rebind(&mStorage[0]);
         numCols.rebind(&mStorage[1]);
-        matrix.rebind(&mStorage[2], inNumRows, inNumCols);
+        matrix.rebind(&mStorage[2], static_cast<int>(inNumRows), 
+            static_cast<int>(inNumCols));
```
fschopp, line 100: Don't use int.

src/modules/regress/multilogistic.cpp
```diff
@@ -219,9 +219,9 @@ class MLogRegrIRLSTransitionState {
...
     // Number of categories after pivoting (We pivot around the first category)
-    int16_t numCategories = (args[2].getAs<int>() - 1);
+    int32_t numCategories = (args[2].getAs<int>() - 1);
```
fschopp, line 224: getAs<int>() is a very bad idea and not portable.

src/modules/stats/one_way_anova.cpp
```diff
@@ -270,7 +270,7 @@ class OWATransitionState {
     double grand_mean = state.sum.sum() / state.num.sum();
     double sum_squares_between = 0;
 
-    for (uint64_t idx = 0; idx < state.numGroups; idx++)
+    for (int32_t idx = 0; idx < static_cast<int32_t>(state.numGroups); idx++)
```
fschopp, line 273: state.numGroups is of type uint32_t, so casting to int32_t is not a great idea.

fschopp:
Please write commit messages in the format we have agreed on. Thanks!

fschopp:
Merged with commit d9dc565. Additional warnings fixed with commit 26dc73b.

PR #254, opened 2012-10-19
Fixed null values in the results of naive Bayes

farzadras:
Jira: MADLIB-523

This fix addresses the issue of NULL values which stemmed from attribute values in testing data set that cannot be found in training data set.

### DIFF
--- a/src/ports/postgres/modules/bayes/bayes.py_in
+++ b/src/ports/postgres/modules/bayes/bayes.py_in
@@ -207,11 +207,6 @@ def __get_keys_and_prob_values_sql(**kwargs):
     For class c and the attribute array identified by key k, log_prob is
     log( P(C = c) * P(A = a(k)[] | C = c) ).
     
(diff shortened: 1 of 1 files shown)


iyerr3:
Code review completed with Farzad - the code and logic looks fine. Pending install-check tests.

iyerr3:
@farzadras Please update the initial comment in changed file to reflect that NULL values are no more created for each (key, class) pair (as discussed in our review). The code could also use short inline comments on the logic for each nested subquery.

iyerr3:
@farzadras Please place a comment here after you have updated this with your latest changes (including change in comments).

farzadras:
Sure. There is a new data set that I am looking into before finalizing this pull request.

farzadras:
Updates: Tested with the new data set. Removed the comment about NULL values in code.

iyerr3:
Code merged: https://github.com/madlib/madlib/commit/d7cd128595f980c1ef283c740947d8c8638e9876

PR #255, opened 2012-10-20
CRF: Added ordered aggregates check for install-check

geeg:
Jira: MADLIB-725
Description: Add ordered aggregates check to crf_train_large
so that it is skipped if the db does not support (e.g. PG8.4)


iyerr3:
@walkingsparrow Hai, since you have the 3 environments already setup (per our previous test of CRF), could you please verify this PR and merge if the unit tests for GPDB and PG 8.4, 9.1 pass?

walkingsparrow:
The new CRF module passed all the install-checks on the newest Greenplum and PostgreSQL 8.4, 9.1, 9.2. 

I have merged it with master.

PR #256, opened 2012-10-24
K-means: Fix silhouette function for one centroid

geeg:
Jira: MADLIB-681
Fix silhouette function so that it returns a valid
value when there is only one centroid.

### DIFF
--- a/src/ports/postgres/modules/kmeans_new/kmeans_new.sql_in
+++ b/src/ports/postgres/modules/kmeans_new/kmeans_new.sql_in
@@ -1089,6 +1089,7 @@ AS $$
 DECLARE
     class_rel_source REGCLASS;
     proc_fn_dist REGPROCEDURE;
+    num_closest INT2;
 BEGIN
     class_rel_source := rel_source;
     proc_fn_dist := fn_dist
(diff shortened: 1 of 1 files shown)


cwelton:
For the simplified silhouette the basic calculation is (distance to own centroid) / (distance to next closest centroid).  If there is only once centroid how are you trying to calculate this ratio?

walkingsparrow:
For only one centroid, this quantity is not defined. In R, this calculation
for one centroid simply gives 'NA'. Do we really need to assign some value
to this situation?

Hai

On Wed, Oct 24, 2012 at 9:41 AM, Caleb Welton notifications@github.comwrote:

> For the simplified silhouette the basic calculation is / . If there is
> only once centroid how are you trying to calculate this ratio?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/madlib/madlib/pull/256#issuecomment-9746513.

## 

from GMail

iyerr3:
Hai has a valid point. Currently, it returns '1', indicating the cohesion within the single cluster is perfect. That may not be completely true.

iyerr3:
Would it be better to move silhouette coefficient calculation outside of kmeans? Silhouette is used for clustering, not specific to kmeans.

sujitp:
I like Rahul's suggestion of keeping it separate.

geeg:
I agree with Hai and Rahul; for one centroid, it would be more correct to return "NaN" as the code originally did.  However, there still needs to be some testing of the code to ensure that "NaN" is being returned for the correct reason (and not because some other part of the code is buggy).

walkingsparrow:
The original code is correct. The distances are set to an initial value of infinity in the C++ code. When there is only one centroid, distances[1] has some finite value and distances[2] keeps the initial infinite value, the result becomes NAN.

No need to change.

PR #259, opened 2012-10-24
Quantile: Suppress assign float to integer

KevinYL:
Jira: MADLIB-601

quantile_big used BIGINT to store the number of rows. The code need to calculate quantile_size by multiplying the total number of rows with the quantile, whose result type is a FLOAT8. The code tries to assign this result to a BIGINT directly, and this operation causes problem.

To fix this, we make a coerce on the result.


iyerr3:
Merged with commit: https://github.com/madlib/madlib/commit/5d05088d64ea87fbbf352395062d75e174d4ae6a

PR #260, opened 2012-10-24
Kmeans_new: make kmeans++ method support SVEC

renyi533:
Jira: madlib-730
In current code, we cannot correctly handle SVEC input. In this fix, I explicitly convert SVEC to array whenever necessary to fix the issues.


walkingsparrow:
Merged with master.

PR #261, opened 2012-10-25
kmeans_new: handle infinite numbers

renyi533:
Jira: madlib_729
Filter out infinite numbers from kmeans' input.


iyerr3:
@renyi533 Yi, please squash the two commits together before we go ahead with the merge. Thanks.

walkingsparrow:
The two commits are squashed, and merged with master.

PR #262, opened 2012-10-25
sketch bugfix: deal with performance degeneration that leads by strictly contents checking

KevinYL:
JIRA: MADLIB-689

In order to fix the security issues of the sketch method described in
MADLIB-630, at the entry point of each UDF, we did strictly checking of
the contents stored in a vararea to make sure that each access to
elements in the vararea is safe. This checking degenerates the
performance.

Because these step functions only access each item at most one time on
each call, in order to avoid performance degeneration, we only do
checking when access the specific element. The performance degeneration
decreases from 50% at most to 20%.


iyerr3:
@farzadras Please update the commit message header before merge

iyerr3:
Merged: https://github.com/madlib/madlib/commit/4ee6c4446c30112c3a094eb8f569131584c97890

PR #263, opened 2012-10-26
K-means: Fix parameter k to be of type integer

geeg:
instead of small int

JIRA: MADLIB-697
Fixed kmeans seeding functions and closest_columns to allow number
of centroids k to be of type integer.  Also added in parameter
checks to ensure k is within a valid range.


walkingsparrow:
Merged with commit 00facf86b8.

PR #265, opened 2012-10-27
K-means: Add additional non-finite value checks

geeg:
JIRA: MADLIB-403
Add non-finite value check for kmeans cset
and add error message when initial centroids have
non-finite values


walkingsparrow:
Merged with master

PR #268, opened 2012-10-30
Multilogistic: Add input validation for dependent variable

iyerr3:
Jira: MADLIB-733

Dependent variable for multinomial logistic regression
should be an integer value since we are performing
classification, hence the variable will be of the form
1 of K values. This change displays an appropriate
error message if the parameter passed is not integer.

### DIFF
--- a/src/ports/postgres/modules/regress/multilogistic.sql_in
+++ b/src/ports/postgres/modules/regress/multilogistic.sql_in
@@ -134,7 +134,7 @@ Note that the categories are encoded as integers with values from {0, 1, 2,...nu
 -# Create the sample data set:
 @verbatim
 sql> SELECT * FROM data;
-                  r1                      | val 
+                  r1                      | val
 ---------------------------------------------+-----
  {1,3.01789340097457,0.454183579888195}   | 1
  {1,-2.59380532894284,0.602678326424211}  | 0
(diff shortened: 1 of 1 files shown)


src/ports/postgres/modules/regress/multilogistic.sql_in
```diff
@@ -348,6 +348,16 @@ DECLARE
     fnName VARCHAR;
     theResult MADLIB_SCHEMA.mlogregr_result;
 BEGIN
+    IF (SELECT atttypid::regtype != 'INTEGER'::regtype
```
haradh1: It works, but for SQLer '<>' is the standard non-equal operator.

src/ports/postgres/modules/regress/multilogistic.sql_in
```diff
@@ -348,6 +348,16 @@ DECLARE
...
+        FROM pg_attribute
+        WHERE attrelid = $1::regclass AND attname = $2) THEN
```
haradh1: I wonder why we're using $1 and $2 instead of variable name while argument variable optimizer is used below.

src/ports/postgres/modules/regress/multilogistic.sql_in
iyerr3: I was following the convention from this file and kmeans. Will change if <> is preferred.

iyerr3:
Changes made as suggested. @haradh1, thanks for your comments!

walkingsparrow:
Merged with master

PR #269, opened 2012-10-30
Naive Bayes: Fix install-check errors

farzadras:
Jira: MADLIB-738

Install-check produced errors because of type cast problems in naive Bayes code that was taken care of in this fix.

### DIFF
--- a/src/ports/postgres/modules/bayes/bayes.py_in
+++ b/src/ports/postgres/modules/bayes/bayes.py_in
@@ -244,7 +244,7 @@ def __get_keys_and_prob_values_sql(**kwargs):
 				classify.key,
 				featureprobs.class,
 				classify.attr, max(classify.value) as cvalue,
-				max(classify.value) || array_agg(featureprobs.value) as fpvalue,
+				max(classify.value)::DOUBLE PRECISION || array_agg(featureprobs.value) as fpvalue,
 				0::bigint || array_agg(featureprobs.cnt) as fp_cnt,
 				max(featureprobs.attr_cnt) as fp_attr_cnt,
 				classpriors.class_cnt,
--- a/src/config/Modules.yml
+++ b/src/config/Modules.yml
@@ -6,6 +6,7 @@ modules:
     - name: assoc_rules
       depends: ['svec']
     - name: bayes
+      depends: ['array_ops']
     - name: compatibility
       depends: ['utilities']
     - name: conjugate_gradient


src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -244,7 +244,7 @@ def __get_keys_and_prob_values_sql(**kwargs):
 				classify.key,
 				featureprobs.class,
 				classify.attr, max(classify.value) as cvalue,
-				max(classify.value) || array_agg(featureprobs.value) as fpvalue,
+				max(classify.value)::DOUBLE PRECISION || array_agg(featureprobs.value) as fpvalue,
```
haradh1, line 247: It is better to use COALESCE() to avoid unexpected NULL element in the array, unless you're sure classify.value never be NULL.

Also, IIRC array_agg is not supported in GPDB 4.0.  You may need to wrap it with m4 macro, or you have to disable bayes module entirely in GPDB 4.0.  If latter, take a look at src/ports/greenplum/4.0/config/Modules.yml.  In any case, you'd better to run install-check in GPDB 4.0.

iyerr3:
install-check passed on GPDB main, PG 9.1, and PG 9.2. Checking with GPDB 4.0 per Hitoshi's suggestion

farzadras:
@haradh1 'classify.value' can't be NULL. So, we should be good in this case.

iyerr3:
Since array_agg is not available on 4.0, and to ensure an immediate solution for the install-check issue, I would like to merge this, while disabling Bayes on GPDB 4.0. There are other modules that have been disabled on 4.0 and on discussion with Hitoshi, the consensus is that 4.0 is not commonly used and we can disable Bayes till an alternative fix is implemented.

fschopp:
Quick note: The `array_ops` module has its own `array_agg`.

cwelton:
Given that we have the array_ops version of array_agg it should be trivial to get this to work in GPDB 4.0.

iyerr3:
I will wait for @farzadras to update before merging.

farzadras:
@iyerr3 I think this is ready to merge. I am testing with GPDP 4.0 simultaneously.

iyerr3:
Discussed with @farzadras, but verifying with @haradh1, @cwelton : the `array_agg` would need to be wrapped with `m4_ifdef`, isn't it?

farzadras:
Sure. My database is not responding right now. I have to fix that before resuming my investigations.

iyerr3:
Due to time constraints, compatibility concerns have been taken care in new [PR 275](https://github.com/madlib/madlib/pull/275). Closing this PR.

PR #275, opened 2012-10-31
Naive Bayes: Rectify type cast errors

iyerr3:
Jira: MADLIB-738

Commit d7cd128 introduced type casts errors triggering a failure in
install-checks. This fix takes care of the type cast error.
Also, GPDPB 4.0 does not support array_agg. Hence we use m4 macros to
detect the platform and use the array_agg defined in array_ops
if necessary.

Other changes:
    Removed trailing whitespace


walkingsparrow:
Merged with commit 238d4dddf0

PR #279, opened 2012-11-05
Revert to old Naive Bayes (from commit f32a706)

geeg:
JIRA: MADLIB-745, MADLIB-746
Revert to old Naive Bayes to avoid out-of-memory error and performance regression.  Still need to check for correctness (MADLIB-679).


iyerr3:
Taken care of in https://github.com/madlib/madlib/commit/be87d2f6436c1f26a55b9906393f2c179f57c4ed

PR #281, opened 2012-11-05
Naive Bayes: Explanations added to documentation for several possible

walkingsparrow:
problems

Jira MADlib-679

(1) The probabilities computed on the platforms of PostgreSQL and Greenplum
database have very tiny difference due to the nature of floating point
computation. Usually this is not important. However, if a data point has
equal (or very close) probabilities for two classes, this data point might
be classified into diferent classes on PostgreSQL and Greenplum. This leads
to the classifications with a small difference on PostgreSQL and Greenplum
for some data sets, but does not affect the quality of the results.

(2) The current implementation of Naive Bayes classification is only suitable
for discontinuous attributes.

(3) One can still feed continuous data into the naive Bayes classification
function. No warning is raised, but the result is not correct.


iyerr3:
Merged: https://github.com/madlib/madlib/commit/87817903f7e8560d6646c09599c3bf13983a7c71

PR #282, opened 2012-11-06
NB Documentation: Add more explanation for classification

walkingsparrow:
Jira: MADlib-679

When the probabilities for the two classes are exactly equal, the result is
an array of these two classes, but the order of the two classes is random.

One can still provide continuous (floating point) data to the naive Bayes
classification function. Floating point numbers can be used as categorial data.


iyerr3:
Code merged: https://github.com/madlib/madlib/commit/159622d

PR #283, opened 2012-11-09
Naive Bayes: Fix the missing value in the result

walkingsparrow:
Jira MADlib-523

When the test set has values that are in the training set, we use the
Laplace smoothing to avoid a NULL output.

Extra changes:
    The computation code is re-used in create_classification_function;
    Documentation is changed to reflect the fix;
    Useless notices and warnings during computation are removed

### DIFF
--- a/src/ports/postgres/modules/bayes/bayes.py_in
+++ b/src/ports/postgres/modules/bayes/bayes.py_in
@@ -107,7 +107,7 @@ def __get_feature_probs_sql(**kwargs):
         """.format(**kwargs)
 
 
-def __get_attr_values_sql(**kwargs):
+def __get_attr_values_sql(**kwargs): # need to filter out NULL values
     """
     Return SQL query with columns (attr, value).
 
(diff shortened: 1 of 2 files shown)


src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -232,17 +232,42 @@ def __get_keys_and_prob_values_sql(**kwargs):
     # {classifySource} cannot be a subquery, because we use it more than once in
     # our generated SQL.
     return """
-    SELECT t2.key, t2.class, t1.log_prob FROM
+    SELECT t1.key, t1.class,
+    CASE WHEN {smoothingFactor} = 0 THEN
+      t1.log_prob
+    ELSE
+      -- log_prob is separated into two parts:
```
iyerr3, line 239: Hai, I really like the comments explaining the logic. It would, however, be great to have the user/dev doc updated with the equations that you showed me (instead of here in the code). That way we don't need to add a big explanation about the logic here.

src/ports/postgres/modules/bayes/bayes.py_in
```diff
@@ -232,17 +232,42 @@ def __get_keys_and_prob_values_sql(**kwargs):
...
+      -- The following part of log_prob is for denominators, whose number should always be {numAttrs}
+      -- A strange thing: When all of this query is put into a function definition, all branches of "case" will be kind of evaluated
+      -- and if {smoothingFactor} is 0 at this point, we will have "logrithm of zero" error. Since we are going to re-use
+      -- this code in __get_prob_values_sql, where everything will be put into a function definition, we have to use
+      -- a small trick to avoid this error: use "case when $2=0 then 1 else $2 end" inside log(), which is "sf" in the following.
+      -- Radihka is looking into this and trying to find a way to avoid the extra computation inside a function.
```
iyerr3: Maybe we shouldn't add names with future work here. Could possibly add in the commit comment.

cwelton:
This fix may apply to the current version of Naive Bayes, but there are other known issues with the method.

Plan at the moment is to redesign the NB method at which point this PR won't apply directly.  To be evaluated for priority assessment.

PR #284, opened 2012-11-12
resolve crf inference bottleneck 

livingstream:
This pull request resolve crf inference bottleneck. 
I changed the schema of viterbi_rtbl from  to  "(seg_text text, score integer[])"  to  "(seg_text text, label integer, score integer)".
This change will make the construction of the  'rfactors' table in the 'vcrf_label' function 44(tag set size -1) times faster ideally.

### DIFF
--- a/src/ports/postgres/modules/crf/crf_feature_gen.sql_in
+++ b/src/ports/postgres/modules/crf/crf_feature_gen.sql_in
@@ -244,12 +244,12 @@ $$
     	segment_hashtbl = "pg_temp._madlib_segment_hashtbl"
     	unknown_segment_hashtbl = "pg_temp._madlib_unknown_segment_hashtbl"
     	rtbl = "pg_temp._madlib_rtbl"
-        rtbls = "pg_temp._madlib_rtbls"
+        rtbl_sum = "pg_temp._madlib_rtbl_sum"
     	mtbl = "pg_temp._madlib_mtbl"
     	tmp_segmenttbl = "pg_temp._madlib_tmp_segmenttbl"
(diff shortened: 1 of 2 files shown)


geeg:
This code checks out on GPDB 4.2 and PG 9.1, but I ran crf_test_fgen and vcrf_label on a segment table with 115781 rows, 11078 distinct "words", and the running time of this code and the old code was generally the same.  I can do more in-depth analysis in the following week.

livingstream:
Grace, your dataset is still not small. I  once used  datasets with 100k sentences(100k_30 rows) and 1 million sentences(1million_30 rows) and found the previous code has the performance bottleneck I mentioned.

geeg:
Hi Kun, do you have these larger data sets available for us to use for performance tests?  The data set I have been using is from here: http://www.cnts.ua.ac.be/conll2000/chunking/, and using all training and test sentences (10948 sentences -> 259104 rows)  to test crf_test_fgen and vcrf_label, I haven't seen much difference in performance between this pull request and the current implementation in the master branch.  We are also working on building more thorough CRF tests for QA testing, so having these large data sets to work with would be helpful.

livingstream:
Hi Grace, the file is 260 MBs after compressed. It is too big to send by email. What is the best way to send it to you?
Dropbox? What is your Dropbox ID?

geeg:
This code has been verified and is ready for merge.

cwelton:
This may be suitable for merge based on comments in the PR.  

Currently there is no automated testing for this method so proper evaluation of completeness cannot be verified.  

As this is still a module marked as early stage development it could be a candidate for completing the PR and integrating into the product.

PR #285, opened 2012-11-20
Build System: Change malformed PGXN version string.

haradh1:
Jira: MADLIB-761

It turned out PGXN version string should be x.y.z{modifier} and we missed
z.  This might be applied to v0.5 branch, but we keep it only in master
for now.


iyerr3:
Merged: https://github.com/madlib/madlib/commit/755290847d6844fa4c5601a0b188bd15da2055b9

PR #286, opened 2012-11-28
CRF: allow schema names to be used for Viterbi tables

geeg:
JIRA: MADLIB-765
Fix crf_test_fgen, vcrf_label, and vcrf_top1_table (was: vcrf_top1_view)
 to allow schema names in table name parameters. Output of vcrf_label is
 now a table instead of a view (see comments in vcrf_top1_table).


cwelton:
Should be something we can resolve with the other outstanding CRF PR.

haying:
Close as Sunny expects this is fixed with the commit [4165b4009a](https://github.com/madlib/madlib/commit/4165b4009aa11d2a6765ff89e498e3c66d23048c).

PR #287, opened 2012-12-08
kmeans: Fixed the dist_angle metric to return values when norm is 0

walkingsparrow:
Jira MADlib-768

When distance metric 'madlib.dist_angle' is used, which is defined
as the angle between two vectors, Inf is obtained when either of the two
norms is 0. The angle is not defined mathematically. Just let the metric
return \pi under such circumstances to avoid error. This should not
affect the qulity of the result.


iyerr3:
Merged to integration with https://github.com/madlib/madlib/commit/3d85c1c015b890633171ee5f70d1deb5116941f9

PR #288, opened 2012-12-17
cart: allow BIGINT for rf_train

renyi533:
Jira: madlib-764

Currently, rf_train won't work if the ID field is a BIGINT. I modified
the function of __sample_with_replacement to correctly handle BIGINT
numbers.


cwelton:
Looks super straight forward and should be easily pulled.

mwellons414:
This pull-request was integrated into MADlib.

PR #289, opened 2012-12-18
Compressed bitmap implementation.

aojwang:
JIRA: MADLIB-767
I implement the bitmap as a user defined type. The underlying storage is int4 array. The following functions are provided in this version:
1) aggregate: bitmap_agg
2) operators: |, &, ~ and #.
3) casts: int4[]/int8[] to bitmap, varbit to bitmap, bitmap to int4[] and bitmap to varbit.
4) functions: bitmap_set, bitmap_test, bitmap_nonzero_positions, bitmap_nonzero_count, and bitmap_unnest
5) comparator for btree operator class: >, <, >=, <=, =, <>.

### DIFF
--- a/src/modules/assoc_rules/assoc_rules.cpp
+++ b/src/modules/assoc_rules/assoc_rules.cpp
@@ -47,7 +47,7 @@ typedef struct perm_fctx
 void *
 gen_rules_from_cfp::SRF_init(AnyType &args) {
     perm_fctx     *myfctx        = NULL;
-    char          *positions     = args[0].getAs<char*>();
+    char          *positions     = text_to_cstring(args[0].getAs<text*>());
 
     // allocate memory for user context
     myfctx            = new perm_fctx();
--- a/src/modules/bitmap/BitmapUtil.hpp
+++ b/src/modules/bitmap/BitmapUtil.hpp
@@ -1,0 +1,651 @@
+#ifndef MADLIB_MODULES_BITMAP_BITMAP_UTILITY_HPP
+#define MADLIB_MODULES_BITMAP_BITMAP_UTILITY_HPP
+
+#include <dbconnector/dbconnector.hpp>
+
+#include "Bitmap_proto.hpp"
+#include "Bitmap_impl.hpp"
+
+namespace madlib {
+namespace modules {
+namespace bitmap {
+
+using madlib::dbconnector::postgres::TypeTraits;
+using madlib::dbconnector::postgres::madlib_get_typlenbyvalalign;
+
+/**
+ * @brief This class encapsulates the interfaces for manipulating the bitmap.
+ *        All functions are static.
+ *
+ */
+class BitmapUtil{
+protected:
+    enum BITMAPOP {EQ = 0, GT = 1, LT = -1} ;
+public:
+
+/**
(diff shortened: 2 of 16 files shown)


cwelton:
Needs code review & testing before this can be pulled.

cwelton:
Another issue with this PR is that there are changes to the db abstraction layer that need some careful review.

src/modules/assoc_rules/assoc_rules.cpp
```diff
@@ -47,7 +47,7 @@
 void *
 gen_rules_from_cfp::SRF_init(AnyType &args) {
     perm_fctx     *myfctx        = NULL;
-    char          *positions     = args[0].getAs<char*>();
+    char          *positions     = text_to_cstring(args[0].getAs<text*>());
```
cwelton, line 50: Not related to this change, should be removed from pull request

src/modules/bitmap/BitmapUtil.hpp
```diff
@@ -0,0 +1,651 @@
...
+
+#include "Bitmap_proto.hpp"
+#include "Bitmap_impl.hpp"
+
+namespace madlib {
+namespace modules {
+namespace bitmap {
+
+using madlib::dbconnector::postgres::TypeTraits;
+using madlib::dbconnector::postgres::madlib_get_typlenbyvalalign;
```
cwelton, line 14: `madlib_get_typlenbyvalalign` breaks the dbconnector abstraction, code should be refactored so that this is not necessary.

src/modules/bitmap/BitmapUtil.hpp
```diff
@@ -0,0 +1,651 @@
...
+ *
+ */
+template<typename T>
+static
+const ArrayType*
+bitmap_and
+(
+    AnyType &args
+){
+    Bitmap<T> bm1(GETARG_IMMUTABLE_BITMAP(args[0]));
```
cwelton, line 140: Would be much cleaner if we had getAs() support for Bitmap

src/modules/bitmap/BitmapUtil.hpp
```diff
@@ -0,0 +1,651 @@
...
+template<typename T>
+static
+const ArrayType*
+bitmap_set
+(
+    AnyType &args
+){
+
+    madlib_assert(!args[0].isNull() && !args[1].isNull() && !args[2].isNull(),
+            std::invalid_argument("the input parameters should not be null"));
```
cwelton, line 226: Rather than the assert, why isn't this function simply defined as strict?  To that end, wouldn't returning NULL in this case be more appropriate than raising an error?

src/modules/bitmap/BitmapUtil.hpp
```diff
@@ -0,0 +1,651 @@
...
+ *
+ * @param user_fctx     the user context
+ * @param is_last_call  is it the last call
+ *
+ * @return the current value
+ */
+template <typename T>
+static
+int64_t
+bitmap_unnest_next(void *user_fctx, bool *is_last_call) {
```
cwelton, line 469: Rather than utilizing the current poor dbconnector implementation for set returning functions it would be better to fix the interface.

PR #290, opened 2012-12-20
Improve the performance of Apriori by using compressed bitmap rather tha...

aojwang:
Improve the performance of Apriori by using compressed bitmap rather sparse vector (this PR must be merged after PR-289, which is about compressed bitmap).

JIRA: MADLIB-771.
Replace the svec to compressed bitmap and remove some unnecessary columns from the tables.


cwelton:
Dependent on https://github.com/madlib/madlib/pull/289 and cannot be pulled without it.

PR #291, opened 2013-01-08
Abstraction Layer: Allow UDFs to access the protected member fcinfo and ...

yangs16:
...mSysInfo of class AnyTypet

Jira: MADLIB-775
The UDFs cannot access the protected members fcinfo and mSysInfo of class
AnyType, although the class UDF is declared as friend of class AnyType (Note
that the friendship cannot be inherited in C++). The reasons for a UDF to
access fcinfo include judging whether it is called in an aggregate state or
window state. The reasons for a UDF to access mSysInfo include using
mSysInfo->user_fctx to keep some internal states.

### DIFF
--- a/src/ports/postgres/dbconnector/AnyType_proto.hpp
+++ b/src/ports/postgres/dbconnector/AnyType_proto.hpp
@@ -41,9 +41,17 @@ public:
     bool isComposite() const;
     AnyType& operator<<(const AnyType& inValue);
 
-    inline SystemInformation * getSysInfo(){
-        return this->mSysInfo;
-    }
+    // FIXME: A temporary workaround for UDF to get/set user_fctx, a better
+    // solution is desired which complies with the design principles of DB
+    // abstraction layer.
+    // For some analytic algorithms (e.g. LDA), a stateful function would be
+    // very useful which allows to carry some states accross invocations within
+    // a query. Alternatively, one can achieve similar functionaltiy using a
+    // windowed aggregator, but this may suffer from severe performance
+    // degradation.  
+    void * getUserFuncContext();
+    void setUserFuncContext(void * user_fctx);
+    MemoryContext getCacheMemoryContext();
 protected:
     /**
      * @brief RAII class to temporarily change \c sLazyConversionToDatum
--- a/src/ports/postgres/dbconnector/AnyType_proto.hpp
+++ b/src/ports/postgres/dbconnector/AnyType_proto.hpp
@@ -44,9 +44,6 @@ public:
     inline SystemInformation * getSysInfo(){
         return this->mSysInfo;
     }
-    inline FunctionCallInfo getFCInfo(){
-        return this->fcinfo;
-    }
 protected:
     /**
      * @brief RAII class to temporarily change \c sLazyConversionToDatum
--- a/src/ports/postgres/dbconnector/AnyType_proto.hpp
+++ b/src/ports/postgres/dbconnector/AnyType_proto.hpp
@@ -41,6 +41,12 @@ public:
     bool isComposite() const;
     AnyType& operator<<(const AnyType& inValue);
 
+    inline SystemInformation * getSysInfo(){
+        return this->mSysInfo;
+    }
+    inline FunctionCallInfo getFCInfo(){
+        return this->fcinfo;
+    }
 protected:
     /**
      * @brief RAII class to temporarily change \c sLazyConversionToDatum
--- a/src/ports/postgres/dbconnector/AnyType_impl.hpp
+++ b/src/ports/postgres/dbconnector/AnyType_impl.hpp
@@ -13,6 +13,16 @@ namespace dbconnector {
 
 namespace postgres {
 
+inline void * AnyType::getUserFuncContext(){
+    return this->mSysInfo->user_fctx;
+}
+inline void AnyType::setUserFuncContext(void * user_fctx){
+    this->mSysInfo->user_fctx = user_fctx;
+}
+inline MemoryContext AnyType::getCacheMemoryContext(){
+    return this->mSysInfo->cacheContext;
+}
+
 inline
 AnyType::AnyType(FunctionCallInfo inFnCallInfo)
   : mContentType(FunctionComposite),


yangs16:
A better place to add the access for fcinfo and mSysInfo is in UDF_proto.hpp. We can add two protected member variables to class UDF and set their values in UDF::invoke(AnyType & args) method. However, in the current design, all the member functions of class UDF are static, this will lead to a lot of changes to class UDF. Thus, it seems that the current solution (add two public functions in class AnyType) is still a good choice. Besides,  these methods should only be called in UDF_inherited::run(AnyType & args). In this case, the parameter args is just constructed via fcinfo in UDF::call(FunctionCallInfo fcinfo), which contains the parameters and context information of the function call.

src/ports/postgres/dbconnector/AnyType_proto.hpp
```diff
@@ -41,6 +41,12 @@ class AnyType {
     bool isComposite() const;
     AnyType& operator<<(const AnyType& inValue);
 
+    inline SystemInformation * getSysInfo(){
+        return this->mSysInfo;
+    }
+    inline FunctionCallInfo getFCInfo(){
+        return this->fcinfo;
+    }
```
fschopp: `AnyType` is a generic abstraction class that is supposed to be used completely independent of PostgreSQL, so a function like `getFCInfo` should not become part of the public API.

cwelton: We had a brief conversation about this last night, with roughly the same conclusion.

The overall goal is to be able to set/retrieve fcinfo->user_ctx, for which I feel that:
    a) Anytype is the wrong place to hook into, it should be part of the overall abstraction for a function invocation.
    b) Exposing FCInfo directly breaks the abstraction notion and we should have targeted methods directly for the notion user context.

If you have particular pointers to where in the abstraction layer these changes could be made it may prove helpful.

yangs16:
Florian, you are right. It's not a good idea to add an pubic API to AnyType. For FunctionCallInfo, what I want is to judge the call context. There is a function in Compatibility.hpp as follows:

<pre><code>
int AggCheckCallContext(FunctionCallInfo fcinfo, MemoryContext *aggcontext);
</code></pre>

But to call the function, we still need to have access to fcinfo. My problem is how to get the call context in the run() method of a UDF. I just found that fcinfo is a protected member of class Allocator, which can be accessed in the run() method. Is this the right way?

By the way, it seems that SystemInformation::user_fctx is designed for users to put their own context, but the problem is how it can be accessed in the run() method of a UDF. Can you give some guides on this? What I want is somehow like this:

<pre><code>
args.getSysInfo()->user_fctx =
             MemoryContextAllocZero(
                     args.getSysInfo()->cacheContext,
                     __state_size * sizeof(int32_t));
</code></pre>

Thanks a lot.

yangs16:
Unfortunately, it seems that the member variable fcinfo of class Allocator is never initialized and used.

yangs16:
Based on discussion with Caleb and the team, we decide not to expose fcinfo to UDF.

fschopp:
The question is why and where you want to call AggCheckCallContext().

yangs16:
Hi Florian, I just want to check the call context state (AggregateState, WindowState, or others). If the state function of an aggregator is called not in the AggregatState, it can raise an exception and end early. But it is not a necessity. After discussions with Caleb and the team today, we decide not to expose fcinfo in any way.

yangs16:
By the way, Florian, can you give some comments on my second question? How can I get SystemInformation::user_fctx to put some context data in a UDF? And could you please take a look to http://jira.madlib.net/browse/MADLIB-774 when you have time? It's also related with DB Abstraction Layer.

fschopp:
That sounds as if you want to call AggCheckCallContext() from the transition function, i.e., you want to make a call from above the C++ AL to a (PostgreSQL-specific) function below the C++ AL. That breaks the purpose of an abstraction layer, and I would advise strongly against it. Bypassing ALs is not just smelly but code that stinks. :)

fschopp:
Putting information into `SystemInformation::user_fctx` again looks like a way to bypass the AL. Why do you need it? If there is a need, probably the C++ AL needs a new feature. But UDFs should never access `SystemInformation::user_fctx` directly.

Regarding [MADLIB-774](http://jira.madlib.net/browse/MADLIB-774), I added a comment to the JIRA.

yangs16:
Thanks Florian. I need SystemInformation::user_fctx to maintain some states across UDF calls. For some analytic algorithms, we need to read and update some global states across UDF calls. Although we can implement such kind of functionality via window function, but it will involve unnecessary overhead and may result in very serious performance problem. If we can transfer some states in a UDF, things will become simple and very efficient.

fschopp:
Maybe you can describe what you actually need. In 99.9% of the cases, maintaining global state is an (poor) implementation detail, but not a requirement. Can you elaborate? What is the algorithm? Why could you not maintain the state outside of the UDF (and pass it as an argument)? Why could you not implement the functionality on top of a UDA, and use a "normal" transition state?

fschopp:
Perhaps it is also worth moving the discussion to the developer forum.

src/ports/postgres/dbconnector/AnyType_proto.hpp
```diff
@@ -41,6 +41,9 @@ class AnyType {
     bool isComposite() const;
     AnyType& operator<<(const AnyType& inValue);
 
+    inline SystemInformation * getSysInfo(){
+        return this->mSysInfo;
+    }
```
fschopp: Also this is not something that belongs into `AnyType`. `SystemInformation` is a port-specific class, i.e., a class that only exists in the PostgreSQL port of the C++ AL.

yangs16:
This feature is needed when implementing LDA. For LDA, we need to maintain a count matrix of word by topic. In the UDF, we do some computations based on the count matrix and some other parameters, update the count matrix, and return some results. The count matrix will be passed to the UDF in the first call as a parameter, and then it will be put in memory, because we need read and update it in the coming calls. As I mentioned, we can implement the functionality on top of a UDA, and call the UDA as a window function (because we need to return some results for each  record), but it is very slow as compared with a "stateful" UDF. I have experimented with both implementations. I will move the discussion to the developer forum later.

cwelton:
Some background and initial discussion moved to devel@madlib.net as suggested by Florian.

PR #292, opened 2013-01-08
Abstraction Layer: Incompatibility of construct_array and construct_md_a...

yangs16:
...rray

between Greenplum and PostgreSQL

Jira: MADLIB-776
The construct_array and construct_md_array have different behaviors in
Greenplum and PostgreSQL. When the first parameter is set to NULL in both
functions, Greenplum can work well, but PostgreSQL will crash the system. The
reason is that PostgreSQL does not check for null pointer and always calls
CopyArrayEls even the pointer is null.


cwelton:
Was this closed because we pulled the changes into one of our branches?

yangs16:
Yes, the changes  have been merged into V0.6 together with LDA.

PR #294, opened 2013-01-10
Added Ridge and LASSO regulation for linear regression; Cross-validation

walkingsparrow:
(1) Ridge regulation for linear regression
(2) LASSO regulation for linear regression
(3) A general cross-validation function for modules that comply with a
consistent SQL API
(4) Specific cross-validation functions for Ridge and LASSO
(5) More functions in convex module, which can be used by other modules,
    but they are not exposed by any SQL interface.


haying:
Good to see my code comes back:) What's the plan for this pull request?

cwelton:
After review from the team and addition of more test cases we plan to pull it in and incorporate it into MADlib v0.6.

jhellerstein:
Woot!

PR #295, opened 2013-03-20
Created Slope One implementation

michaelbrand:
Hi!

In order to try out the "contribution guidelines" for MADlib for the first time, I created a very simple branch. This branch implements the Slope One algorithm. Essentially it adds only a single file: slope_one.sql_in. It has documentation inside the file, which should comply with the MADlib standards.

Please let me know if I missed any steps, or whether you can take it from here and merge this contribution into MADlib.

Best,
Michael.


cwelton:
Need to evaluate if it is desirable to have this method in MADlib.  If yes then we can evaluate for what changes are required to incorporate this into the product.

PR #296, opened 2013-05-07
Fix profile/quantile/summary functions for table/column names when starting with numeric and/or non-ascii characters.

snaga:
Fix profile/quantile/summary functions to allow table/column name which starts with numeric characters (0-9) and/or multi-byte characters. BTW, I'm not sure how other functions need to be dealt with.

### DIFF
--- a/src/ports/postgres/modules/data_profile/profile.py_in
+++ b/src/ports/postgres/modules/data_profile/profile.py_in
@@ -131,7 +131,7 @@ def __get_profile_data( schema, table, numcols, non_numcols, aggs, funclist, buc
     for c in numcols:
         for a in aggs[ funclist + '_num']:
             i += 1;
-            sql += ', ' + a.replace('%%',c).replace('()','('+c+')') + ' AS "' + str(i) + '"'
+            sql += ', ' + a.replace('%%',c).replace('()','("'+c+'")') + ' AS "' + str(i) + '"'
             rowset.append( {  'schema_name': schema
                             , 'table_name': table
                             , 'column_name': c
@@ -143,7 +143,7 @@ def __get_profile_data( schema, table, numcols, non_numcols, aggs, funclist, buc
     for c in non_numcols:
         for a in aggs[ funclist + '_nonnum']:
             i += 1;
-            sql += ', ' + a.replace('%%',c).replace('()','('+c+')') + ' AS "' + str(i) + '"'
+            sql += ', ' + a.replace('%%',c).replace('()','("'+c+'")') + ' AS "' + str(i) + '"'
             rowset.append( {  'schema_name': schema
                             , 'table_name': table
                             , 'column_name': c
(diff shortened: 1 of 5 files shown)


cwelton:
Confirmed that the issue exists and needs to be fixed.  

PR needs review, tests, and integration.

iyerr3:
Adding double-quotes to the table name would be a problem if the user inputs double quotes as part of the table name string. An alternative solution is to expect the user to add double quotes with the table name and we unquote the quotes when the table name is used within a 'where' clause. 
eg. SELECT 1 FROM information_schema.table WHERE table_name = '{table}'

Here double quotes will have to removed from {table}. We will need to take a decision on whether the user is expected to add double quotes for such table/column names or we add them within a query.

cwelton:
Desired design goals.
1. Be able to support any table/column that can be created in the database, including ones with multibyte characters
2. Have 'natural' database case sensitivity resolution.  E.g. 'mytable' and 'myTable' both naturally resolve to the same entity in SQL semantics unless they have been explicitly double quoted.

It looks like this pull request achieves design goal 1, but not design goal 2.

Some test scenarios:

```
CREATE TABLE mytable1(a int, "B" int, "多バイト文字" int);
CREATE TABLE "myTable2"(a int, "B" int, "多バイト文字" int);
CREATE TABLE "早見3"(a int, "B" int, "多バイト文字" int);

-- "regular" scenarios, should not be case sensitive
SELECT * FROM mytable1;
SELECT * FROM myTable1;
SELECT madlib.summary('mytable1', 'mytable1_summary1);     -- produces 'mytable1_summary1' as output
SELECT madlib.summary('myTable1', 'mytable1_Summary2');  -- produces 'mytable1_summary2' as output

-- The following would error due to table quoting rules
SELECT * FROM mytable2;
SELECT * FROM myTable2;
SELECT * FROM 早見3;
SELECT madlib.summary('mytable2', 'mytable2_summary_error');
SELECT madlib.summary('myTable2', 'myTable2_summary_error');
SELECT madlib.summary('早見3', '早見3_summary_error');

-- The following should work using regular database quoting rules
SELECT * from "myTable2";
SELECT * FROM "早見3";
SELECT madlib.summary($$ "myTable2" $$, $$ "myTable2_summary_error" $$);
SELECT madlib.summary($$ "早見3" $$, $$ "早見3_summary_error" $$);

-- With similar usage scenarios for column names in addition to table names.
```

haradh1:
We should probably use quote_ident() to avoid double-quoted input problem.

cwelton:
I thought the same thing, but the issue with quote_ident is that it WILL add double quotes to 'myTable' when we want normal database case sensitivity rules to apply.

fschopp:
Just happened to just see this. Hope all is well. In the code written by me it was a deliberate decision not to use double quotes (or quote_ident). The design document might even contains a note on this question. The idea was that, by the principle of least surprise, query-like function should behave as similar as normal queries, which includes being case-insensitive. As a separate concern, SQL injection is not a problem, either, because the functions run with the same permissions as the user that executes them (at least if the function is not declared with SECURITY DEFINER, which last year no MADlib function was).

PR #297, opened 2013-07-15
Doc edits

litzell:
In the linear and logistic regression modules, move the technical discussion to a new background section. In the logistic regression module, replace the example, which had an excerpt of data, with a working example taken from the the test module. In svec.sql_in, capture ways to create SVECs from forum email.


cwelton:
This has been pulled into the integration branch.  Closing PR.

PR #299, opened 2013-11-19
Removing top-level PCA doc.

vatsan:
Removing doc eyesore in PCA, until high levels docs for all top level modules are added :-)


iyerr3:
@vatsan Thanks for the helpful changes! Only suggestion would be to create a separate branch (and start PR from it) while pushing commits - makes it easier for us to merge the changes (since we don't use the direct merge button available here).

vatsan:
Sorry, didn't notice that Rahul. Will create a branch for future pull requests.

iyerr3:
PR Merged: 597fa7d69f7aa17ebdc68b4ca5c0a175a80064f6

PR #301, opened 2014-01-29
Specifically use /bin/bash in madpack

thinkerbot:
In general /bin/sh is equivalent to /bin/bash, but not on ubuntu where /bin/sh is actually /bin/dash.  This matters because the madpack script uses `pushd` and `popd` which are bash-specific enhancements; on ubuntu this script causes errors like:

```
/usr/local/madlib/bin/madpack: pushd: not found
/usr/local/madlib/bin/madpack: popd: not found
```

The fix is to specifically use bash.  For more info:
- http://askubuntu.com/questions/141928/what-is-difference-between-bin-sh-and-bin-bash

### DIFF
--- a/src/bin/madpack
+++ b/src/bin/madpack
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # This script does the following:
 # 0. If indicated by environent variables, look for DBMS-supplied Python


cwelton:
Change looks reasonable, will be pulled into next release.

haying:
Merged in [201927ac52](https://github.com/madlib/madlib/commit/201927ac5287181954361e7f90063feca81f4893).

PR #302, opened 2014-02-19
Fixing the elasticnet prediction functions to work with cross-validation

ronert:
When elasticnet net selects no variables at all during cross-validation (high penalty), the prediction functions generate empty arrays because they only use the selected variables. These empty arrays (array[]) then cause SQL syntax errors. This patch fixes the problem.

Jira: MADLIB-824
Signed-off-by: Ronert Obst ronert.obst@gmail.com


haying:
@walkingsparrow Could you take a look at this?

walkingsparrow:
The patch seems OK. We can go ahead to merge it.

haying:
Merged in [252a0cab08](https://github.com/madlib/madlib/commit/252a0cab08c44f8ffad6ea168156afd365d98af6).

PR #303, opened 2014-03-27
array_ops: Add array_scalar_add function

ihuston:
Jira: MADLIB-828

Adding a function to add a scalar to an array in a similar way to array_scalar_mult does for multiplication.


haying:
Merged in master: c75c357406.

PR #304, opened 2014-05-05
Add TypeTrait<const char*> structure

randomizedthinking:
Without it, assignment of a const char\* to AnyType will generate an error message.


haying:
It seems std::string (instead of char_) is the type you want type traits for, right? The downside of char_ (actually also const char_) is that it allows buggy code to crash the database, e.g. returning a char_ (or const char*) that does not set `'\0'` properly may cause memory error in `cstring_to_text()`. By using std::string and putting `WITH_TO_PG_CONVERSION(PointerGetDatum(cstring_to_text(value.c_str())));` can avoid this. I suggest this instead

``` c
template <>
struct TypeTraits<std::string> {
    typedef std::string value_type;

    WITH_OID( TEXTOID );
    WITH_TYPE_CLASS( dbal::SimpleType );
    WITH_MUTABILITY( dbal::Immutable );
    WITH_DEFAULT_EXTENDED_TRAITS;
    WITH_TO_PG_CONVERSION(PointerGetDatum(cstring_to_text(value.c_str())));
    WITH_TO_CXX_CONVERSION(text_to_cstring(DatumGetTextPP(value)));
};
```

src/modules/regress/glm_family.cpp
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @brief Return the Gaussian family object: family, linkfun, etc
+ */
+AnyType
+gaussian::run(AnyType &args) {
+	std::string linkfunc;
+
+	// 1. get the linkfun argument: default value -- "identity"
+	if (length(args) == 0) then linkfunc = "identity";
+	else linkfunc = args[0].getAs<char*>();
```
haying, line 67: FYI. we usually do string manipulation in python, and we have built some utility functions around it (src/ports/postgres/modules/utilities/*.py_in). But, of course, it is not a hard requirement and I have not yet fully understood your implementation. So this is just for information and, of course, you are free to have your own design. Great to see new contributors to MADlib!

randomizedthinking:
Thanks, Xixuan... It makes a lot of sense.

On Mon, May 5, 2014 at 11:30 PM, Feng, Xixuan (Aaron) <
notifications@github.com> wrote:

> It seems std::string (instead of char_) is the type you want type traits
> for, right? The downside of char_ (actually also const char_) is that it
> allows buggy code to crash the database, e.g. returning a char_ (or const
> char*) that does not set '\0' properly may cause memory error in
> cstring_to_text(). By using std::string and putting
> WITH_TO_PG_CONVERSION(PointerGetDatum(cstring_to_text(value.c_str())));can avoid this. I suggest this instead
> 
> template <>struct TypeTraitsstd::string {
>     typedef std::string value_type;
> 
> ```
> WITH_OID( TEXTOID );
> WITH_TYPE_CLASS( dbal::SimpleType );
> WITH_MUTABILITY( dbal::Immutable );
> WITH_DEFAULT_EXTENDED_TRAITS;
> WITH_TO_PG_CONVERSION(PointerGetDatum(cstring_to_text(value.c_str())));
> WITH_TO_CXX_CONVERSION(text_to_cstring(DatumGetTextPP(value)));};
> ```
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/madlib/madlib/pull/304#issuecomment-42271476
> .

## 

Gang Liang
Life is beautiful!

src/modules/regress/glm_family.cpp
randomizedthinking, line 67: Oops, never mind. This code is scratch: I know it won't work. It was
created after the pull request, but somehow the master was set for the
request. Need to get familiar with github...

On Mon, May 5, 2014 at 11:35 PM, Feng, Xixuan (Aaron) <
notifications@github.com> wrote:

> In src/modules/regress/glm_family.cpp:
> 
> > +#include "glm_family.hpp"
> > +
> > +namespace madlib {
> > +namespace modules {
> > +namespace regress {
> > +
> > +/**
> > - \* @brief Return the Gaussian family object: family, linkfun, etc
> > - */
> >   +AnyType
> >   +gaussian::run(AnyType &args) {
> > -   std::string linkfunc;
> >   +
> > -   // 1. get the linkfun argument: default value -- "identity"
> > -   if (length(args) == 0) then linkfunc = "identity";
> > -   else linkfunc = args[0].getAs<char*>();
> 
> FYI. we usually does string manipulation in python, and we have built some
> utility functions around it (src/ports/postgres/modules/utilities/*.py_in).
> But, of course, it is not a hard requirement and I have not yet fully
> understood your implementation. Just for information and, of course, you
> are free to have your own design. Great to see new contributors to MADlib!
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/madlib/madlib/pull/304/files#r12312115
> .

## 

Gang Liang
Life is beautiful!

PR #307, opened 2015-02-19
include materialized views when considering whether a table exists for postgresql

no0p:
I've found it's convenient to create training sets as materialized views.  

At present, the validation that a table exists relies on a query to information_schema.tables which unfortunately does not include materialized views.  This patch simply adds to the query to check the pg_catalog for a matching materialized view.

A workflow I'm enjoying is to create training sets and test sets as materialized views.  For example, you can take a simple random sample over large tables and do cpu expensive transforms, creating the data in a flattened materialized view containing parameter fields and independent variable fields. 

This nicely allows for creating a number of sampled sets and keeping them around.

### DIFF
--- a/src/ports/postgres/modules/regress/test/logistic.sql_in
+++ b/src/ports/postgres/modules/regress/test/logistic.sql_in
@@ -43,6 +43,9 @@ INSERT INTO patients(ID, second_attack, treatment, trait_anxiety) VALUES
 (21, 0, 0, NULL),
 (22, 0, NULL, 60);
 
+CREATE MATERIALIZED VIEW patients_mat_view AS 
+  (SELECT * FROM patients) ;
+
 drop table if exists temp_result;
 drop table if exists temp_result_summary;
 select logregr_train(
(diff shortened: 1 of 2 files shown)


haying:
Hi Robert,

The idea is nice. I think all we require is that the input is a relation (table, materialized view, or view, etc.). Thus, including also normal views may be something along the same line too. If you could add a test case (or more) in the install-check of a module that uses this `table_exists()` function (e.g. regressions, ARIMA or decision trees etc.), which train a model on a materialized view. It will be ready to be merged.
- Aaron

no0p:
Dear Aaron,

Sounds good.  I was giving it some more thought, and this may have backwards compatibility issues prior to 9.3.  It looked like there was an sql function available which might solve the compatibility issues.  I'll take another look soon.

-Robert

haying:
Sounds good. We were even thinking just try and catch "select \* from {input} limit 0". But you could try anything out. As long as test cases look fine, it is okay.

Thanks,
Aaron

no0p:
Aaron,

It looks like I need some guidance as the solution I would propose has some potential complications.

The committed updates work as expected, including the test case which exercises materialized views in the Postgresql regression module.

However this approach relies on the madlib_schema.__table_exists function and in this non-merge-ready patch the schema name is hard coded to madlib.  Thus this is presented only for consideration of the following problem.

It looks like other validation methods pass the madlib_schema value, e.g. _validate_args.  Correctly using this approach would potentially require updating all calls to table_exists (of which there are many) to pass in the schema name as a variable.  Of course I'm not particularly familiar with Python so there may be a more clever way.

Additionally I would expect this test case to fail for PG_VERSIONS < 9.3 and am not entirely sure how version dependent tests are handled.

Perhaps the select \* from -> catch exception approach is preferable for the above reasons.  Thoughts?

haying:
Hi Robert,

My feeling that "select \* from {input}" sounds better as it does not care any constructs in catalog tables, which makes it more maintainable and understandable. For the use of try-catch, you could refer to function `is_var_valid()` in the same file. Your install-check case also looks fine.

no0p:
Aaron,

I think this should match expectations.  

There is still the issue of materialized views only being implemented for PG_VERSION >= 9.3.  

-Robert

haying:
Hi Robert,

After some testing, I found out that the try-catch approach somehow hit an PL/Python issue in some old PostgreSQL versions that we still wants to support (See [link](http://www.postgresql.org/message-id/AANLkTi=aaz=kv_08QH1tGPhK4b_Q1TwRk1tMbQX3fWVk@mail.gmail.com)).

I have an idea to do this by using the `relkind` column in the `pg_class` table to do this (not using `pg_matviews` due to backward compatibility).

> r = ordinary table, i = index, S = sequence, v = view, m = materialized view, c = composite type, t = TOAST table, f = foreign table"
> Probably also joining `pg_namespace` to get schema names.

Let me know if you would like to give it a try. Or I might just go ahead to write it.

no0p:
Aaron, please go ahead as I do not have a complete sense of compatibility and maintenance goals.  I look forward to using materialized views in future versions.

haying:
Merged in [3a7bc49](https://github.com/madlib/madlib/commit/3a7bc494248bd66e6b52e809375eb173bfbb0fc0)

PR #308, opened 2015-02-24
FreeBSD adjustments

oberstet:
Using this patch, I was able to build MADlib trunk on FreeBSD 10.1 and PostgreSQL 9.4 (using gcc48 - clang currently has more issues)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,11 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
     set(SOLARIS TRUE)
 endif(${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
 
+if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+    set(FREEBSD TRUE)
+    include_directories("/usr/local/include")
+endif(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RelWithDebInfo CACHE STRING
       "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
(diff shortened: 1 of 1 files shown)


haying:
Hi Tobias,

The diff looks okay, although I guess it would also work if you set CXX and CC to add the include, like

```
$ CXX="g++ -I/usr/local/include" CC="gcc -I/usr/local/include" cmake /path/to/madlib_src
```

I am not familiar with FreeBSD and am curious what header files are needed from `/usr/local/include` in your environment? If they are expected to be needed in many environment, I am good to merge this PR.

Thanks,
Aaron

oberstet:
Hi Aaron,

`/usr/local/include` is the standard place on FreeBSD for headers from everything not in the base system (coming from "ports collection"). I think it was `libintl.h` it was missing. Note that GCC is no longer the base system compiler on FreeBSD (it is clang nowerdays).

Yes, I could have set include path from CXX/CC .. the m4 patch is definitely required though (the BSD m4 bails out on the GNU `--prefix-builtins` option).

For completeness, here is what I did:

```
pkg add gmake gcc48 bash-static python27
export CC=`which gcc48`
export CXX=`which g++48`
ln -s /usr/local/bin/python2.7 /usr/local/bin/python
ln -s /usr/local/bin/bash /bin/bash
```

The latter 2 are ugly. The latter symlink is for the `madpack` script to work (which assumes `/bin/bash` is shebang). The former symlink I also needed.

oberstet:
@haying anything you want me to change or add for merging?

haying:
Hi Tobias,

I will merge the m4 option change. Given GCC is not the default compiler for FreeBSD and we have a known work-around, I will skip the include path. The code should be in sometime this week.

haying:
Merged in [edef045](https://github.com/madlib/madlib/commit/edef0453e5c87b7c4eaa886686a0848e0ba121d4)

oberstet:
thanks!

PR #309, opened 2015-04-04
fix sparse vector encoding

raphaelhoffmann:
madlib.__gen_svec sometimes generates wrong sparse vector encodings when indexes are larger than 255. Here's an example that reproduces the error:

select count(*) from (select madlib.svec_unnest( madlib.__gen_svec('{7855,8623}', '{1,1}', 8624)) as b) a where not b = 0;

I expect an output of 2 (since two entries in the vector have a non-zero entry), but I get 1. 

This should fix it.

### DIFF
--- a/methods/svec_util/src/pg_gp/generate_svec.c
+++ b/methods/svec_util/src/pg_gp/generate_svec.c
@@ -1,5 +1,6 @@
 #include <postgres.h>
 
+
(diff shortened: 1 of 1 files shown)


haying:
Merged in [7fe307f](https://github.com/madlib/madlib/commit/7fe307f538260e473f2b5bc9c749baa61c1e991e).

